PR #8, opened 2016-02-26
Pinch Example Module

Amarcolina:
This is an example module that includes some starter/scaffold code for working with pinch gestures.  Features include:

- The LeapPinchDetector component, which provides easy-to-access information using the new pinch distance API about whether a hand is pinching or not, and other event based information like begin and end events.
- The LeapRTS component, which is a component that allows users to rotate, translate, and scale a game object using a combination of one handed or two handed pinch gestures. 

There is also a couple of demo scenes that show off the rts component, as well as using it to draw lines.  

- [x] Verify that the LeapPinchDetector fits into our architectural style and seems properly coded.
- [x] Verify that the RTS demo works as expected, you should be able to move around the fractal cube, rotate it, and scale it larger and smaller.
- [x] Verify that the Pinch Drawing demo works as expected, you should be able to draw lines with either hand.

@protodeep


Amarcolina:
Closing for now, going to re-open once @protodeep 's PR goes through tomorrow.

Issue #15, opened 2016-02-26
Leap.Finger.TransformedCopy doesn't propagate the fingerId correctly


cherullo:
The Finger constructor sets 

`_id = (handId *10) + fingerId`

The fingerId itself is not stored in the Finger class. Later, when TransformedCopy is called, this _id is passed as fingerId to the constructor, which once again adds handId * 10.
The result is that the Id of the transformed copy is different from the original object.

jo3w4rd:
Fixed by platform #5561

Issue #6, opened 2016-02-25
Issues in Leap.Hand.Basis and Leap.Hand.TransformedCopy


cherullo:
There are two issues in this method:
```
//TODO verify this calculation for both hands
_basis.zBasis = -Direction;
_basis.yBasis = -PalmNormal;
_basis.xBasis = _basis.zBasis.Cross(_basis.yBasis);
```
The first is that, as the comment present in code suggests, the cross product must be different for each hand, if a left-handed basis is expected for left hands, and a right-handed basis is expected for right hands.
The second is *much* more subtle. When you call the method Hand.TransformedCopy, a new Hand instance is created where the vectors Direction and PalmNormal are multiplied by a matrix, usually the LeapMatrix. Now, the LeapMatrix _M_, as currently provided by the method Controller.GetLeapMatrix, has a reflection on the Z axis, and thus we have that: 

_M_(-Direction x -PalmNormal) != _M_(-Direction) x _M_(-PalmNormal)

Which is to say that  the value of Hand.TransformedCopy(...).Basis.xBasis is wrong.

The following snippet illustrates the issue:
```
      Matrix leapMat = GetLeapMatrix();
      Frame frame = leap_controller_.Frame(); 
      Frame transformed = frame.TransformedCopy( leapMat );

      if (frame.Hands.Count > 0)
      {
          Assert.AreEqual(
              // _M_(-Direction x -PalmNormal)
              leapMat.TransformDirection(frame.Hands[0].Basis.xBasis).Normalized,

              // _M_(-Direction) x _M_(-PalmNormal)
              transformed.Hands[0].Basis.xBasis
              );
      }

```
The assertion always fails.

jo3w4rd:
This has been fixed by the change to use Quaternions and the change to use a consistent basis type for both hands -- always right-handed.

PR #16, opened 2016-02-26
Implemented interfaces for Frame, Hand, Finger, Bone and Arm,  along …

cherullo:
…with zero-copy implementations of TransformedCopy.
The objective was to eliminate most memory allocations that contribute to GC stutter, most of which happened on Frame.TransformedCopy.
This was achieved by creating classes which transform the original values on demand, instead of transforming and copying all the values before-hand. Instances of these classes are obtained calling TransformedShallowCopy.
Of course, these Transformed* classes have different performance characteristics, but in the demo scene, using the CapsuleHand, the aggregated performance was improved. Even thou TransformedShallowCopy is much faster than TransformedCopy, the following scripts are slower since the actual calculation is happening lazily. In the end it's a clear win.

I needed to check whether those implementations were correct, so I created an Asserter class which compares two IFrame instances (and all it's children objects). This uncovered a few bugs in the original implementations and the new Transformed classes. For example, the bug #6 reported early doesn't affect the new classes. Anyway, there is some new code in LeapProvider.Update to optionally compare Frames transformed by TransformedCopy and TransformedShallowCopy.

Once those interfaces were created, I also created Invalid versions of those objects. These implementations throw NotImplementedExceptions (should change to something more appropriate like  IllegalOperationException) for every method, except IsValid and Equals. This allows a much more defensive coding style, where you MUST check IsValid before accessing an object, otherwise an exception is thrown. The static Invalid properties then were changed to return appropriate Invalid* singletons too, further reducing memory allocation.

Another huge source of memory allocations was in LeapImageRetriever. The _combinedTexture was being destroyed every frame just to be recreated. This was fixed too.

Looking at the Unity profiler, there is just a 44 byte allocation of a iterator in UpdateHandRepresentations now. 

Remaining fixes and improvements:
FIX: Leap.Arm.TransformedCopy didn't normalize Direction.
FIX: Fixed FingerId propagation issue (#15)
IMP: TransformedCopy methods now receive Matrix by ref.
IMP: Methods FingerList.{Left, Front, Right}most don't allocate a temporary invalid Finger.
IMP: Methods HandList.{Left, Front, Right}most don't allocate a temporary invalid Finger.
IMP: CapsuleHand.updateCapsules performance improvements
IMP: LeapProvider.GetFixedFrame: only transform the returned frame.

### DIFF
--- a/Assets/LeapC/Finger.cs
+++ b/Assets/LeapC/Finger.cs
@@ -32,13 +32,15 @@ namespace Leap
    * @since 1.0
    */
 
-    public class Finger
+    public class Finger : IFinger
     {
-        Bone[] _bones = new Bone[4];
+        IBone[] _bones = new IBone[4];
         FingerType _type = FingerType.TYPE_UNKNOWN;
          int _frameId;
-         int _id = 0;
          int _handID = 0;
+         int _fingerId = 0;
+         int _id = 0;
+         
          Vector _tipPosition;
          Vector _tipVelocity;
          Vector _direction;
@@ -72,10 +74,10 @@ namespace Leap
                            float length,
                            bool isExtended,
                            Finger.FingerType type,
-                           Bone metacarpal,
-                           Bone proximal,
-                           Bone intermediate,
-                           Bone distal)
+                           IBone metacarpal,
+                           IBone proximal,
+                           IBone intermediate,
+                           IBone distal)
         {
             _type = type;
             _bones [0] = metacarpal;
@@ -83,8 +85,9 @@ namespace Leap
             _bones [2] = intermediate;
             _bones [3] = distal;
             _frameId = frameId;
-            _id = (handId * 10) + fingerId;
+            _fingerId = fingerId;
             _handID = handId;
+            _id = (handId * 10) + fingerId;
             _tipPosition = tipPosition;
             _tipVelocity = tipVelocity;
             _direction = direction;
@@ -96,12 +99,18 @@ namespace Leap
             _timeVisible = timeVisible;
         }
 
-        public Finger TransformedCopy(Matrix trs){
+        public IFinger TransformedShallowCopy(ref Matrix trs)
+        {
+            return new TransformedFinger(ref trs, this);
+        }
+
+        public IFinger TransformedCopy(ref Matrix trs) {
+
             float dScale = trs.zBasis.Magnitude;
             float hScale = trs.xBasis.Magnitude;
             return new Finger(_frameId,
                               _handID,
-                              _id,
+                              _fingerId,
                               _timeVisible,
                               trs.TransformPoint(_tipPosition),
                               trs.TransformPoint(_tipVelocity),
@@ -111,10 +120,10 @@ namespace Leap
                               _length * dScale,
                               _isExtended,
                               _type,
-                              _bones[0].TransformedCopy(trs),
-                              _bones[1].TransformedCopy(trs),
-                              _bones[2].TransformedCopy(trs),
-                              _bones[3].TransformedCopy(trs));
+                              _bones[0].TransformedCopy(ref trs),
+                              _bones[1].TransformedCopy(ref trs),
+                              _bones[2].TransformedCopy(ref trs),
+                              _bones[3].TransformedCopy(ref trs));
         }
 
 
@@ -128,7 +137,7 @@ namespace Leap
      * @returns The Bone that has the specified bone type.
      * @since 2.0
      */
-        public Bone Bone (Bone.BoneType boneIx)
+        public IBone Bone (Bone.BoneType boneIx)
         {
             return _bones[(int)boneIx];
         }
@@ -185,6 +194,13 @@ namespace Leap
             } 
         }
 
+        public int FrameId
+        {
+            get
+            {
+                return _frameId;
+            }
+        }
         /**
      * The Hand associated with a finger.
      *
@@ -356,9 +372,11 @@ namespace Leap
      * @returns The invalid Finger instance.
      * @since 1.0
      */
-        public static Finger Invalid {
+        static private IFinger invalidFinger = new InvalidFinger();
+
+        public static IFinger Invalid {
             get {
-                return new Finger ();
+                return invalidFinger;
             } 
         }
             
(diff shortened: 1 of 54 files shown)


codemercenary:
Holy macrel!

I'd like to get this merged, but I do have a request:  Would you mind breaking it up a bit more?  Perhaps submitting the bugfixes in a separate PR, so we can understand and comment on them separately?

codemercenary:
Nevermind, looks like the work is too tied together.  We'll leave it as it is.

Assets/LeapC/Finger.cs
```diff
@@ -356,9 +372,11 @@ public override string ToString ()
      * @returns The invalid Finger instance.
      * @since 1.0
      */
-        public static Finger Invalid {
```
codemercenary: We actually deprecated the `Invalid` concept at this layer in the new API, you should be able to just remove all of this code once we update the wrappers here.

Assets/LeapC/FingerList.cs
```diff
@@ -55,7 +55,7 @@ public FingerList Extended ()
      */
         public FingerList FingerType (Finger.FingerType type)
         {
-            return (FingerList) this.FindAll (delegate (Finger finger) {
+            return (FingerList) this.FindAll (delegate (IFinger finger) {
                 return finger.Type == type;
             });
         }
```
codemercenary: We removed these, too, they are a bit atrocious.

Assets/LeapC/Frame.cs
```diff
@@ -50,7 +50,12 @@ public Frame (long id, long timestamp, float fps, InteractionBox interactionBox)
             InteractionBox = interactionBox;
         }
 
-        public Frame TransformedCopy (Matrix trs)
+        public IFrame TransformedShallowCopy (ref Matrix trs)
+        {
+            return new TransformedFrame(ref trs, this);
```
codemercenary: I see, this is the reason you needed interfaces everywhere.

Assets/LeapC/InvalidArm.cs
```diff
@@ -0,0 +1,93 @@
+﻿using UnityEngine;
+using System.Collections;
+
+namespace Leap
+{
+    public class InvalidArm : IArm
+    {
+        public Vector ElbowPosition
+        {
+            get { throw new System.NotImplementedException(); }
```
codemercenary: Definitely a superior strategy to implement an invalid concept--but we chose to just deprecate the whole thing instead.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
```diff
@@ -89,6 +89,10 @@ public class LeapTextureData {
       TextureFormat format = getTextureFormat(image);
 
       if (_combinedTexture != null) {
+				if ((_combinedTexture.height == combinedHeight) &&
```
Amarcolina: You would need to check image format as well, since that can theoretically change even if the width/height remains the same.  However, this method should only be called when a reconstruct is needed, if it is being called every frame, that is a bug elsewhere and should be addressed elsewhere.  (I think it has been fixed, the stale flag wasn't being reset before).  

Since it is only going to be called infrequently (upon device connection / flip / swap) I would be hesitant to try adding complexity to this method to try to detect whether or not every property is the same, as it might cause difficult-to-track-down edge cases like this one.

Amarcolina:
This PR should target the develop branch instead of the master branch.  Develop branch is for developmental work, master branch is only for releases.

cherullo:
Let me apologize for doing things in an awkward manner. I didn't know that the UnityModules repo existed, let alone that it was publicly visible. So I didn't expect to ever create a pull request, I was prepared to keep my private copy of those scripts with our changes, and periodically merge with the official releases. I was about to blow my timebox, so I rushed it even more. Next time I'll create issues to discuss possible solutions instead of dropping a PR from out of nowhere.

Also, does the C# project have automated testing? I really needed the test code in LeapProvider and the Asserter class, it helped a lot in finding bugs, but it sure shouldn't be there.

Overall, I'm glad that this PR was received with such warmth, and I'll find some time to rewrite it.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
cherullo: Ok, I'll take this out. I just "stopped the bleeding", but I didn't know what was really going on.

Assets/LeapC/InvalidArm.cs
cherullo: Even better. I'll take these classes out of the PR then.

Assets/LeapC/Frame.cs
cherullo: Well, the first implementation

Assets/LeapC/FingerList.cs
cherullo: Ok, I'll take this out.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
codemercenary: Hah.  Good reasoning.

codemercenary:
Don't worry about it at all, it's easy enough to fix, though we will need to open a new PR.

The current system is three-layer API.  LeapC is on the bottom, then LeapCSharp (which is what this PR touches), and then at the top are the Unity wrappers.  Originally, LeapC didn't exist, and LeapCSharp was generated code.  We've always known LeapCSharp is horrible but until recently we couldn't do anything about it.

So, we spent the last two months rewriting the underlying API just so that we could change the wrapper.  We actually did fix a lot of issues in LeapCSharp, but our focus was just in getting LeapC completely stable.  We were about to blow our timebox for the Orion launch, so we focused all of our efforts on LeapC.  LeapCSharp had bugs, but it was at least functional.

Four engineers were dedicated to this project for two months for the sole purpose of allowing us to correcting these specific garbage collection problems.  And then, along comes this pull request.  Holy guacamole.  This is amazing.

THANK YOU for doing this, if you hadn't, it might have been another month before this stuff got out there.  If this isn't a testament to the power of open source, I don't know what is.

codemercenary:
Also, we do have unit tests.  They're in a monolithic repository, though, and so I can't show you them directly.  I'm working on splitting up the monolith now, when it's done I'll make that stuff open source as well.

codemercenary:
Closing this as superceded by #21

PR #18, opened 2016-02-27
Fix check image setting

### LINKED ISSUE #9: Disabling Images Reduces FPS to 50
If you disable images via control panel or in Unity it reduces the core assets fps to 50 and produces weird judder artifacts.

jo3w4rd:
Added switch on whether to request images. 
This switch is checked when LeapImageRetriever is enabled (in controller connect event).
A coroutine is started to poll the setting if images are turned off and if an image fails to complete. 

- [x] Fixes #9


Assets/LeapMotion/Scripts/LeapImageRetriever.cs
```diff
@@ -322,41 +321,42 @@ public class EyeTextureData {
...
+      while (!_requestedImage.IsComplete) {
+        if (controller.Now() - start > ImageTimeout) break;
+      }
+      if (_requestedImage.IsComplete) {
+        if (_eyeTextureData.CheckStale(_requestedImage, _requestedImage)) {
+          _eyeTextureData.Reconstruct(_requestedImage, _requestedImage);
+        }
+        _eyeTextureData.UpdateTextures(_requestedImage, _requestedImage);
+      } else if(!checkingImageState){
+        StartCoroutine(CheckImageMode(2));
```
Amarcolina: Probably should make this 2second constant defined as a const float at the top of the class.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
```diff
@@ -322,41 +321,42 @@ public class EyeTextureData {
...
+        checkingImageState = true;
```
Amarcolina: You can put this set to true inside of CheckImageMode (before the waitforseconds), that way starting it is always a 1 line operation.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
```diff
@@ -37,18 +38,14 @@ public class LeapImageRetriever : MonoBehaviour {
...
   protected Image _requestedImage = new Image();
-
-  //Debug vars for image retrieval
-  protected int _requestedImages = 0;
-  protected int _lateImages = 0;
-  protected int _superLateImages = 0;
-  protected float _startTime = 0;
-  protected float _lastTime = 0;
+  [SerializeField]
+  protected long ImageTimeout = 9000; //microseconds
```
Amarcolina: This variable doesn't seem like it is being used?

jo3w4rd: It is used in OnPreRender (329) in place of the 9000 microsecond hardcoded timeout value.

Amarcolina: Oh awesome, didn't find it for some reason

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
jo3w4rd: Done.

Assets/LeapMotion/Scripts/LeapImageRetriever.cs
jo3w4rd: Okay, done.

Amarcolina:
Works great!

PR #19, opened 2016-02-27
Feature hand pool improvements

protodeep:


Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -14,29 +14,20 @@
...
     HandFactory
-
   {
-    public IHandModel LeftGraphicsModel;
-    public IHandModel RightGraphicsModel;
-    public IHandModel LeftPhysicsModel;
-    public IHandModel RightPhysicsModel;
+
+    [SerializeField]
+    public List<IHandModel> ModelCollection;
```
Amarcolina: We probably should make this field private (and keep the SerializeField attribute so it shows up in the inspector)

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -14,29 +14,20 @@
...
     public List<IHandModel> ModelPool;
```
Amarcolina: It might be a good idea to consider making this private so it doesn't show up in the inspector.  We run the risk of having users modify the pool manually, which might cause undesired behavior

Assets/LeapMotion/Scripts/HandProxy.cs
```diff
@@ -6,7 +6,6 @@ public class HandProxy:
     HandRepresentation
   {
     HandPool parent;
```
Amarcolina: We should move the field handModel from HandRepresentation to HandProxy

Assets/LeapMotion/Scripts/Hands/HandEnableDisable.cs
```diff
@@ -4,12 +4,12 @@
...
 public class HandEnableDisable : HandTransitionBehavior {
 
-  public override void Reset() {
+  protected override void HandReset() {
     gameObject.SetActive(true);
   }
 
 	// Use this for initialization
-	public override void HandFinish () {
+	protected override void HandFinish () {
```
Amarcolina: Looks like there are some weird tabs in this file.

Assets/LeapMotion/Scripts/Hands/HandTransitionBehavior.cs
```diff
@@ -4,7 +4,17 @@
 namespace Leap {
   public abstract class HandTransitionBehavior : MonoBehaviour {
 
-    public abstract void Reset();
-    public abstract void HandFinish();
+    protected abstract void HandReset();
+    protected abstract void HandFinish();
+    protected virtual void Awake(){
+      IHandModel iHandModel = GetComponent<IHandModel>();
```
Amarcolina: We might want to throw some null checks in here, that way we can throw warnings about HandTransitionBehavior requiring an IHandModel to function correctly, instead of throwing null reference exceptions if there isn't one.

Amarcolina:
Looks good to me!

PR #21, opened 2016-02-29
Implemented TransformedShallowCopy

cherullo:
This is a rewrite of the PR #16, after the feedback from @Amarcolina and @codemercenary .

It's basically the same functionality, with the following differences:
- Kept CapsuleHand.updateCapsules performance improvements out of the PR.
- Removed the Invalid* classes.
- Created PR against the develop branch.
- Copied the methods' docs to the new interfaces.
- Avoided the use of FingerList and HandList classes. Had to edit them anyway.

The Asserter class and the test code in LeapProvider is still there. It's easy to remove, but it's great to increase the confidence in the new code. It's asserting a difference in the Arm values when a Hand is detected, so there may be a bug lurking somewhere.


Amarcolina:
Closing this for now for the reason outlined in #28 by @jo3w4rd.

PR #28, opened 2016-03-01
Update leap c sharp

jo3w4rd:
This merge's cherullo's PR with conflicts fixed.

### DIFF
--- a/Assets/LeapC/Arm.cs
+++ b/Assets/LeapC/Arm.cs
@@ -14,7 +14,7 @@ namespace Leap
    * The Arm class represents the forearm.
    *
    */
-  public class Arm : Bone
+  public class Arm : Bone, IArm
   {
     /**
      * Constructs a default Arm object.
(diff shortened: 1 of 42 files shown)


Amarcolina:
There seem to be some remaining compile errors that need to be fixed.  Also, I would want to do a detailed performance analysis on the changes.  Since the change makes transformed frames more expensive to access, it is theoretically possible that this change can make applications slower if they access transformed frame data frequently, since the extra cost of access might counteract the lower initial cost for transformation.

jo3w4rd:
As far as performance, this change gives you the option to do a "shallow copy" which transforms the original data on demand or use the full copy, which transforms the data up front (at a cost of about 4K GC allocation up front).

jo3w4rd:
Investigating pooled objects as an alternative to in-place modifications of a "current" frame object -- which has both performance implications and the potential to cause subtle problems when some one holds a reference to the current frame and it changes identity.

Issue #2, opened 2016-02-25
LeapUnityExtensions need some fixing


jo3w4rd:
FlipZ was changed to not flip at all. It should be set to (1, 1, -1). 
ToUnityScaled() doesn't scale. In its current form, it is equivalent to ToVector3(), simply converting a Leap.Vector object to a Unity Vector3 object.

There are several uses of ToUnityScaled() in the Orion assets. These also need to be changed to ToVector3().

PR #29, opened 2016-03-02
Converting all uses of ToUnityScaled and ToUnity to ToVector3 and mak…

protodeep:
…ing Flip Z flip again

Addressing Issue #2

### DIFF
--- a/Assets/LeapMotion/Scripts/Utils/LeapUnityExtensions.cs
+++ b/Assets/LeapMotion/Scripts/Utils/LeapUnityExtensions.cs
@@ -20,7 +20,7 @@ namespace Leap {
     /** Scale factor from Leap units (millimeters) to Unity units (meters). */
     public const float INPUT_SCALE = 0.001f;
     /** Constant used when converting from right-handed to left-handed axes.*/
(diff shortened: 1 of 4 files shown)


Assets/LeapMotion/Scripts/Utils/LeapUnityExtensions.cs
```diff
@@ -98,7 +98,7 @@ public static class UnityMatrixExtension {
      * @param mirror If true, the operation is reflected along the z axis.
      */
     public static Vector3 Translation(this Matrix matrix, bool mirror = false) {
-      return matrix.TransformPoint(LEAP_ORIGIN).ToUnityScaled();
+      return matrix.TransformPoint(LEAP_ORIGIN).ToVector3();
```
jo3w4rd: This substantially changes the way the function works -- it no longer converts the coordinate system. However, there isn't a clear winner between the old way and the new way -- if the caller has a matrix from an already transformed frame, they want the new version, if they have one from a pre-transformed frame, they want the old way.

Also, it would be a bit more efficient to implement this as:

    return matrix.origin.ToVector3();

or

    return matrix.origin.ToUnityScaled();

PR #30, opened 2016-03-02
Reducing memory allocation on worker thread

Amarcolina:
The Marshal.PtrToStruct(IntPtr, Type) method allocates an object every time it is called.  These objects are boxed structs, so the memory is thrown away as soon as the objects are cast to the struct type, generating garbage.  This garbage is consequently as large or larger than a frame object.

I have added StructMarshal to allow converting from IntPtr's to Structs without allocation through the use of an intermediate class instance and the Marshal.PtrToStruct(IntPtr, object) method.  The alternate marshal method does not allocate memory, and instead simply copies the data.  The util then simply extracts the struct value and returns it.  Another utility method was also created for array access.

It is difficult to directly observe the savings, since the Unity profiler does not directly show the allocations done on the worker thread.  However I have noticed that the total allocated memory seems to climb about twice as slow when using the StructMarshal.  

@jo3w4rd


jo3w4rd:
Not really the right place for this change as the LeapC files live in platform. I have made a pull request with this change on platform and will commit this here in the meantime.

PR #33, opened 2016-03-02
Fix remove dependency from image retriever

protodeep:
- [ ] Run Leap_Hands_Demo_VR scene and note absence of LeapImageRetriever.

- [ ] Run Leap_Hands_Demo_AR and note presence of LeapImageRetriever.


Amarcolina:
Looks great to me!

PR #36, opened 2016-03-04
Removing ToUnity from LeapUnityExtensions.Rotation()

protodeep:
Changed conversions within LeapUnityExtensios' Rotation() method to use .ToVector3 vs. ToUnity().


Amarcolina:
Works great!

PR #37, opened 2016-03-04
Removing IgnoreCollisionsWithHands() from LeapHandController (not use…

protodeep:
…d and should move is brought back)


Amarcolina:
Looks good to me!

Issue #22, opened 2016-02-29
GetFixedFrame transforms more frames than it needs to


Amarcolina:
The current implementation of LeapProvider.GetFixedFrame calls GetTransformedFrame every time it obtains a new frame inside the inner loop.  All but one of these frames is going to be discarded, and so the transformation is wasteful.  It should acquire un-transformed frame objects, and only transform the once that is selected.

cherullo:
This was also solved in PR #16.
By the way, regarding @protodeep 's commit above, it doesn't actually avoid multiple transformations, since the assignment to `closest` can happen more than once.
The `leapMat` assignment can also be taken out of the loop, since it's not supposed to change each iteration.
Also, isn't `leap_controller_.Frame() == leap_controller_.Frame(0)` ? If that's so, then shouldn't the loop begin with `searchHistoryIndex = 1` ?

codemercenary:
Tracking research is actually interested in building an extrapolation mechanism into the API which puts the interpolation responsibility on the API side.  This would allow us to feed a timestamp down in to LeapC and retrieve an interpolated frame corresponding to that timestamp.

We should find a solution here which enables us to swap out the way the interpolation works at a later time, once it's supported by the underlying system.

PR #43, opened 2016-03-04
In LeapProvider.FixedUpdate() only transform the selected frame Issue…

protodeep:
… #22

### DIFF
--- a/Assets/LeapMotion/Scripts/LeapProvider.cs
+++ b/Assets/LeapMotion/Scripts/LeapProvider.cs
@@ -136,7 +136,7 @@ namespace Leap {
       //which frame to deliver
     }
     public virtual Frame GetFixedFrame() {
-
+      int fixedFrameIndex = 0;
       //Aproximate the correct timestamp given the current fixed time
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Scripts/LeapProvider.cs
```diff
@@ -136,7 +136,7 @@ public class LeapProvider :
       //which frame to deliver
     }
     public virtual Frame GetFixedFrame() {
-
+      int fixedFrameIndex = 0;
```
Amarcolina: This variable is no longer being used.

Assets/LeapMotion/Scripts/LeapProvider.cs
```diff
@@ -145,7 +144,7 @@ public class LeapProvider :
       for (int searchHistoryIndex = 0; searchHistoryIndex < 60; searchHistoryIndex++) {
 
         leapMat = UnityMatrixExtension.GetLeapMatrix(this.transform);
```
Amarcolina: Last change I think, this recalculation of the leap matrix probably should be brought out of the loop.  Possible place it right before line 162?

Amarcolina:
Great!  Merging this in

PR #45, opened 2016-03-05
Fix hot reloading null refs

protodeep:
This brings in

- connection improvements and struct marshalling in LeapCSharp from platform
- latest LeapC.dll's
- simple stopping of editor playback on hot reload


Assets/LeapC/Logger.cs
```diff
@@ -20,7 +20,7 @@ public static void Log(object message)
     {
 #if DEBUG
 #if UNITY_EDITOR
-                    UnityEngine.Debug.Log(message);
+                    //UnityEngine.Debug.Log(message);
```
Amarcolina: Any reason we are commenting out this log statement?

Assets/LeapC/StructMarshal.cs
```diff
@@ -16,8 +16,7 @@ private class StructContainer {
     private static int _sizeofT = Marshal.SizeOf(typeof(T));
 
     /**
-     * Converts an IntPtr to a struct of type T.  Does not allocate any
-     * garbage unlike Marshal.PtrToStruct(IntPtr, Type).
+     * Converts an IntPtr to a struct of type T.
```
Amarcolina: Any reason this comment was changed?  The lack of garbage is a pretty important aspect of its operation.

Amarcolina:
Merging this in for now, since it has important changes.  I'll make an issue pointing back here to address the comments.  Everything works great!

Assets/LeapC/Logger.cs
jo3w4rd: Probably just a spurious change done while debugging and not properly restored.

PR #55, opened 2016-03-08
Fix #39, build settings should be empty.

### LINKED ISSUE #39: Build settings should not be building anrdoid by default
The build settings should either be empty, or set to something like the LeapDesktop scene.  Currently on develop the editor build settings are building the android scene.

hham:
Fixes #39


Amarcolina:
Wohoo!  Looks great to me!

PR #58, opened 2016-03-09
Fix remove last to unity

protodeep:


jo3w4rd:
I'm not sure we want to remove ToUnity functions. We at least need some docs on how to replace them. Maybe that change should be in a different PR.

PR #79, opened 2016-03-19
Update libLeapC.so for Android Unity

hham:
Update libLeapC.so.
- Fetched from 9fbc183 from develop of platform
- stripped


Amarcolina:
Tested this in the new demo, works great!

PR #83, opened 2016-03-22
PinchDetector components moved to their own transforms, position upda…

protodeep:
…ted with hand, warning if on IHandModel

- [x] verify that pinch demo scenes work properly

- [x] Warning is thrown on Awake() is PinchDetector is directly attached to IHandModel's transform

### DIFF
--- a/Assets/LeapMotionModules/PinchUtility/Scripts/LeapPinchDetector.cs
+++ b/Assets/LeapMotionModules/PinchUtility/Scripts/LeapPinchDetector.cs
@@ -44,6 +44,9 @@ namespace Leap.Unity.PinchUtility {
     }
 
     protected virtual void Awake() {
+      if (GetComponent<IHandModel>() != null) {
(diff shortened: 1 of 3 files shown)


Assets/LeapMotionModules/PinchUtility/Scripts/LeapPinchDetector.cs
```diff
@@ -157,6 +160,7 @@ public class LeapPinchDetector : MonoBehaviour {
 
       float pinchDistance = hand.PinchDistance * MM_TO_M;
       transform.rotation = hand.Basis.Rotation();
+      transform.position = hand.PalmPosition.ToVector3();
```
Amarcolina: This line does nothing, as the position is assigned to a couple lines below.  The transform is already correctly being assigned, so there doesn't need to be anything additional added here.

Amarcolina:
In both scenes the PinchDetector has been moved on the right hand but not the left.

Amarcolina:
Sorry, not both scenes, just the PinchDrawScene

protodeep:
...and done

PR #85, opened 2016-03-23
Interaction engine fix debug draw crash

ajohnston33:


ajohnston33:
Grr.  This was against master.

PR #88, opened 2016-03-23
Add Data/Render fps

hham:
Data fps and render fps are added in Leap_Hands_Demo_HMD scene.

### DIFF
--- a/Assets/LeapMotion/Scripts/LeapHandController.cs
+++ b/Assets/LeapMotion/Scripts/LeapHandController.cs
@@ -79,14 +79,11 @@ namespace Leap.Unity {
         prev_graphics_id_ = frame.Id;
 
       }
-			Debug.LogWarning("Update");
-
       float interp = Time.deltaTime / (0.5f + Time.deltaTime);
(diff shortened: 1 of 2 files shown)


Amarcolina:
We should create a separate FPS component to keep the responsibilities of HandController simple and contained.

jdonald:
I'll continue building off this branch for debugging. Sounds like we don't plan to merge it upstream, and will hold out for a separate FPS component?

It has some Debug prints and bad indentation left over here too :)

testmasterbright:
Seems like this would be easy to port over to the VR Visualizer, no?

hham:
Does VR Visualizer mean RiggedHandVisualizer?
Agree. This doesn't need to be merged. Let's just keep it in a separate branch.

testmasterbright:
VR Visualizer Unity project: https://github.com/leapmotion/UnityModules/tree/master/Assets/LeapMotionModules/VRVisualizer

hham:
Thank you for letting me know. It should be easy.
Closing Now.

Issue #67, opened 2016-03-10
Playing and stopping hangs Editor if service not running


protodeep:

veronicaz41:
If I play when the service is not running, my Unity stops responding.

PR #91, opened 2016-03-23
Fix fps Drop caused by lock

veronicaz41:
- [ ] fixes issue: FPS Drop in LeapC w/passthrough in Robust Mode or with no hand in view #5400 (on platform)
- [ ] no regression on issue: Editing a script in Unity while running a scene in the editor causes Unity to crash #5204 (on platform)
  @protodeep it kind of works even if I take out this piece of code: https://github.com/leapmotion/UnityModules/blob/develop/Assets/LeapMotion/Scripts/LeapServiceProvider.cs#L126-L131, however after Unity reload, the images still works, the tracking stops
- [ ] no regression on https://github.com/leapmotion/UnityModules/issues/67
  - Start LeapService, play a scene, Stop LeapService, stop the scene, it should not hang
  - Start LeapService, play a scene, Stop LeapService, Start LeapService, stop the scene, it should not hang.
   - Never start LeapService, only play a scene, then stop the scene, it should not hang (this originally works on Mac, but not on Windows)

- [ ] need to merge this to platform

### DIFF
--- a/Assets/LeapC/Connection.cs
+++ b/Assets/LeapC/Connection.cs
@@ -48,8 +48,8 @@ namespace LeapInternal
     private DistortionData _currentDistortionData = new DistortionData();
 
     private IntPtr _leapConnection;
+    private bool _isRunning = false;
     private Thread _polster;
-    private readonly object _connLocker = new object();
 
     //Policy and enabled features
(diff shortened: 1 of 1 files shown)


veronicaz41:
Retarget this to develop, closing it

PR #104, opened 2016-03-28
Fix bone rotations

jo3w4rd:
Modifies TestHandFactory to produce all right-handed bone basises.


jo3w4rd:
Barret, this will screw up the rigged hands again -- but I'm reasonably sure that the data is correct based on looking at the test hand factory position using the new debug hands. There must be something wrong with the bones in the model.

PR #110, opened 2016-03-30
Set a default position and rotation for the hands in the editor view

jo3w4rd:
Allows combined hand models to be lined up easier (at position 0,0,0 and no rotation) when added to a hand controller. Position is a compromise between the desktop and VR-style scenes -- must work for both.


jo3w4rd:
Closing until Adrian's Quaternion changes are merged.

PR #112, opened 2016-03-31
add soft contact support.

ajohnston33:


Amarcolina:
We should keep InteractionBehaviour agnostic to whether or not the object has a rigidbody or not, since there are going to be cases with and without a rigidbody.  InteractionBehaviour is the abstract base class so we can add a subclass with rigidbody support or something.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -283,6 +286,21 @@ public abstract class InteractionBehaviour : MonoBehaviour {
     }
 
     /// <summary>
+    /// Called by InteractionController when the velocity of an object is changed.
+    /// </summary>
+    public virtual void OnVelocityChanged(UnityEngine.Vector3 linearVelocity, UnityEngine.Vector3 angularVelocity)
```
Amarcolina: There is already a method stubbed out for this, called OnPush.  We can use the name OnVelocityChanged, but we should remove OnPush if that's going to be the case.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -297,6 +315,13 @@ public abstract class InteractionBehaviour : MonoBehaviour {
 
       _isRegisteredWithManager = true;
       _manager.RegisterInteractionBehaviour(this);
+
+      _rigidbody = GetComponent<Rigidbody>();
+      if (_rigidbody)
+      {
+        _rigidbodyHadUseGravity = _rigidbody.useGravity;
+        _rigidbody.useGravity = false;
```
Amarcolina: Is this because bullet is going to be accounting for the force of gravity?  If so we will need a way to set the bullet gravity force so that it matches the one in Unity.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionC/InteractionC.cs
```diff
@@ -398,6 +404,34 @@ public class InteractionC {
...
+      if (rs != eLeapIERS.eLeapIERS_Success || nVelocities == 0)
+      {
+        velocities = null;
+        return rs;
+      }
+
+      velocities = new LEAP_IE_VELOCITY[nVelocities];
+      for (int i = 0; i < nVelocities; i++)
+      {
+        velocities[i] = StructMarshal<LEAP_IE_VELOCITY>.ArrayElementToStruct(ppVelocitiesBuffer, i);
```
Amarcolina: is ppVelocitiesBuffer a pointer to an array of pointers?  If so, I think you would want StructMarshal<IntPtr> instead

ProjectSettings/DynamicsManager.asset
```diff
@@ -12,4 +12,4 @@ PhysicsManager:
   m_SolverIterationCount: 6
   m_QueriesHitTriggers: 1
   m_EnableAdaptiveForce: 0
-  m_LayerCollisionMatrix: ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```
Amarcolina: We like to not use layers in our core assets, since that can cause incompatibilities with developers projects.  We might need to brainstorm an alternate solution.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
ajohnston33: Yes.  I was going to talk to you about that.  This is somewhat placeholder.

ProjectSettings/DynamicsManager.asset
ajohnston33: I don't love this one, however as things are I need physics hands and I need them to not collide with interaction objects.  My thinking was that this was OK for the demo, and that we have to document this requirement.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionC/InteractionC.cs
ajohnston33: No, it is a pointer to a pointer to an array.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
ajohnston33: Until I get the gravity thing finalized "OnVelocityChanged" is more honest.

ProjectSettings/DynamicsManager.asset
Amarcolina: We could always have no physics hands in our demos, and use the documentation to suggest the layer strategy if the user wants to add physics hands.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionC/InteractionC.cs
Amarcolina: Ah got it, awesome.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
ajohnston33: Unfortunately I am not accumulating velocities, I have to clear *everything* and then set them.  There is some unfortunate stuff going on around working with PhysX.

PR #115, opened 2016-04-01
Feature interaction engine quaternions

ajohnston33:
This is still going to require testing.  I am assigning this to Alex to as for an early review.


ajohnston33:
Grr...

PR #116, opened 2016-04-01
initial draft moving bones to quaternions

ajohnston33:
This is still going to require testing. I am assigning this to Alex to as for an early review.


Assets/LeapC/LeapC.cs
```diff
@@ -388,24 +388,27 @@ public LEAP_VECTOR(UnityEngine.Vector3 unity)
...
-                             y_basis.ToLeapVector(),
-                             z_basis.ToLeapVector());
+      w = q.w;
+      x = q.x;
+      y = q.y;
+      z = q.z;
     }
 
-    public LEAP_MATRIX(Leap.Matrix leap)
+    public LEAP_QUATERNION(UnityEngine.Quaternion unity)
```
jo3w4rd: Do we need an if defined UnityEngine block here so that it compiles outside of Unity?

ajohnston33: Either that or I move that functionality into Unity specific code.

Assets/LeapC/LeapQuaternion.cs
```diff
@@ -0,0 +1,204 @@
...
+    public LeapQuaternion Multiply(LeapQuaternion rhs)
+    {
+      return new LeapQuaternion(
+        w * rhs.x + x * rhs.w + y * rhs.z - z * rhs.y,
+        w * rhs.y + y * rhs.w + z * rhs.x - x * rhs.z,
+        w * rhs.z + z * rhs.w + x * rhs.y - y * rhs.x,
+        w * rhs.w - x * rhs.x - y * rhs.y - z * rhs.z);
+    }
+
+    public static readonly LeapQuaternion Identity = new LeapQuaternion(0, 0, 0, 1);
```
Amarcolina: I thought that a quaternion with (0,0,0) for the xyz value was invalid?  Shouldn't it be something like (1,0,0,0) ?

PR #122, opened 2016-04-05
Fix problem with temporal warping disabling itself

### LINKED ISSUE #121: Hands can move when HMD changes camera angle
This is caused by the temporal warping being turned off, which happens because it is checking for device type before any device messages have been received from the service.

jo3w4rd:
Fixes #121 

What was happening was the DeviceList was empty when LeapVRTemporalWarping checked for device type, resulting in the check returning "DEVICE_INVALID" and LeapVRTemporalWarping turning itself off.

This change doesn't check for device type until a device is available and also handles unplugging and plugging in different types of device.

### DIFF
--- a/Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
+++ b/Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
@@ -182,13 +182,12 @@ namespace Leap.Unity{
     }
   
     protected void Start() {
-
       Controller controller = provider.GetLeapController();
       controller.Device += OnDevice;
     }
-    
+
     protected void OnDevice(object sender, DeviceEventArgs args) {
-       deviceInfo = provider.GetDeviceInfo();
+      deviceInfo = provider.GetDeviceInfo();
       if (deviceInfo.type == LeapDeviceType.Invalid) {
         Debug.LogWarning("Invalid Leap Device -> enabled = false");
         enabled = false;
@@ -199,16 +198,14 @@ namespace Leap.Unity{
       LeapVRCameraControl.OnValidCameraParams += onValidCameraParams;
    }
 
-    protected void OnEnable()
-    {
+    protected void OnEnable() {
       if (deviceInfo.type != LeapDeviceType.Invalid){
         LeapVRCameraControl.OnValidCameraParams -= onValidCameraParams; //avoid multiple subscription
         LeapVRCameraControl.OnValidCameraParams += onValidCameraParams;
       }
     }
 
-    protected void OnDisable()
-    {
+    protected void OnDisable() {
       LeapVRCameraControl.OnValidCameraParams -= onValidCameraParams;
     }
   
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -182,17 +182,28 @@ protected struct TransformData {
...
+    
+    protected void OnDevice(object sender, DeviceEventArgs args) {
+       deviceInfo = provider.GetDeviceInfo();
       if (deviceInfo.type == LeapDeviceType.Invalid) {
         Debug.LogWarning("Invalid Leap Device -> enabled = false");
         enabled = false;
         return;
       }
+      //Get a callback right as rendering begins for this frame so we can update the history and warping.
+      LeapVRCameraControl.OnValidCameraParams += onValidCameraParams;
```
Amarcolina: Subscribing here means that every time the user unplugs and replugs their device in while running the app, there will be an extra subscription added to OnValidCameraParams

jo3w4rd: Hmm, your right. It seems the easiest way to avoid double subscription is to unsubscribe before subscribing (since unsubscribing will be ignored if you aren't actually subscribed). Is that too hacky?

Leaving the subscribe in Start() means you would potentially be updating the temporal warping without knowing the device.

Amarcolina:
And now this PR is a great candidate to apply our newly adopted coding style :dancer:

jo3w4rd:
Doesn't it already conform? At least the changes... Or are you suggesting that we reformat files when we touch them for other reasons?

Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -182,17 +182,31 @@ protected struct TransformData {
...
-      LeapVRCameraControl.OnValidCameraParams += onValidCameraParams;
-  
-      deviceInfo = provider.GetDeviceInfo();
+
+      Controller controller = provider.GetLeapController();
+      controller.Device += OnDevice;
+    }
+    
+    protected void OnDevice(object sender, DeviceEventArgs args) {
+       deviceInfo = provider.GetDeviceInfo();
```
Amarcolina: extra space here.

Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -182,17 +182,31 @@ protected struct TransformData {
...
         enabled = false;
         return;
       }
+      //Get a callback right as rendering begins for this frame so we can update the history and warping.
+      LeapVRCameraControl.OnValidCameraParams -= onValidCameraParams; //avoid multiple subscription
+      LeapVRCameraControl.OnValidCameraParams += onValidCameraParams;
+   }
+
+    protected void OnEnable()
+    {
```
Amarcolina: Brace on next line here

Amarcolina:
Woohoo, looks great.  Merging in

PR #123, opened 2016-04-05
Merge quaternion C API from interaction engine

ajohnston33:


Assets/LeapC/LeapC.cs
```diff
@@ -366,33 +366,49 @@ public Leap.Vector ToLeapVector()
     {
       return new Leap.Vector(x, y, z);
     }
+
+    public UnityEngine.Vector3 ToUnityVector()
```
Amarcolina: Looks like these somehow made it into LeapC.  I don't think they are supposed to be here, since they reference UnityEngine.  We might want to add some extension methods for conversion (for Quaternion too)

Assets/LeapC/LeapQuaternion.cs
```diff
@@ -0,0 +1,175 @@
...
+      return obj is LeapQuaternion && Equals((LeapQuaternion)obj);
+    }
+
+    /**
+     * Returns true if all of the quaternion's components are finite.  If any
+     * component is NaN or infinite, then this returns false.
+     */
+    public bool IsValid()
+    {
+      return float.IsNaN(x) && float.IsNaN(y) && float.IsNaN(z) && float.IsNaN(w);
```
Amarcolina: This IsValid method doesn't work.  It only returns true if all 4 components are NaN.  Also IsNaN does not check for infinite values, you need to use float.IsInfinity for that.

ajohnston33: Funny story.  I copied that function from our Vector class which was also wrong.  Both are fixed now.

Assets/LeapC/LeapC.cs
ajohnston33: Thanks for catching that.

Assets/LeapC/LeapQuaternion.cs
Amarcolina: Yeahh a PR just went through on platform that fixed the Vector method.  PR #5504 if you are interested.

ajohnston33:
Joe has merged a subset of these changes back to platform.  They tested OK there.

Amarcolina:
Cool, and these seem to be working great on my end as well.  I'm gonna merge this in if there are no objections.

Issue #98, opened 2016-03-24
Add FPS counter to the Unity VR Visualizer


testmasterbright:
##### Description:
Hyoung recently added a Data FPS and Render FPS to the SamsungVR demo (https://github.com/leapmotion/UnityModules/pull/88). This should be easily ported to the VR Visualizer.

digitalweaver:
@protodeep This is the Data FPS we talked about last week that you mentioned you could get into the 1.3 visualizer this week. Does that still work for you?

PR #126, opened 2016-04-06
Adding vr visualizer fps labels

protodeep:
Adding counters for VRVisualizer's render framerate and Leap data framerate for issue #98


mmenicutch:
We are seeing some issues with the new VR visualizer. It works correctly in the editor but when it's built we are seeing a few different issues - for me the hand ID constantly changes and none of the information is displayed, it also runs at an extremely slow framerate. When Matt tried it on 1.3. oculus he saw the same issue with no information being displayed and the hand ID constantly changing, but it did not run slowly and instead crashed after about 10 seconds of use (The controller disconnected screen appears and then the app crashes) after that he was unable to relaunch the app (it launched to a gray screen.) 

It worked in the unity editor though and we also noticed that when running it in the unity editor the new data is not displayed in the oculus.

Assets/LeapMotionModules/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
```diff
@@ -94,6 +97,21 @@ void Update()
             m_controller.SetPolicy(Controller.PolicyFlag.POLICY_OPTIMIZE_HMD);
         }
       }
+      //update render frame display
+      m_deltaTime += (Time.deltaTime - m_deltaTime) * 0.1f;
```
Amarcolina: We should be using SmoothedFloat for all smoothing related needs, since SmoothedFloat is framerate independent, while traditional exponential smoothing is not.

Amarcolina:
Everything looks good!

testmasterbright:
were the issues that mike mentioned addressed?

PR #130, opened 2016-04-06
integate physics api changes

ajohnston33:


Amarcolina:
Everything looks mostly good.  Getting an IsOrthogonal error when touching a cube, but we can address that in a different PR

PR #134, opened 2016-04-11
Set child primitives to current layer on creation.

### LINKED ISSUE #133: CapsuleHands not respecting layer setting
When a capsule hand is created, the child primitives it creates are not set to use the current CapsuleHand layer setting.

jo3w4rd:
Fixes #133


Amarcolina:
Ah that's a nice catch!

PR #137, opened 2016-04-12
LeapImageRetriever is destroyed when the device is lost. This change …

jo3w4rd:
Somewhere, the LeapImageRetriever component is destroyed when a device is lost and then recreated. This change cleans up the LeapImageRetriever event subscriptions so that the event handlers of the destroyed object are not invoked.


jo3w4rd:
Superseded.

PR #140, opened 2016-04-13
Feature hands module

protodeep:


Assets/LeapMotion/Editor/LeapHandPoolEditor.cs
```diff
@@ -0,0 +1,19 @@
+﻿using UnityEngine;
+using UnityEditor;
+using System.Collections;
+
+namespace Leap.Unity {
+  [CustomEditor(typeof(HandPool))]
+  public class LeapHandPoolEditor : Editor {
```
Amarcolina: We should have this extend CustomEditorBase, which can help reduce errors and also help mimic the custom inspector more reliably.

Assets/LeapMotion/Editor/LeapModelPairList.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections;
+using UnityEditor;
+using UnityEngine;
+
+public static class LeapModelPairList {
+	
+	public static void Show (SerializedProperty list) {
```
Amarcolina: This looks like it should be a method that belongs in LeapHandPoolEditor, instead of being in a separate class.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -17,53 +17,94 @@ public class HandPool :
...
+          Chirality handChirality = hand.IsRight ? Chirality.Right : Chirality.Left;
+          isCorrectHandedness = model.Handedness == handChirality;
+          if (!EnforceHandedness || model.Handedness == Chirality.Either) {
+            isCorrectHandedness = true;
+          }
+          bool isCorrectModelType;
+          isCorrectModelType = model.HandModelType == modelType;
+          if (isCorrectModelType && isCorrectHandedness) {
+            group.modelList.RemoveAt(i);
+            --i;
```
Amarcolina: This line decrementing i does nothing, since we break out of the loop right after.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -17,53 +17,94 @@ public class HandPool :
...
+            models.Add(model);
             break;
+          }
         }
       }
+      if (models != null) {
```
Amarcolina: This check will always pass, since the list is always constructed and even if it is empty, it will always be non-null.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -17,53 +17,94 @@ public class HandPool :
...
+        ModelGroup newModelGroup = new ModelGroup(new List<IHandModel>());
+        newModelGroup.modelList.Add(pair.LeftModel);
+        modelGroupMapping.Add(pair.LeftModel, newModelGroup);
+        newModelGroup.modelList.Add(pair.RightModel);
+        modelGroupMapping.Add(pair.RightModel, newModelGroup);
+        ModelPool.Add(newModelGroup);
       }
       controller_ = GetComponent<LeapHandController>();
     }
+    void Update() {
```
Amarcolina: This Update method is empty.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -17,53 +17,94 @@ public class HandPool :
     HandFactory {
 
     [SerializeField]
-    private List<IHandModel> ModelCollection;
-    public List<IHandModel> ModelPool;
+    private List<ModelPair> ModelCollection;
+    public List<ModelGroup> ModelPool;
     public LeapHandController controller_ { get; set; }
```
Amarcolina: This variable is not used and should be removed.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -17,53 +17,94 @@ public class HandPool :
     HandFactory {
 
     [SerializeField]
-    private List<IHandModel> ModelCollection;
-    public List<IHandModel> ModelPool;
+    private List<ModelPair> ModelCollection;
+    public List<ModelGroup> ModelPool;
```
Amarcolina: ModelPool should be made private (or projected) with a SerializeField attribute, since it is not something we want other scripts being able to mess with.

Amarcolina:
Looks like there are some conflicts as well.

protodeep:
Addressed all the comments above.

PR #141, opened 2016-04-13
Added public properties for RTS PinchDetector properties

jo3w4rd:
This change adds public properties for the PinchDetector objects used by the RTS script. This allows RTS-using components to be made into prefabs. Otherwise, the connections to the pinch detectors are lost when the object is made into a prefab and, with only private variables, the connection cannot be restored at runtime.


Amarcolina:
The awake method should also be changed to the Start method instead.  Awake is called right when a prefab is instantiated, so the warnings will be thrown before the user has a chance to set the values.

Amarcolina:
Cool, looks good to me!

PR #151, opened 2016-04-19
Feature enable disable hand pairs

protodeep:


Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -90,6 +102,21 @@ public class ModelGroup {
...
     }
 
+    public void EnableDisablePair(string groupName, bool isEnabled) {
+      for (int i = 0; i < modelGroupMapping.Count; i++) {
+        ModelGroup group = modelGroupMapping.ElementAt(i).Value;
+        if (group.GroupName == groupName) {
+          IHandModel model = modelGroupMapping.ElementAt(i).Key;
+          if (model.IsTracked) {
+            model.InitHand();
+            model.transform.gameObject.SetActive(isEnabled);
```
Amarcolina: We should not be enabling or disabling hand models by explicitly calling SetActive(), that makes an assumption as to what the game object wants to do.  We should follow the IHandModel lifecycle that we have defined (using BeginHand/UpdateHand/FinishHand).

protodeep:
OK. Done.  Agreed, that's cleaner.  Thanks!

PR #154, opened 2016-04-21
Feature interaction engine tune

Amarcolina:


Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -42,9 +42,15 @@ public class InteractionManager : MonoBehaviour {
     [SerializeField]
     protected float _untrackedTimeout = 0.5f;
 
+    // TODO: eLeapIESceneInfoFlags_PhysicsEnabled
     [Tooltip("Allow the Interaction Engine to modify object velocities when pushing.")]
     [SerializeField]
-    protected bool _enablePushing = true;
+    protected bool _enableContact = true;
```
Amarcolina: I would recommend having the name of this property match whatever name we use in InteractionBehaviour.  I don't care as much what the name is, as long as it matches.

PR #165, opened 2016-04-26
Feature enable disable hand pairs

protodeep:
- [ ] Use "O" & "P" keys to disable and enable Poly_Hands


Assets/LeapMotion/Scripts/HandProxy.cs
```diff
@@ -44,11 +34,38 @@ public class HandProxy:
...
+
+    public override void AddModel(IHandModel model) {
+      if (handModels == null) {
+        handModels = new List<IHandModel>();
+      }
+      handModels.Add(model);
+      if (model.GetLeapHand() == null) {
+        model.SetLeapHand(MostRecentHand);
+        model.InitHand();
+        model.UpdateHand();
```
Amarcolina: There is a subtle bug here.  When Init is called, Update is called right after, even though Begin has not been called yet.  Begin should always be called before Update is called.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -15,37 +16,61 @@ namespace Leap.Unity {
...
-
+    public bool EnforceHandedness = false;
     [SerializeField]
     private List<ModelPair> ModelCollection;
     [SerializeField]
-    private List<ModelGroup> ModelPool;
-    public bool EnforceHandedness = false;
-    
+    public List<ModelGroup> ModelPool;
+    public List<HandRepresentation> ActiveHandReps = new List<HandRepresentation>();
```
Amarcolina: These variables should not be publicly modifiable, since any changes could disrupt the entire pooling mechanism.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -15,37 +16,61 @@ namespace Leap.Unity {
    */
   public class HandPool :
     HandFactory {
-
+    public bool EnforceHandedness = false;
```
Amarcolina: should be private with a SerializeField, if the value needs to be accessed outside this class, a separate public getter can be added.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -59,35 +84,63 @@ public class ModelGroup {
...
+        if (ModelPool[i].GroupName == groupName) {
+          group = ModelPool[i];
+          for (int m = 0; m < group.modelsCheckedOut.Count; m++) {
+            IHandModel model = group.modelsCheckedOut[m];
+            HandProxy handProxy;
+            if (ModelToHandProxyMapping.TryGetValue(model, out handProxy)) {
+              handProxy.RemoveModel(model);
+            }
+          }
+          group.modelsCheckedOut = new List<IHandModel>();
```
Amarcolina: Should clear the list here instead of creating a new one from scratch.  Optionally a better solution might be to have a method ReturnToGroup inside of ModelGroup that returns a model to the group.  This method would take care of removing the model from the checked out list.

This line could be an assert that asserts that checked out list is empty, since it always should be after returning all models to the group.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -59,35 +84,63 @@ public class ModelGroup {
...
+          group.IsEnabled = false;
+        }
+      }
+    }
+    public void ReturnToPool(IHandModel model) {
       ModelGroup modelGroup = modelGroupMapping[model];
```
Amarcolina: The model is removed from modelsCheckedOut in this case, resulting in the list growing by one every time a new hand comes into the frame but never getting smaller.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -59,35 +84,63 @@ public class ModelGroup {
...
-            break;
+      Chirality handChirality = hand.IsRight ? Chirality.Right : Chirality.Left;
+      HandRepresentation handRep = new HandProxy(this, hand, handChirality, modelType);
+      for (int i = 0; i < ModelPool.Count; i++) {
+        ModelGroup group = ModelPool[i];
+        if (group.IsEnabled) {
+          IHandModel model = group.TryGetModel(handChirality, modelType);
+          if (model != null) {
+            handRep.AddModel(model);
+            group.modelsCheckedOut.Add(model);
```
Amarcolina: Instead of forcing HandPool to manually take care of the logic when checking out models, TryGetModel should handle this logic automatically to prevent error.  There might be a matching ReturnModel method as well that automatically removes it from the modelsCheckOut list.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -59,35 +84,63 @@ public class ModelGroup {
      * @param hand The Leap Hand data to be drive an IHandModel
      * @param modelType Filters for a type of hand model, for example, physics or graphics hands.
      */
+
     public override HandRepresentation MakeHandRepresentation(Hand hand, ModelType modelType) {
-      HandRepresentation handRep = null;
+      //HandRepresentation handRep = null;
       List<IHandModel> models = new List<IHandModel>();
```
Amarcolina: There is a list here that isn't used for anything

protodeep:
Alright, think I've addressed the comments.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -59,35 +93,62 @@ public class ModelGroup {
...
+          for (int m = 0; m < group.modelsCheckedOut.Count; m++) {
+            IHandModel model = group.modelsCheckedOut[m];
+            HandRepresentation handRep;
+            if (modelToHandRepMapping.TryGetValue(model, out handRep)) {
+              handRep.RemoveModel(model);
+              group.ReturnToGroup(model);
+              m--;
+            }
+          }
+          Debug.Assert(group.modelsCheckedOut.Count == 0, group.GroupName + "'s modelsCheckedOut List has not been cleared");
```
Amarcolina: You can use the UnityEngine.Assertions namespace for greater control over asserts.  They allow the user to be able to turn asserts off (where they are compiled out), allow them to be logged to console, or allow them to throw exceptions.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -15,37 +16,70 @@ namespace Leap.Unity {
...
-
+    [SerializeField]
+    private bool EnforceHandedness;
     [SerializeField]
     private List<ModelPair> ModelCollection;
     [SerializeField]
     private List<ModelGroup> ModelPool;
-    public bool EnforceHandedness = false;
-    
+    public List<HandRepresentation> ActiveHandReps = new List<HandRepresentation>();
```
Amarcolina: This variable is still public and should not be, since it should not be able to be modified by outside forces.

PR #169, opened 2016-04-28
Feature allow spawning

protodeep:


Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -147,9 +162,10 @@ public class ModelGroup {
       }
     }
     public void ReturnToPool(IHandModel model) {
-      ModelGroup modelGroup = modelGroupMapping[model];
-      modelGroup.ReturnToGroup(model);
-      //Todo: add check to see if representation of chirality and type exists
+      ModelGroup modelGroup;
+      if(modelGroupMapping.TryGetValue(model, out modelGroup)){
```
Amarcolina: Is there any case where this can fail?  If not, we should add an assert.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -60,6 +60,17 @@ public class ModelGroup {
             return model;
           }
         }
+        //Todo: if spawning enabled
+        for (int i = 0; i < modelsCheckedOut.Count; i++) {
+          if (modelsCheckedOut[i].Handedness == chirality && modelsCheckedOut[i].HandModelType == modelType) {
+            IHandModel modelToSpawn = modelsCheckedOut[i];
+            GameObject spawnedGO = GameObject.Instantiate(modelToSpawn.gameObject);
```
Amarcolina: You can actually just instantiate the hand model directly and it will automatically instantiate the game object it is on and return the new hand model.

PR #170, opened 2016-04-29
Feature allow spawning

protodeep:
- Chirality handling is now controlled exclusively by IHandModel's Handedness enum and EnforceHandedness has been removed.
- Runtime copying is now controlled based on a public bool CanDuplicate per ModelPair


Amarcolina:
ModelGroup has a second constructor hidden within the variable definitions that isn't used.

Amarcolina:
ModelPair is a serialized class specifically for usage in the inspector, so it doesn't need explicit constructors.

Amarcolina:
ActiveHandReps is public and should be made non-public

Amarcolina:
Because ModelPair is separate from ModelGroup, there is a non-intuitive result when playing.  I can check and uncheck the IsEnabled checkbox on a pair and nothing happens.  

There are a number of possible solutions, but one might be to combine ModelPair and ModelGroup to be the same class.  There already is a 1to1 mapping from the set of pairs to the set of groups, so it might not be terrible to implement.

Amarcolina:
We should also remove the enable/disable key helpers from HandPool as well, since we don't want those to be in a release.

Amarcolina:
The method ValidateIHandModelPrefab is no longer used and should be removed.

Amarcolina:
The pattern:

If(condition)
var = true
else
var = false

can be simplified to 
var = condition

This can be utilized in OnValidate()

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -51,23 +54,37 @@ public class ModelGroup {
...
+        this.CanDuplicate = canDuplicate;
+        this.modelList = modelList;
+        this.modelsCheckedOut = new List<IHandModel>();
+        this._handPool = handPool;
+      }
+
       public IHandModel TryGetModel(Chirality chirality, ModelType modelType) {
         for (int i = 0; i < modelList.Count; i++) {
-          if (modelList[i].Handedness == chirality && modelList[i].HandModelType == modelType) {
+          if (modelList[i].Handedness == chirality && modelList[i].HandModelType == modelType
```
Amarcolina: This conditional can be simplified to be
modelList[i].HandModelType == modelType && 
(modelList[i].Handedness == chirality || modelList[i].Handedness == Chirality.Either)

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -51,23 +54,37 @@ public class ModelGroup {
...
-        for (int i = 0; i < modelsCheckedOut.Count; i++) {
-          if (modelsCheckedOut[i].Handedness == chirality && modelsCheckedOut[i].HandModelType == modelType) {
-            IHandModel modelToSpawn = modelsCheckedOut[i];
-            IHandModel spawnedModel = GameObject.Instantiate(modelToSpawn);
-            _handPool.modelGroupMapping.Add(spawnedModel, this);
-            modelsCheckedOut.Add(spawnedModel);
-            return spawnedModel;
+        if (CanDuplicate) {
+          for (int i = 0; i < modelsCheckedOut.Count; i++) {
+            if (modelsCheckedOut[i].Handedness == chirality && modelsCheckedOut[i].HandModelType == modelType
```
Amarcolina: Same check here can be simplified (might also be worthwhile to make into a helper function since it is used more than once)

Amarcolina:
Also make sure that you have downloaded the code formatting file from the coding conventions thread, there are some brackets that do not match up (especially near else clauses).  You can press ctrl-K then ctrl-D to auto-format a file.

PR #174, opened 2016-04-30
Low poly capsule hand upgrade

Amarcolina:
This upgrades the capsule hand script to have qualities that improve performance.  The user is now able to specify the resolution of the cylinders (which are no longer capsules!) and they are able to specify the mesh to use on the joints.  

A medium and low poly sphere fbx model has been included for convenience, and the medium poly sphere is being used by default.  With the cylinder resolution set to 8 and using the medium poly spheres, dynamic batching of the hands is now possible!

 - [ ] Verify that hands still work and render properly in all scene.
 - [ ] Make sure that the prefabs still work properly when dragged into a fresh scene.
 - [ ] Verify that creating a new capsule hand from scratch doesn't give any issues.


protodeep:
When attaching a Capsule Hand script to an empty GameObject, the hand appears pink and without spheres.  After playback and stopping, the new Hand Appears normal.

protodeep:
Intersections between low poly cylinders and low poly spheres don't look as good as before.  Should we make the defaults higher poly for typical VR scenes?

PR #200, opened 2016-05-04
Removing DebugFinger components

protodeep:
Verify that DebugHand works


jo3w4rd:
You can remove the child transforms, too.

protodeep:
Done

PR #213, opened 2016-05-11
Interaction improvements for velocity grasp

Amarcolina:
- Pulled in physics interpolation from PR on develop
- Moved simulation step to FixedUpdate instead of Coroutine (was causing 1 frame latency!)
- Added temporal warping to interaction behaviour for both kinematic and velocity based grasp
- Added warp curve to disable warping when the object is at a distance
- Added a lever and a constrained box to the example scene
- Turned on shadows
- Overrode gravity to be -5m/s^2 instead of normal earth gravity

### DIFF
--- a/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
+++ b/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
@@ -287,9 +287,7 @@ namespace Leap.Unity {
 
           warpedRotation = warpedRotation * transform.localRotation;
 
-          Vector scale = new Vector(transform.lossyScale.x * 1e-3f, transform.lossyScale.y * 1e-3f, transform.lossyScale.z * 1e-3f);
-
-          leapTransform = new LeapTransform(warpedPosition.ToVector(), warpedRotation.ToLeapQuaternion(), scale);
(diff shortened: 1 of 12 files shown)


Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -253,7 +279,23 @@ public class LeapServiceProvider : LeapProvider {
...
-        toUpdate = source.TransformedCopy(transform.GetLeapMatrix());
+        LeapTransform leapTransform;
+        if (_temporalWarping != null) {
+          Vector3 warpedPosition;
+          Quaternion warpedRotation;
+          _temporalWarping.TryGetWarpedTransform(LeapVRTemporalWarping.WarpedAnchor.CENTER, out warpedPosition, out warpedRotation, source.Timestamp);
+
+          warpedRotation = warpedRotation * transform.localRotation;
+
+          Vector scale = new Vector(transform.lossyScale.x * 1e-3f, transform.lossyScale.y * 1e-3f, transform.lossyScale.z * 1e-3f);
```
ajohnston33: Function call please.

PR #216, opened 2016-05-12
Interaction engine feature improvements

Amarcolina:
- Removed hacked layer code in Interaction Behaviour
- Added auto-layer generation to the Interaction Manager (off by default to force user to setup layers)
- Added the ability for interaction objects to allow or disallow suspension.
- Added a SingleLayer property for use with the layer settings.
- InteractionBehaviour sets drag to zero when grasp occurs, sets it back when ungrasped
- Added events to the base interaction behaviour


ajohnston33:
You really committed a removed tab with the description "Removed tab".  I have to up my game.

PR #218, opened 2016-05-13
feature-ui-input-module

jselstad:
Basic Input Functionality, Prefabs, and Example Scene


Amarcolina:
We should not have any scenes in the build settings, currently two scenes from PinchUtility are there.

Amarcolina:
We should also not be using any layers / tags, since users will not be importing tag settings when they import module packages.

Amarcolina:
Looks like you have an empty Assets/Editor folder that is leaving a meta file around.

Assets/LeapMotionModules/UIInput/Scripts/LeapInputModule.cs
```diff
@@ -20,15 +20,22 @@ public class LeapInputModule : BaseInputModule {
     [Tooltip("An optional alternate detector for pinching on the right hand.")]
     public Leap.Unity.PinchUtility.LeapPinchDetector RightHandDetector;
     [Tooltip("How many hands and pointers the Input Module should allocate for.")]
-    public int NumberOfHands = 2;
+    int NumberOfHands = 2;
```
Amarcolina: NumberOfHands was made non-public, was this intentional?  If so, it should probably be a constant unless it is expected to mutate.

Amarcolina:
So this is looking all good for a merge, but I see that the most recent commit has the words 'first pass' in it, do you want me to hold off?

Amarcolina:
Ok everything looks good to me!  Merging in!

PR #223, opened 2016-05-17
Feature attachments and detectors

jo3w4rd:
Detectors go into core, with examples in a DetectionExamples "module". This PR supersedes the PinchUtilities module, which still exists, but we should deprecate and remove after one more release.

Attachments go into Attachments module. Attachments still need an real example scene (the existing ones are just wip testing).


Assets/LeapMotion/Scenes/Leap_Hands_Demo_Desktop.unity
```diff
@@ -186,6 +186,19 @@ Prefab:
       value: 1
       objectReference: {fileID: 0}
     - target: {fileID: 11406336, guid: b9c7aaf0d314cb847a3b2a20e893f1b6, type: 2}
+<<<<<<< HEAD
```
Amarcolina: There is a merge conflict in this scene file

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
```diff
@@ -0,0 +1,53 @@
...
+      if (IsActive) {
+        IsActive = false;
+        OnDeactivate.Invoke();
+      }
+      IsActive = false;
+    }
+  }
+
+  /** 
+  * Settings for handling pointing conditions
```
Amarcolina: Does this belong in the base class? I don't think it would apply in all cases for extending detectors.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
```diff
@@ -0,0 +1,122 @@
...
+        handModel = gameObject.GetComponentInParent<IHandModel>();
+      }
+    }
+
+    private void OnEnable () {
+      StartCoroutine(fingerPointingWatcher());
+    }
+  
+    private void OnDisable () {
+      StopCoroutine(fingerPointingWatcher());
```
Amarcolina: Calling fingerPointingWatcher() creates a new IEnumerator which is different than the one used to start the coroutine, so this method will do nothing.  You could save the IEnumerator or use StopAllCoroutines, or just start the coroutine in Start and never stop it.

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -29,5 +29,60 @@ public static class Utils {
...
+    /* Visual debugging utilities */
+    public static void DrawCircle(Vector3 center, 
+                           Vector3 normal, 
+                           float radius, 
+                           Color color, 
+                           int quality = 32, 
+                           float duration = 0, 
+                           bool depthTest = true) {
+      Vector3 planeA = Vector3.Slerp(normal, -normal, 0.5f);
+      Vector3 planeB = Vector3.Cross(normal, planeA).normalized;
```
Amarcolina: This variable is not used.

Assets/LeapMotionModules/Attachments/Scripts/Transition.cs
```diff
@@ -0,0 +1,115 @@
...
+        localPosition.z = ZPosition.Evaluate(interpolationPoint) * RelativeOnPosition.z;
+        transform.localPosition = localPosition;
+      }
+      if(AnimateRotation){
+        Quaternion transitionRotation = Quaternion.Euler(transform.localRotation.x + XRotation.Evaluate(interpolationPoint) * RelativeOnRotation.x,
+                                                         transform.localRotation.y + YRotation.Evaluate(interpolationPoint) * RelativeOnRotation.y,
+                                                         transform.localRotation.z + ZRotation.Evaluate(interpolationPoint) * RelativeOnRotation.z);
+        transform.localRotation = transitionRotation;
+      }
+      if (AnimateRotation && false) {
```
Amarcolina: Impossible condition

Amarcolina:
The pinch utility should probably be removed so that things don't get mixed up.  Currently we have a duplicate component for everything that was in that module, and DetectionExample scenes are referencing the wrong script in several places.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
```diff
@@ -0,0 +1,53 @@
...
+    /**
+    * Invoked when this detector deactivates.
+    * Subclasses must call this function when the detector's conditions change from true to false.
+    */
+    public virtual void Deactivate(){
+      if (IsActive) {
+        IsActive = false;
+        OnDeactivate.Invoke();
+      }
+      IsActive = false;
```
Amarcolina: This statement is unneeded.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
```diff
@@ -0,0 +1,53 @@
...
+    /**
+    * Invoked when this detector activates.
+    * Subclasses must call this function when the detector's conditions become true.
+    */
+    public virtual void Activate(){
+      if (!IsActive) {
+        IsActive = true;
+        OnActivate.Invoke();
+      }
+      IsActive = true;
```
Amarcolina: Also unneeded.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
```diff
@@ -0,0 +1,53 @@
+﻿using UnityEngine;
+using UnityEngine.Events;
+using System.Collections;
+using Leap;
+
+namespace Leap.Unity {
+
+  public class Detector : MonoBehaviour {
+    public bool IsActive{ get{ return _isActive;} private set { _isActive = value;}}
```
Amarcolina: Having a private setter in this case isn't needed since there already is a way to set the variable privately.  Having a setter also might open the door for potential bugs if additional code is later added to the setter unintentionally.

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
```diff
@@ -0,0 +1,79 @@
...
+    public void RemoveDetector(Detector detector){
+      detector.OnActivate.RemoveListener(CheckDetectors);
+      detector.OnDeactivate.RemoveListener(CheckDetectors);
+      Detectors.Remove(detector);
+    }
+
+    public void AddAllDetectors(){
+      Detector[] detectors = GetComponents<Detector>();
+      for(int g = 0; g < detectors.Length; g++){
+        if ( detectors[g].GetInstanceID() != this.GetInstanceID() && detectors[g].enabled) {
```
Amarcolina: any reason you are using instanceID instead of regular equality?

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
```diff
@@ -0,0 +1,79 @@
...
+          AddDetector(detectors[g]);
```
Amarcolina: We should check if the detector is already added, since a user might easily add a sibling detector component to the list.

Amarcolina: Ah I see, that is done inside the AddDetector method, nevermind!

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
```diff
@@ -0,0 +1,79 @@
+﻿using UnityEngine;
+using UnityEngine.Events;
+using System.Collections.Generic;
+
+namespace Leap.Unity {
+
+  public class DetectorLogicGate : Detector {
+    public List<Detector> Detectors;
```
Amarcolina: This list should not be public, since modification would cause incorrect behaviour.

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
```diff
@@ -0,0 +1,79 @@
...
+    public bool AddAllDetectorsOnAwake = true;
```
Amarcolina: Maybe rename to AddSiblingDetectorsOnAwake instead?  It's not immediately obvious what 'All' refers to.

Assets/LeapMotion/Scripts/DetectionUtilities/ExtendedFingerDetector.cs
```diff
@@ -0,0 +1,110 @@
...
+  
+    public PointingState Thumb = PointingState.Either;
+    public PointingState Index = PointingState.Either;
+    public PointingState Middle = PointingState.Either;
+    public PointingState Ring = PointingState.Either;
+    public PointingState Pinky = PointingState.Either;
+  
+    void Start () {
+      if(handModel == null){
+        handModel = gameObject.GetComponentInParent<IHandModel>();
```
Amarcolina: Start happens after OnEnable, so if the hand model is not assigned, the coroutine will give an error since it tries to access handModel which might be null.

Assets/LeapMotion/Scripts/DetectionUtilities/ExtendedFingerDetector.cs
```diff
@@ -0,0 +1,110 @@
...
+          }
+        } else if(IsActive){
+          Deactivate();
+        }
+        yield return new WaitForSeconds(Period);
+      }
+    }
+  
+    private bool matchFingerState(Finger finger, int ordinal){
+      switch(ordinal){
```
Amarcolina: This could be simplified so that the switch only selects which PointingState to use, and then the conditional logic doesn't have to be repeated.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
```diff
@@ -0,0 +1,122 @@
...
+      while(true){
+        if(handModel != null){
+          hand = handModel.GetLeapHand();
+          if(hand != null){
+            targetDirection = selectedDirection(hand.Fingers[selectedFinger].TipPosition.ToVector3());
+            fingerDirection = hand.Fingers[selectedFinger].Direction.ToVector3();
+            float angleTo = Vector3.Angle(fingerDirection, targetDirection);
+            if(handModel.IsTracked && angleTo <= OnAngle){
+              Activate();
+            } else if (!handModel.IsTracked || angleTo >= OffAngle) {
```
Amarcolina: the elseif is not needed, just an else would have the same effect.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
```diff
@@ -0,0 +1,122 @@
...
+        case PointingType.RelativeToWorld:
+          return PointingDirection;
+        case PointingType.AtTarget:
+          return TargetObject.position - tipPosition;
+        default:
+          return PointingDirection;
+      }
+    }
+
+    private int selectedFingerOrdinal(){
```
Amarcolina: These are hardcoded indexes into the finger array.  Is that order ensured to be the same?  And if so, we should provide a way to use the enum to index the array within our Finger class instead of a strange switch statement over here.

Assets/LeapMotion/Scripts/DetectionUtilities/PalmDirectionDetector.cs
```diff
@@ -0,0 +1,107 @@
...
+      Vector3 normal;
+      while(true){
+        if(handModel != null){
+          hand = handModel.GetLeapHand();
+          if(hand != null){
+            normal = hand.PalmNormal.ToVector3();
+            float angleTo = Vector3.Angle(normal, selectedDirection(hand.PalmPosition.ToVector3()));
+            if(angleTo <= OnAngle){
+              Activate();
+            } else if(angleTo > OffAngle) {
```
Amarcolina: Unneeded elseif, else is fine.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
jo3w4rd: removed.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
jo3w4rd: removed

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
jo3w4rd: removed.

Assets/LeapMotion/Scripts/DetectionUtilities/Detector.cs
jo3w4rd: Moved to palm pointing detector.

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
jo3w4rd: privatized.

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
jo3w4rd: renamed

Assets/LeapMotion/Scripts/DetectionUtilities/ExtendedFingerDetector.cs
jo3w4rd: Changed to Awake.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
jo3w4rd: Odd, I thought I had switched them all over to saving a reference to the IEnumerator... Well, they do now.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
jo3w4rd: I don't think so -- I don't want to deactivate until angleTo becomes grater than OffAngle,  not just when angleTo is no longer less than or equal to OnAngle.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
jo3w4rd: I'd still have to write a function to get the ordinal because the FingerType has a TYPE_UNKNOWN, which I can't use.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
Amarcolina: Ah yes of course, sorry missed the OnAngle being different than OffAngle :)

Assets/LeapMotion/Scripts/DetectionUtilities/PalmDirectionDetector.cs
jo3w4rd: As above, I think it is needed.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
Amarcolina: But currently TYPE_UNKNOWN can be used by this switch statement, and defaults to the index finger.

Assets/LeapMotion/Scripts/DetectionUtilities/PalmDirectionDetector.cs
Amarcolina: Yup, you are very correct!

Assets/LeapMotion/Scripts/DetectionUtilities/ExtendedFingerDetector.cs
jo3w4rd: Simplified.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
jo3w4rd: Maybe I'm misunderstanding your suggestion... I can't use the values of the selected enum member to index the array even though they match up with the finger array position, because TYPE_UNKNOWN is -1. Were you suggesting a change to the Finger class?

Assets/LeapMotion/Scenes/Leap_Hands_Demo_Desktop.unity
jo3w4rd: deconflicted.

Assets/LeapMotion/Scripts/DetectionUtilities/DetectorLogicGate.cs
jo3w4rd: Probably no good reason. Changed.

Assets/LeapMotion/Scripts/Utils/Utils.cs
jo3w4rd: ANd removed.

Assets/LeapMotionModules/Attachments/Scripts/Transition.cs
jo3w4rd: Removed rejected alternative calculation.

Assets/LeapMotion/Scripts/DetectionUtilities/FingerDirectionDetector.cs
Amarcolina: Yeah I am proposing that we add a method to Hand that takes in a finger type and returns the finger of that type (or null / error if you pass in type unknown).  The hand class could use the enum value as an index, just with some extra checks for unknown enum types.  That way this functionality (which is quite useful in all cases!) can be used all over the place.

jo3w4rd: That will have to be a different PR.

jo3w4rd:
@Amarcolina I think I've addressed all your current comments.

Amarcolina:
Everything works great!

PR #237, opened 2016-05-19
Fix model group toggle

protodeep:


jo3w4rd:
Did you really mean to include all these changes?

PR #241, opened 2016-05-20
Fix vive position

jo3w4rd:
Seems tacky, but now adjusting the position of the world to compensate for initial shift in head position.

(Also fixes a null object bug in PinchDetector)


jselstad:
Short fix works some of the time

If the user starts the application with his Vive sitting on the floor, there is no way to get at any of the scene components; the "Press R to Recenter" functionality of the Timewarping script doesn't do anything.  Also the script isn't incorporated into all of the scenes

My suggestion is to incorporate the coordinate space conversion into the Recenter button in the Timewarping script to move the transform.parent.parent in such a way that the HMD matches its original position before the button was pressed IF the user is in 5.4 or above and OpenVR was detected as the HMD.

Since the way HMDs are queried was changed from 5.3 to 5.4 (where Vive support was added), you'll have to use a platform define like: 
```
      if (Input.GetKeyDown(recenter) && VRSettings.enabled && VRDevice.isPresent) {
        InputTracking.Recenter();
#if UNITY_5_4_OR_NEWER
        if(VRDevice.model.Equals("OpenVR"){
                *reset transform.parent.parent*
        }
#endif
      }
```
or something.

Some docs:
http://docs.unity3d.com/540/Documentation/Manual/PlatformDependentCompilation.html
http://docs.unity3d.com/540/Documentation/ScriptReference/VR.VRDevice-model.html

jo3w4rd:
John has a better fix pending.

PR #244, opened 2016-05-23
Feature vr floor offset

Amarcolina:
Re-Adds the VRHeightOffset script.  This script offsets a transform based on the vr device that has been loaded.  By default, when oculus is loaded it offsets upwards by about a meter.  Also updates the default camera rig prefab.


Assets/LeapMotion/Scripts/VR/VRHeightOffset.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Linq;
+
+public class VRHeightOffset : MonoBehaviour {
+
+  [Serializable]
+  public class DeviceHeightPair {
+    public string deviceName;
+    public float heightOffset;
+
```
jo3w4rd: Since these are public, shouldn't they be capitalized: DeviceName and HeightOffset?

PR #246, opened 2016-05-24
Fix scene height

jo3w4rd:
Added the VRHeightOffset script to the DetectionExample scenes.
Fixed a null hand exception while drawing gizmos.
Changed the physics hand organization in the scenes.
And a few scene tweaks


jselstad:
Nice, this fixes the PinchDetector errors too

PR #248, opened 2016-05-24
UI Input Module Restoration

jselstad:
Adding back the UI Input Module Example Scene and Prefabs

Lots of refinements; commit log gives a pretty good summary.


Amarcolina:
Looks good to me!

PR #258, opened 2016-05-26
Adding framerate graph module 

Amarcolina:
People keep wanting to use this graph so I figured I would add it as an actual module so at the very least we have an easy and official way to maintain it.  Renamed from RuntimeProfiler to FramerateGraph because it is not really a profiler at all, just a graph of framerate.


Assets/LeapMotionModules/FramerateGraph/README.txt
```diff
@@ -0,0 +1,18 @@
...
+
+Graph:
+The graph contains an upper and a lower graph, which have different meanings.  The upper graph shows the
+time spent in script execution such as FixedUpdate / Update callbacks.  The lower graph shows the time 
+spent doing rendering for all cameras.  
+
+Right Panel:
+The right panel has three numbers, a top, middle, and bottom.
+The top and bottom numbers correspond with the top and bottom graphs, and represent the average time in
+miliseconds spent on the update cycle or the render cycle respectively.
```
jo3w4rd: typo: milliseconds

Assets/LeapMotionModules/FramerateGraph/README.txt
```diff
@@ -0,0 +1,18 @@
...
+The middle number shows the average time in miliseconds spent on the entire frame, rendering and update
```
jo3w4rd: Also here

Assets/LeapMotionModules/FramerateGraph/Scripts/FramerateGraph.cs
```diff
@@ -0,0 +1,227 @@
...
+
+    void FixedUpdate() {
+      if (_updateStartTick == -1) {
+        _updateStartTick = _stopwatch.ElapsedTicks;
+      }
+    }
+
+    void Update() {
+      if (_hasCompletedFrame) {
+        float totalRenderMilis = _totalRenderTick / (float)System.Diagnostics.Stopwatch.Frequency * 1000.0f;
```
jo3w4rd: Technically these should all be "Millis", too, but I don't know if we care about the spelling of local variables that much.

Amarcolina:
Will make name fixes

Amarcolina:
Ok names are fixed

PR #271, opened 2016-06-01
Restructure Gear VR related stuff to android module

veronicaz41:
This is supposed to work the same as branch orion-gearvr. After this is merged, we can get rid of that branch.

The Android Signature files, PinchDrawDemo, PinchMoveDemo are not included here. They will be kept in a separate branch (https://github.com/leapmotion/UnityModules/tree/orion/gearvr) for now. Because PinchDrawDemo and PinchMoveDemo depends on DetectionExamples module, and we want to avoid inter module dependencies. We need to figure out what to do with them in the future.
 
* When build for Gear VR, need to go to player settings -> other settings, and enable "Multithreaded Rendering"


testmasterbright:
@veronica41 I am getting merge conflicts locally. I'm curious why there are no conflicts on the PR.

PR #272, opened 2016-06-01
Allow user to modify physics scale

ajohnston33:


Amarcolina:
ideally the manager uses the scale of the provider to determine the scale instead of the user needing to enter it manually.

PR #285, opened 2016-06-06
Android Cardboard support

veronicaz41:
This has
1. The a subset of the latest Google VR SDK for Unity (https://developers.google.com/vr/unity/download)
    Only added a subset which is necessary for our demo scene to work. Since the whole thing is too large.
2. Cardboard HMD scene 'Leap_Hands_Demo_HMD.unity'


protodeep:
Does the GoogleVR folder need to go be at the top level of the project?  Could it be placed under LeapMotionModules/Android/?

veronicaz41:
Moved it into LeapMotionModules/Android/Cardboard.

PR #291, opened 2016-06-09
Turning offscreen updating back on for skinned meshes

protodeep:
- [ ] Check three scenes in /LeapMotionModules/Hands/Examples/Scenes to verify UpdateWhenOffscreen is set to True for skinned meshes


Amarcolina:
This changes the settings in the scene but it does not change the prefab.  The change should probably be in the prefab and not in the scene.

protodeep:
OK, offscreen updating has been set in prefabs as well

PR #303, opened 2016-06-16
Upgraded framerate graph to runtime graph

Amarcolina:
- Framerate graph is now called Runtime graph
- The runtime graph supports 6 different types of information.
   - Framerate
   - Frame Delta
   - Tracking Framerate
   - Tracking Latency
   - Update Delta
   - Render Delta
- The graph is now supported by the unity gui system to allow it to be navigated in mobile if needed
- The graph now has options to allow variable sample rate
- The graph is now dynamically resizing, allowing a wider range of values

To Test:
 - [x] Load up a VR scene and drop the runtime graph prefab in.  Make sure you add a Unity EventSystem with an input module of your choice!  Jons works well.

@protodeep 
@jselstad 
@testmasterbright


Assets/LeapMotionModules/RealtimeGraph/Scripts/RealtimeGraph.cs
```diff
@@ -0,0 +1,297 @@
...
+      _slidingMax = new SlidingMax(_historyLength);
+
+      _smoothedMax = new SmoothedFloat();
+      _smoothedMax.delay = _maxSmoothingDelay;
+
+      _smoothedValue = new SmoothedFloat();
+      _smoothedValue.delay = _valueSmoothingDelay;
+    }
+
+    protected virtual void Start() {
```
jo3w4rd: Maybe try to find a LeapProvider if one hasn't been set manually.

Amarcolina: Good idea, added

PR #305, opened 2016-06-21
Hierarchy Traversal UI Module Fix

jselstad:
This should unbreak compatibility with default Scroll Views and allow users to press infinitely tall compressible buttons with ease (where even the text and image on the button have "RaycastTarget" enabled)...

This is still largely untested...


jselstad:
stupid master branch why do we have it

PR #318, opened 2016-06-28
Added LoPolyHands to Android Module

jselstad:
Swapped Capsule Hands out for LoPolyHands (also copied them from the Hands module into the Android Module)

NOTE: LoPolyHand Scripts are duplicated (with different names)!
This is not a permanent solution!  At some point we're going to have _the talk_ (about intermodule dependencies).


protodeep:
iHandModels need to be dragged from hierarchy into the HandPool slots.

PR #332, opened 2016-07-05
Useability changes and IE toggling (settable ContactEnabled, Brushing…

nickjbenson:
…Enabled, GraspingEnabled)

Setting Contact, Brush, or GraspingEnabled in Interaction Manager will appropriately call UpdateSceneInfo(), along with some brush logic changes to decouple it from Soft Contact, allows all three features to be toggled independently and dynamically at runtime.

The changes related to InteractionBehaviour allow smart defaults to happen if you do an AddComponent<InteractionBehaviour>() at runtime, relying on a public defaultInteractionMaterial field in InteractionManager to set a default InteractionMaterial and searching the scene for an appropriate InteractionManager for the behaviour as well. Otherwise you get an error, have to set the material and manager fields, call Reset(), and then re-enable the object (four/five lines per instantiation goes down to one with this change) in order to add an InteractionBehaviour component at runtime (ew!)

### DIFF
--- a/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
+++ b/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
@@ -6,6 +6,7 @@ using UnityEngine;
 using UnityEngine.Assertions;
 using Leap.Unity.Interaction.CApi;
 using LeapInternal;
+using UnityEditor;
 
 namespace Leap.Unity.Interaction {
 
@@ -121,6 +122,9 @@ namespace Leap.Unity.Interaction {
     public InteractionMaterial material {
       get {
         return _material;
+      }
+      set {
+        _material = value;
       }
     }
 
@@ -437,6 +441,23 @@ namespace Leap.Unity.Interaction {
     public override void NotifyBrushDislocated() {
       _dislocatedBrushCounter = 0;
       updateContactMode();
+    }
+
+    #endregion
+
+    #region UNITY EVENTS
+
+    protected override void Reset() {
+      base.Reset();
+      if (_material == null) {
+        if (_manager == null) {
+          return;
+        }
+        else {
+          Debug.LogWarning("No InteractionMaterial specified; will use the default InteractionMaterial as specified by the InteractionManager.");
+          _material = _manager.defaultInteractionMaterial;
+        }
+      }
     }
 
     #endregion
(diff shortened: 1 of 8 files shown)


ProjectSettings/TagManager.asset
```diff
@@ -13,8 +13,8 @@ TagManager:
   - UI
   - 
   - 
-  - 
-  - 
+  - ButtonStopper
+  - Button
   -
```
ajohnston33: As a Unity module I don't think we are allowed to use tags.

ProjectSettings/DynamicsManager.asset
```diff
@@ -4,12 +4,12 @@
 PhysicsManager:
   m_ObjectHideFlags: 0
   serializedVersion: 2
-  m_Gravity: {x: 0, y: -9.81, z: 0}
```
ajohnston33: We have a script to change the settings in our demos.  I'll get you sorted offline.

ProjectSettings/TagManager.asset
nickjbenson: Yeah this is no good, I shouldn't have added this to this commit, this was a part of the example project rather than these module modifications

nickjbenson:
Hold this thought, I've got some fixes to make first. I got too excited and submitted this before checking some of the newer commits I pulled against my changes. (3:32 PM)

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -6,6 +6,7 @@
 using UnityEngine.Assertions;
 using Leap.Unity.Interaction.CApi;
 using LeapInternal;
+using UnityEditor;
```
Amarcolina: Make sure to wrap this in an ifdef or else unity will yell at you about missing namespaces when you try to build

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionBehaviourBase.cs
```diff
@@ -450,6 +450,11 @@ public abstract class InteractionBehaviourBase : IInteractionBehaviour {
     #endregion
 
     #region UNITY MESSAGES
+
+    protected virtual void Awake() {
+      this.Reset();
```
Amarcolina: This will become problematic if the user extends the Reset functionality of their own implementation to do things that don't make sense to do in Awake.  There probably should be separate logic (maybe a different protected method even) that takes care of ensuring that the properties are valid.

PR #333, opened 2016-07-05
Removed lazy transform copy logic

Amarcolina:
Frames are no longer transformed lazily and secretly the moment you ask for a frame.  They are only transformed under two conditions.  They are transformed right after the frame has been obtained from the service.  They are transformed if the user specifically asks that a transform take place.  This should ensure that maximum performance is possible with explicit control over when a transform takes place.

Frame events have been added to the provider to allow scripts to listen for a frame that they know is correctly transformed (since it just happened).  HandController has been updated to use this method.  If a user feels like moving the camera, they will need to make sure to either move it intelligently, or re-transform the frame before accessing positional information.

I also went through the scenes and made sure to de-parent the hands from the head.  All models are now a child of a 'HandModels' empty that is in camera-rig space.  I also made sure that all service providers were linked with the temporal warping script, and that settings were correct.

To test:
- [x] Make sure that all affected scenes work properly and that there are no errors
- [x] Create a script that moves the camera rig really fast and verify the hands don't get jacked up
- [x] Create a script that also places a sphere at the position of the index finger, in addition to moving the script.  Make sure to call ReTransformFrames() before accessing the data.  Verify that the sphere does indeed follow the finger correctly even when the camera rig is moving at high speeds.

@jo3w4rd 
@protodeep


jselstad:
Issues:
There was a Missing Script on the CameraRig in Leap_Hand_Demo_VR.unity (this must have been your testing scene); I fixed this one since it was easy.


Testing Notes (not issues):
Made a script to move the camera around in Update; 
1) hands look fine when HandModels are parented to the CameraRig (which is the default)
2) hands look bad when HandModels are deparented from the Rig (where the rig is moved in Update)
3) hands look fine when HandModels are deparented from the Rig and it is moved in FixedUpdate (before the Frame gets transformed)

Calling ReTransformFrames in Update does nothing for the hand because the hands positions are set independently of this call, so that's expected.

Using LeapServiceProvider.ReTransformFrames(); works as expected; finger sphere is off when camera is moved in Update, but it sticks reliably to the hand when ReTransformFrames is called before setting its position in Update.

Overall, it looks good.

Overall, it looks good; merging.

PR #338, opened 2016-07-08
UI Input Module V1.1

jselstad:
Compressible UI:
-Renaming fields for clarity
-Changed how compression speed is taken into account (doesn't apply to hovering motions anymore, only compress and retract events)
-Using Re-orderable List
-Layers are named after their gameobject
-Hide Event Trigger

Prefabs:
-Fully formed canvas prefab with buttons
--Removed Audiosource per button in canvas prefab

Bugs that got fixed:
-Deep copies of frames
-Dragging doesn't feel right compared to normal input modules
-Per frame memory allocation from Linq

Some other misc. changes too


protodeep:
There are still audiosources for each element in the example scene.  Can we consolidate those at least for the mobile scene?

PR #350, opened 2016-07-20
Useability and toggling for IE contact/grasping and dynamically spawn…

nickjbenson:
…ing InteractionBehaviours.

Toggling: Just added appropriate setters for ContactEnabled and GraspingEnabled.

Spawning InteractionBehaviours: Spawning InteractionBehaviours onto objects from a script is not fun because you get lots of errors about missing an attached InteractionManager and InteractionMaterial even if you set them right after creating the behaviour. This solves that problem by adding a default InteractionMaterial setting to the InteractionManager and also has InteractionBehaviours try to find a manager on-load if it doesn't have one.

### DIFF
--- a/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
+++ b/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
@@ -42,6 +42,10 @@ namespace Leap.Unity.Interaction {
     protected string _ldatPath = "InteractionEngine/IE.ldat";
 
     [Header("Interaction Settings")]
+    [Tooltip("The default Interaction Material to use for Interaction Behaviours if none is specified, or for Interaction Behaviours created via scripting.")]
+    public InteractionMaterial defaultInteractionMaterial;
+
+    [Header("Interaction Settings")]
     [Tooltip("Allow the Interaction Engine to modify object velocities when pushing.")]
     [SerializeField]
     protected bool _contactEnabled = true;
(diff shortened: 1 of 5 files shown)


Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -42,6 +42,10 @@ public class InteractionManager : MonoBehaviour {
     protected string _ldatPath = "InteractionEngine/IE.ldat";
 
     [Header("Interaction Settings")]
+    [Tooltip("The default Interaction Material to use for Interaction Behaviours if none is specified, or for Interaction Behaviours created via scripting.")]
+    public InteractionMaterial defaultInteractionMaterial;
+
+    [Header("Interaction Settings")]
```
Amarcolina: We now have two 'Interaction Settings' headers.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -42,6 +42,10 @@ public class InteractionManager : MonoBehaviour {
     protected string _ldatPath = "InteractionEngine/IE.ldat";
 
     [Header("Interaction Settings")]
+    [Tooltip("The default Interaction Material to use for Interaction Behaviours if none is specified, or for Interaction Behaviours created via scripting.")]
+    public InteractionMaterial defaultInteractionMaterial;
```
Amarcolina: Variable should be protected, and only have public getters/setters if needed, to follow along with how other variables are treated in the manager.

Amarcolina:
alright everything looks great!

PR #353, opened 2016-07-21
Feature auto rigging

protodeep:


.gitignore
```diff
@@ -18,6 +18,7 @@
 /Temp
 ExportedObj/
 Temp/
+/Assets/LeapMotionModules/Hands/Models/TestModels/
```
Amarcolina: You can put a gitignore inside the folder that contains the file you want to ignore, that way you don't need to hard-code paths.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -223,6 +223,21 @@ public class ModelGroup {
...
+    public void AddNewGroup(string groupName, IHandModel leftModel, IHandModel rightModel) {
+      ModelGroup newGroup = new ModelGroup();
+      newGroup.LeftModel = leftModel;
+      newGroup.RightModel = rightModel;
+      newGroup.GroupName = groupName;
+      newGroup.CanDuplicate = false;
+      newGroup.IsEnabled = true;
+      ModelPool.Add(newGroup);
+    }
+    public void RemoveGroup(string groupName) {
```
Amarcolina: This method will only remove one group if multiple groups exist with the same name.

Assets/LeapMotion/Scripts/Hands/IHandModel.cs
```diff
@@ -18,7 +18,7 @@ public abstract class IHandModel : MonoBehaviour {
       get { return isTracked; }
     }
 
-    public abstract Chirality Handedness { get; }
+    public abstract Chirality Handedness { get; set; }
```
Amarcolina: It is not necessary to add the ability to set handedness to the base interface if all you need is the ability to set the handedness of your rigged hands.

Assets/LeapMotion/Scripts/Hands/IHandModel.cs
```diff
@@ -49,8 +49,14 @@ public abstract class IHandModel : MonoBehaviour {
 
 #if UNITY_EDITOR
     void Update() {
+      Transform editorPoseSpace;
+      LeapServiceProvider leapServiceProvider = (LeapServiceProvider)FindObjectOfType(typeof(LeapServiceProvider));
```
Amarcolina: You can use the generic version of this method for less verbosity.

Assets/LeapMotionModules/Hands/Scripts/Editor/AutoRigEditor.cs
```diff
@@ -0,0 +1,23 @@
+﻿using UnityEngine;
+using System.Collections;
+using UnityEditor;
+
+namespace Leap.Unity {
+  [CustomEditor(typeof(LeapHandsAutoRig))]
+  public class ObjectBuilderEditor : Editor {
```
Amarcolina: The name of an editor class should match the name of the class that it edits, and should also match the name of the file it exists in.  For this case the name would be LeapHandsAutoRigEditor.

Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
```diff
@@ -0,0 +1,302 @@
+﻿using UnityEngine;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Leap.Unity {
+  /**LeapHandAutoRig automates setting up the scripts that drive 3D skinned mesh hands. */
+  [ExecuteInEditMode]
```
Amarcolina: This attribute does nothing because the behaviour doesn't have an Awake or Update method.

Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
```diff
@@ -0,0 +1,302 @@
...
+  [AddComponentMenu("Leap/Auto Rig Hands")]
+  public class LeapHandsAutoRig : MonoBehaviour {
+    public HandPool HandPoolToPopulate;
+    public Animator AnimatorForMapping;
+
+    public string ModelGroupName = null;
+    [Tooltip("Set to True if each finger has an extra trasform between palm and base of the finger.")] 
+    public bool UseMetaCarpals;
```
Amarcolina: You may want to consider making many/all of these fields non-public.  Since you rely on validation for a lot of the work, and validation is not triggered if the values are changed via script.  It might be a good idea to prevent modification from script altogether to prevent entering an invalid state.

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -6,25 +6,55 @@
...
+    [Header("Values for Stored Start Pose")]
+    [SerializeField]
+    private List<Transform> jointList = new List<Transform>();
+    [SerializeField]
+    private List<Quaternion> localRotations = new List<Quaternion>();
+    [SerializeField]
+    private List<Vector3> localPositions = new List<Vector3>();
 
+    void OnValidate() {
+      if (SetEditorLeapPose == setEditorLeapPose) {
```
Amarcolina: Not sure what is going on here?  SetEditorLeapPose is just a public getter for the private setEditorLeapPose

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -76,25 +106,101 @@ public class RiggedHand : HandModel {
...
-      modelPalmFacing = calculateModelPalmFacing();
+      assignRiggedFingersByName();
+      SetupRiggedFingers();
+      modelPalmFacing = calculateModelPalmFacing(palm, fingers[1].transform, fingers[2].transform);
       modelFingerPointing = calculateModelFingerPointing();
       setFingerPalmFacing();
     }
+    /**Sets up the rigged hand if RiggedFinger scripts have already been assigned using Mecanim values.*/
+    public void AutoRigRiggedHand(Transform palm, Transform finger1, Transform finger2) {
+      Debug.Log("AutoRigRiggedHand()");
```
Amarcolina: Log should either come out or turn into something more descriptive.

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -76,25 +106,101 @@ public class RiggedHand : HandModel {
...
+      modelFingerPointing = new Vector3(0, 0, 0);
+      modelPalmFacing = new Vector3(0, 0, 0);
+      SetupRiggedFingers();
+      modelPalmFacing = calculateModelPalmFacing(palm, finger1, finger2);
+      modelFingerPointing = calculateModelFingerPointing();
+      setFingerPalmFacing();
+    }
+    /**Finds palm and finds root of each finger by name and assigns RiggedFinger scripts */
+    private void assignRiggedFingersByName(){
+      List<string> palmStrings = new List<string> { "palm", "Palm", "PALM" };
```
Amarcolina: One tricks you can use if you want.  I like to convert everything to lowercase before doing a comparison, that way you don't have to have a bunch of different versions of the same word.

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -76,25 +106,101 @@ public class RiggedHand : HandModel {
...
+      List<string> pinkyStrings = new List<string> { "pinky", "pin", "Pinky", "PINKY" };
+      //find palm by name
+      //Transform palm = null;
+      Transform thumb = null;
+      Transform index = null;
+      Transform middle = null;
+      Transform ring = null;
+      Transform pinky = null;
+      Transform[] children = transform.GetComponentsInChildren<Transform>();
+      if (palmStrings.Any(w => transform.name.Contains(w)) ==true ){
```
Amarcolina: ==true is not needed

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -76,25 +106,101 @@ public class RiggedHand : HandModel {
...
+        base.palm = transform;
+      }
+      else{
+        foreach (Transform t in children) {
+          if (palmStrings.Any(w => t.name.Contains(w)) == true) {
```
Amarcolina: ==true is not needed

Assets/LeapMotionModules/Hands/Scripts/RiggedHand.cs
```diff
@@ -118,15 +224,15 @@ public class RiggedHand : HandModel {
       }
       else perpendicular = Vector3.Cross(side1, side2);
       Vector3 calculatedPalmFacing = CalculateZeroedVector(perpendicular);
-      return calculatedPalmFacing;
+      return calculatedPalmFacing * 1; //+works for Mixamo, -reversed LoPoly_Hands_Skeleton and Winston
```
Amarcolina: *1 can be removed.

Amarcolina:
What version of Unity was used to dev this?  I'm getting a bunch of prefab errors upon initial import, and a bunch of the prefabs are broken and won't load.

Assets/LeapMotionModules/Hands/Shaders/BlendTesting.shader
```diff
@@ -0,0 +1,121 @@
+﻿Shader "Character Indicator"
```
Amarcolina: What is this shader for?  It seems to be just a regular alpha-cutoff shader right now, but looks like it might have been used for something else at some point?

Assets/LeapMotionModules/Hands/Shaders/UnLitAlpha.shader
```diff
@@ -0,0 +1,66 @@
+﻿// Unlit alpha-blended shader.
```
Amarcolina: What is this shader used for?  It seems to be a slightly different shader than Unlit/Transparent.

Amarcolina:
In Rigged_Hands_AutoRig_Example, the physics and graphics models are children of the center eye instead of the rig transform.

Amarcolina:
In general this PR also seems to include a bunch of work done on hand rendering and shaders, is this something we are ready to include in our release?

Amarcolina:
Right now there are two material folders, one in Hands/Materials, and the other in Hands/Models/Materials.  The former is materials you created yourself and the other is materials that are automatically imported from the models.  It might be useful to consolidate these folders somehow, but I'm not sure, what are your thoughts?

Amarcolina:
There seems to be a meta file for a non-existant folder being commited.
Assets/LeapMotionModules/Hands/Models/TestModels.meta

Amarcolina:
Alright, all changes look good to me!  Awesome job!

PR #358, opened 2016-07-23
Adding combinable attributes

Amarcolina:
Adding more advanced attribute functionality.  By default property attributes cannot be combined, and Unity just decides to pick one of them.  Using the CombinableAttribute allows attributes to be combined in an extensible way.  

The following features exist for combinable attributes:
 - The ability to add arbitrary constraints to the property value
 - The ability to conditionally disable a property
 - The ability to add in layout elements within the property
 - The ability to completely override the way a property is drawn

The following attributes have been added:
 - MinValue: Restrict a float/int to have a minimum value
 - MaxValue:  Restrict a float/int to have a maximum value
 - Units: Add a label after a field to show what units the field is measured in.
 - MinMax:  Operates on a Vector2 to provide a double slider
 - EnumFlags:  Surfaces the ability to combine enums that represent flag values into the inspector.
 - DisableIf:  Disables a property depending on the value of another property
 - Incrementable:  Adds plus and minus buttons to a value field to increment or decrement it 

To test:
 - [x] Ensure no compile errors or anything bad
 - [x] Ensure that unity can still build without namespace errors
 - [ ] Create a test MonoBehaviour and test that each attribute works properly
 - [x] Make sure to test some of them in combination!  For example Incrementable plus MinValue.  

@jselstad 
@protodeep 
@nickjbenson


jselstad:
The only issue that I've found is that, if the Units Attribute is given an extremely long string, it will clobber the input field (to have a negative x-scale) rather than go off the right side of the inspector like a good value....

PR #361, opened 2016-07-25
Moved data structures and algorithms into core

Amarcolina:
- Moved the RingBuffer data structure to core 
 - Moved the SlidingMax algorithm to core 
 - Moved the smoothed value classes to the new Algorithms folder

This is a non-critical PR that moves files from realtime graph to core.

@jselstad 
@protodeep


jselstad:
I don't suppose there's anything we can do about
>        if (average_dt * 10 < newValue - value)
>          Debug.Log("Average dt was excessive!");

in SmoothedFloat.cs while we're here, is there?   Normal devs probably won't appreciate the stunning insights that this log provides...

ajohnston33:
Please delete.  We should be robust against long pauses anyhow.  It is not clear this indicates a problem.

Amarcolina:
I'll change to

```
if (average_dt != newValue - value)
  Debug.Log("Imperfect dt was imperfect!");
```

instead then.

PR #372, opened 2016-07-28
Fix interaction tests

Amarcolina:
- ThrowingControllerSlidingWindow now handles cases where a throw takes less than 2 frames
 - Interaction Behaviours now return themselves to the default layer upon unregister
 - Unregistering an object that is suspended now dispatches the proper callbacks 
 - Forcing a grasp of an object that is registered but not active now activates the object
 - Added some null checks to validation
 - Added some null checks to the manager to catch cases where objects suddenly decide to kill themselves

@ajohnston33 

To test:
 - [ ] Make sure all tests pass :D


ajohnston33:
N.B. observers,  There are tests that are failing randomly due to apparent timing issues.  These will be resolved later.

PR #374, opened 2016-07-29
Feature interaction tests improve determenism

Amarcolina:
- Made playback based on timestamps instead of frames
 - Made playback framerate independent 
 - Remove frame-limiter since it was not needed anymore
 - Re-recorded recordings at higher framerate (100fps instead of 75fps)
 - Added ability to run tests at a higher timescale (currently 10x) for great justice 

Hopefully this improves the determinism of the playback system.  It should also allow us to run tests more quickly (within reason).


ajohnston33:
Looks good.

PR #379, opened 2016-08-04
Interaction Engine Fixing Sticky Release, C# side.

ajohnston33:
Transitions into soft contact when releasing an object, avoids unruly brush behavior.
Leaves objects in the layer they were using for soft contact when un-registering as that is safest.
Adds visualization the the coverage suite allowing it to be a debug tool.
There is a critical corresponding fix on the C++ this change pairs up with.

### DIFF
--- a/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
+++ b/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
@@ -314,7 +314,7 @@ namespace Leap.Unity.Interaction {
     }
 
     /// <summary>
-    /// Gets or sets the max activation depth.
+    /// Enables the display of proximity information from the library.
     /// </summary>
     public bool ShowDebugLines {
       get {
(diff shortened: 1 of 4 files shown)


Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -313,6 +313,32 @@ public partial class InteractionManager : MonoBehaviour {
       }
     }
 
+    /// <summary>
+    /// Gets or sets the max activation depth.
```
Amarcolina: Comment is wrong

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -313,6 +313,32 @@ public partial class InteractionManager : MonoBehaviour {
...
+        return _showDebugLines;
+      }
+      set {
+        _showDebugLines = value;
+        applyDebugSettings();
+      }
+    }
+
+    /// <summary>
+    /// Gets or sets the max activation depth.
```
Amarcolina: Comment is wrong

PR #383, opened 2016-08-04
When a model is returned to pool, check for another active representa…

protodeep:
…tion of that chirality

- [ ] With CanDuplicate off, Hold up zombie left hand and bring in another left hand.  Model's will jump to latest hand.  Removing hand, will cause model's to jump to remaining hand.

- Also removing the "Either" chirality setting


Amarcolina:
Looks great!

PR #384, opened 2016-08-04
Runtime Gizmos Utility

Amarcolina:
This adds a new utility that replaces Unity's built-in gizmos.  The runtime gizmos offer the exact same api that unity provides, with a number of improvements:
 - Greater control over matrices (push / pop matrices) (draw relative to a transform)
 - Ability to change the shaders used to render the gizmos
 - Ability to change the meshes used to render spheres and cubes
 - VR compatible (Unity's implementation is currently broke both in the game view and in the headset)
 - Runtime compatible!  
 - Zero allocation when enabled
 - Zero overhead when disabled
 - Can selectively disable the gizmos for an entire transform hierarchy 

To test:
 - [x] Create a MonoBehaviour and have it inherit from IRuntimeGizmoDrawer.  Use the RGizmos class to draw your gizmos like you would normally.
 - [x] Make sure that each draw method functions the same was as normal unity gizmos would
 - [x] Make sure that gizmos work in scene view + game view + headset view
 - [x] Make sure that gizmos still work in a build
 - [x] Make sure that the RuntimeGizmoToggle correctly disables the gizmos of all objects that are children of it.

@protodeep 
@nickjbenson 
@jselstad


jselstad:
Gizmos depth test with the scene, but not with UI 乁( ⁰͡ Ĺ̯ ⁰͡ ) ㄏ

Otherwise it looks good to me.

Amarcolina:
Makes sense since the UI doesn't draw to the depth buffer.  There might be a way around it with command buffers, but I'll save that for a later PR

jselstad:
"RuntimeColliderGizmos" isn't in proper namespace

Actually, now that I look at it... "VRHeightControl", "CycleHandPairs", "FlameControl", "Cursor3D", "CenterTransformOnCam", and "SpawnBalls" aren't in a Leap.Unity namespace either

I made the mistake with "Cursor3D", but is it good practice to put examples in a Leap.Unity namespace?  Probably...

Amarcolina:
Good catch, I'll change that

PR #389, opened 2016-08-11
Remove Daydream-specific Android x86 binaries

jdonald:
We do not support Android x86 for Leap + Daydream, and these binaries cause conflicts when exporting for OS X on Unity 5.4.0.


hham:
Looks good.

PR #390, opened 2016-08-11
Gizmo front buffer protection

Amarcolina:
Protect the front buffer with a guard while it is being used for rendering.  This allows us to catch issues where people are holding onto references to drawers.

To test:
 - [x] Everything should still work without errors normally
 - [x] When holding onto a reference to a drawer, you should get an error

@jselstad


jselstad:
Looks good, merging.

PR #393, opened 2016-08-12
Added event behaviours 

Amarcolina:
Added two new behaviours, InteractionGraspEvents and InteractionHandEvents.  They surface some interaction callbacks as Unity events so that users can easily control things without needing to manually code the wiring.


ajohnston33:
This looks a bit frivolous, however I am going to assume there is a Unity use case that makes this super handy.

PR #400, opened 2016-08-16
Fail Quietly if Model Slot is Empty

protodeep:
… empty

- [ ] Remove a hand model from a HandPool slot and play any scene.  Verify that no errors are thrown.


Amarcolina:
Looks great to me!

PR #401, opened 2016-08-16
Parent Spawned Hand Models to Correct Transform

protodeep:
Spawned models should be parented to the same transform as all other hand models.  A ModelsParent transform field is added to HandPool.  This needs to be referenced in all scenes.  Otherwise, a warning is logged.

- [ ] Test one or more scenes to verify that prefabs directly referenced into HandPool slots or extra clones of models allowed by CanDuplicate are all parented to the referenced ModelsParent transform.


Amarcolina:
Looks good to me!

PR #403, opened 2016-08-17
Small ie usability changes

Amarcolina:
- Made sure monitor gizmos are on Interaction Status by default, instead of the more unintuitive activity depth.
 - Removed the 'pause simulation' toggle because it was not useful in any way.

@ajohnston33


ajohnston33:
These are superficial changes that will be safe to release.

PR #413, opened 2016-08-25
Use RuntimeGizmo drawing for attachment hand lines.

jo3w4rd:
Also updates scene file.

### DIFF
--- a/Assets/LeapMotionModules/Attachments/Scripts/HandAttachments.cs
+++ b/Assets/LeapMotionModules/Attachments/Scripts/HandAttachments.cs
@@ -157,7 +157,7 @@ namespace Leap.Unity.Attachments {
     * Also draws the orthogonal basis vectors for the pinch and grab points.
     */
     public void OnDrawRuntimeGizmos(RuntimeGizmoDrawer gizmoDrawer) {
-      if (DrawHand && RuntimeGizmoManager.TryGetGizmoDrawer(gameObject, out gizmoDrawer)) {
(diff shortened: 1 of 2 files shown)


Amarcolina:
To fully support runtime gizmos, you are going to need to take the gizmo drawing code outside of the default OnDrawGizmos method that unity provides.  You should implement the IRuntimeGizmoComponent interface and use it's callback for drawing gizmos instead.

Assets/LeapMotionModules/Attachments/Scripts/HandAttachments.cs
```diff
@@ -145,20 +149,15 @@ public class HandAttachments : IHandModel {
...
 
     /**
     * Draws lines from elbow to wrist, wrist to palm, and normal to the palm.
     * Also draws the orthogonal basis vectors for the pinch and grab points.
     */
-    protected void DrawDebugLines() {
-      RuntimeGizmoDrawer gizmoDrawer;
-      if (RuntimeGizmoManager.TryGetGizmoDrawer(gameObject, out gizmoDrawer)) {
+    public void OnDrawRuntimeGizmos(RuntimeGizmoDrawer gizmoDrawer) {
+      if (DrawHand && RuntimeGizmoManager.TryGetGizmoDrawer(gameObject, out gizmoDrawer)) {
```
Amarcolina: You don't need to try getting the gizmo drawer from inside the callback.  The callback is only called if a drawer is available.

PR #418, opened 2016-08-26
Add InteractionGraspDetector

jo3w4rd:
A Detector that activates when an IInteractionBehaviour instance is grasped under the auspices of the Interaction Engine.

You can specify that any intractable object can activate the detector; you can supply a tag name, and you can supply a specific list of target objects. (Tags and the list are additive.)

### DIFF
--- a/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionGraspDetector.cs
+++ b/Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionGraspDetector.cs
@@ -1,19 +1,33 @@
 ﻿using UnityEngine;
 using System.Collections;
-using System.Collections.Generic;
 using Leap.Unity.Attributes;
 using UnityEngine.Events;
 
 namespace Leap.Unity.Interaction {
+  /**
+  * Detects when an interactable object is grasped.
+  * You can specify whether any interactable object or
+  * any object with a specified tag will a activate the detector. You can also
+  * assign a specific list of activating objects.
+  *
+  * All objects must contain an IInteractionBehaviour component and be managed by the
+  * InteractionManager assigned to this detector. 
+  * @since 4.1.5
+  */
   public class InteractionGraspDetector : Detector {
 
+    /**
+    * The InteractionManager managing the target objects. All
+    * targets must be managed by this manager.
+    * @since 4.1.5
+    */
     [AutoFind(AutoFindLocations.Scene)]
     [Tooltip("The Interaction Manager.")]
     public InteractionManager ieManager = null;
     /**
      * The IHandModel instance to observe. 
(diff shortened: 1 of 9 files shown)


Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionGraspDetector.cs
```diff
@@ -0,0 +1,143 @@
...
+     */
+    [AutoFind(AutoFindLocations.Parents)]
+    [Tooltip("The hand model to watch. Set automatically if detector is on a hand.")]
+    public IHandModel HandModel = null;
+
+    /**
+ * The interval at which to check palm direction.
+ * @since 4.1.5
+ */
+    [Tooltip("The interval in seconds at which to check this detector's conditions.")]
```
Amarcolina: Would be great to use a MinValue attribute here to prevent this value from being negative :D

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionGraspDetector.cs
```diff
@@ -0,0 +1,143 @@
...
+
+    [Tooltip("Activate when any interactable object is grasped.")]
+    public bool AnyInteractionObject = false;
+    /**
+     * The list of IInteractionBehaviour objects which can activate the detector when held.
+     * Grasping objects in this list will activate the detector even if they do not match the tag
+     * defined in TagName.
+     * @since 4.1.5
+     */
+    [Tooltip("The list of target objects.")]
```
Amarcolina: You could use the DisableIf attribute to disable this field (and the TagName) field if 'AnyInteractionObject' is true.  That way it is more clear that these fields do nothing when that box is checked.

jo3w4rd: DisableIf only half works on array variables -- the inspector for the array is still enabled, but the elements themselves are disabled. So you can add elements, but can't edit them. I'm not sure it is an improvement in this case.

Amarcolina: Oh weird, that's good to know.  Yeah sounds like it would not be an improvement in this case.

jo3w4rd: I went ahead and added the DisableIf attribute. It doesn't help so much when first setting up the detector, but could help when debugging since you do get an indication that the objects are disabled in the list, if you happen to look.

Also added a slew of attributes to the other Detectors to improve the Inspectors.

Amarcolina:
Awesome!  Looks great!

PR #425, opened 2016-08-31
Upgrade leapc

Amarcolina:
- Change api usage to match new api, only small changes like removing IsEmbedded
 - Added a better unit test to make sure that TransformCopy produces a binary equivalent frame when using the identity transform.  Should help copy errors in the future.

NOTE: this PR does not yet contain the updated dlls.  This PR is to test with 6791 on platform.  Once 6791 is validated, use that branch to build the dlls to insert into this PR before it is merged.

### DIFF
--- a/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
+++ b/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
@@ -142,7 +142,7 @@ namespace Leap.Unity {
 
         // TODO: Add baseline & offset when included in API
         // NOTE: Alternative is to use device type since all parameters are invariant
-        info.isEmbedded = devices[0].Type == Device.DeviceType.TYPE_KEYBOARD || devices[0].Type == Device.DeviceType.TYPE_LAPTOP;
+        info.isEmbedded = devices[0].Type != Device.DeviceType.TYPE_PERIPHERAL;
         info.horizontalViewAngle = devices[0].HorizontalViewAngle * Mathf.Rad2Deg;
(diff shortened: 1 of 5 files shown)


Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -142,7 +142,7 @@ public class LeapServiceProvider : LeapProvider {
 
         // TODO: Add baseline & offset when included in API
         // NOTE: Alternative is to use device type since all parameters are invariant
-        info.isEmbedded = devices[0].IsEmbedded;
+        info.isEmbedded = devices[0].Type == Device.DeviceType.TYPE_KEYBOARD || devices[0].Type == Device.DeviceType.TYPE_LAPTOP;
```
jo3w4rd: Note that device type isn't actually implemented. But either way, since the service no longer has a concept mapping to "isEmbedded" we might just want to remove it here.

jselstad:
On hold until Image Warping regression is figured out.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
Amarcolina: That sounds good to me.  I guess there should be a revamp for device info at some point, so we can stop hacking around, but I think that all of this should be in a separate PR, since it needs wider changes like fixing Disconnection Notice

Amarcolina:
This PR is blocked for the moment.  The layout of LEAP_HAND changed, breaking IE

Amarcolina:
Closing this until resolved

PR #435, opened 2016-09-12
Feature serializable dictionary

Amarcolina:
This PR adds the class SerializableDictionary, which is a subclass of the normal Dictionary class, but which supports unity serialization.  In order to use it, you will need to create a non-generic subclass for your specific use-case (just like using UnityEvent).  If you want the dictionary to show up nicely in the inspector, make sure to tag the dictionary with the [SDictionary] attribute.

To test:
 - [x] Make sure that when duplicate keys are present, that they are marked as duplicates (should appear yellow)
 - [x] Make sure that a serialized dictionary retains its values when Unity is closed
 - [x] Make sure that a serialized dictionary retains its values when the scene is played
 - [x] Make sure that a serialized dictionary retains its values in a build
 - [x] Make sure that a serialized dictionary retains its values when it is on an object that is duplicated.
 - [x] Make sure that you can use serializable structs as keys
 - [x] Make sure that you can use object references as keys (like gameObject)
 - [x] Make sure that null object-references do not get saved into the dictionary (they should appear red)
 - [x] Make sure that the button to remove duplicates works as expected.
 - [x] Make sure that you can remove pairs when the key is an object reference
 - [x] Make sure things don't crash when you try to multi-edit multiply dictionaries at once.

### DIFF
--- a/Assets/LeapMotion/Scripts/DataStructures/Editor/SerializableDictionaryEditor.cs
+++ b/Assets/LeapMotion/Scripts/DataStructures/Editor/SerializableDictionaryEditor.cs
@@ -196,7 +196,7 @@ public class SerializableDictionaryEditor : PropertyDrawer {
       copy.NextVisible(true);
       while (!SerializedProperty.EqualContents(copy, endProp)) {
         r.height = EditorGUI.GetPropertyHeight(copy);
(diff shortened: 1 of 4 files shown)


Amarcolina:
On hold, found some issues

PR #442, opened 2016-09-20
OnPreCull Latching for Hands

jselstad:
Add a special kind of standard shader that moves the model by a pair of worldspace transforms, which are set by the LeapServiceProvider in an OnPreCull callback.

When enabled (on the service provider and in the material), this will remove a full frame of apparent latency from the tracking for "free".

To validate that it is working, you can have both hands that don't have the material and hands that do have the material, and add "160" to the ManualUpdate call on line 389 in LeapServiceProvider to see them separate (the fingers should update in real time, but the rigidtransforms should be separated).

Or, you can uncomment the realtime graph lines in the Service Provider and verify the precull and  tracking latency are near 16ms.   

- [x] Can work on existing hands
- [x] Works in a build
- [x] Works on GearVR
- [x] Works on Cardboard
- [x] Works on Daydream


Assets/LeapMotion/Resources/VertexOffsetShader.shader
```diff
@@ -0,0 +1,49 @@
...
+#pragma surface surf Standard fullforwardshadows vertex:vert
+
+			// Use shader model 3.0 target, to get nicer looking lighting
+#pragma target 3.0
+
+			uniform float4x4 _handTransforms[2];
+			int _isLeftHand;
+			void vert(inout appdata_full v) {
+				int startIndex = _isLeftHand;
+				v.vertex = mul(unity_WorldToObject, mul(_handTransforms[startIndex], mul(unity_ObjectToWorld, v.vertex)));
```
Amarcolina: It would be great to put this shader logic into an include like how is done for LeapCG.cginc.  That way it makes it very easy for anybody (including us) to create their own late latched shaders.

Assets/LeapMotion/Resources/VertexOffsetShader.shader
```diff
@@ -0,0 +1,49 @@
...
+			LOD 200
+
+			CGPROGRAM
+			// Physically based Standard lighting model, and enable shadows on all light types
+#pragma surface surf Standard fullforwardshadows vertex:vert
+
+			// Use shader model 3.0 target, to get nicer looking lighting
+#pragma target 3.0
+
+			uniform float4x4 _handTransforms[2];
```
Amarcolina: Since this constant is global, we should make extra sure that there are no collisions.  I also think it would be good to follow the same naming convention used by LeapCG, by prefixing everything by _Leap and such.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -177,16 +203,23 @@ public class LeapServiceProvider : LeapProvider {
...
+        Shader.SetGlobalMatrixArray("_handTransforms", _transformArray);
+      }
+      _prevUpdateHandInPrecull = _updateHandInPrecull;
+
       _fixedOffset.Update(Time.time - Time.fixedTime, Time.deltaTime);
 
       if (_useInterpolation) {
 #if UNITY_ANDROID
-        Int64 time = leap_controller_.Now() - (_interpolationDelay+16) * 1000;
+        Int64 time = leap_controller_.Now() - ((_interpolationDelay+16) * 1000) + (_updateHandInPrecull ? (long)(Time.smoothDeltaTime * S_TO_NS / Time.timeScale) : 0);
```
Amarcolina: This will break if timescale is set to 0.  Might need to calculate your own smoothed delta based on real time.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -177,16 +203,23 @@ public class LeapServiceProvider : LeapProvider {
...
         leap_controller_.GetInterpolatedFrame(_untransformedUpdateFrame, time);
 #else
-        _smoothedTrackingLatency.value = Mathf.Min(_smoothedTrackingLatency.value, 25000f);
+        _smoothedTrackingLatency.value = Mathf.Min(_smoothedTrackingLatency.value, 30000f);
         _smoothedTrackingLatency.Update((float)(leap_controller_.Now() - leap_controller_.FrameTimestamp()), Time.deltaTime);
-        leap_controller_.GetInterpolatedFrame(_untransformedUpdateFrame, leap_controller_.Now() - (long)_smoothedTrackingLatency.value - (_interpolationDelay * 1000));
+        leap_controller_.GetInterpolatedFrame(_untransformedUpdateFrame, leap_controller_.Now() - (long)_smoothedTrackingLatency.value - (_interpolationDelay * 1000) + (_updateHandInPrecull ? (long)(Time.smoothDeltaTime * S_TO_NS / Time.timeScale) : 0));
```
Amarcolina: This offset calculation is done in many places, probably should be put in its own method to reduce duplication.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -223,6 +257,19 @@ public class LeapServiceProvider : LeapProvider {
...
     }
 
+    public void ManualUpdateFrame(long temporalOffset = 0) {
+      if (_useInterpolation) {
+        leap_controller_.GetInterpolatedFrame(_untransformedPreCullFrame, leap_controller_.Now() - (long)_smoothedTrackingLatency.value - ((_interpolationDelay + temporalOffset) * 1000));
+      } else {
+        leap_controller_.Frame(_untransformedPreCullFrame);
+      }
+
+      if (_untransformedPreCullFrame != null) {
```
Amarcolina: Is there any point to this null check?

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -288,14 +335,13 @@ public class LeapServiceProvider : LeapProvider {
...
       LeapTransform leapTransform;
       if (_temporalWarping != null) {
-        Vector3 warpedPosition;
-        Quaternion warpedRotation;
-        _temporalWarping.TryGetWarpedTransform(LeapVRTemporalWarping.WarpedAnchor.CENTER, out warpedPosition, out warpedRotation, source.Timestamp);
-
-        warpedRotation = warpedRotation * transform.localRotation;
+        if (resampleTemporalWarping) {
+          _temporalWarping.TryGetWarpedTransform(LeapVRTemporalWarping.WarpedAnchor.CENTER, out warpedPosition, out warpedRotation, source.Timestamp);
+          warpedRotation = warpedRotation * transform.localRotation;
```
Amarcolina: Should warpedPosition also be updated?

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -305,5 +351,62 @@ public class LeapServiceProvider : LeapProvider {
 
       dest.CopyFrom(source).Transform(leapTransform);
     }
+
+    //Precull-Latching Functions
+    protected virtual void OnEnable() {
```
Amarcolina: OnDisable and OnEnable should be placed up next to all the other Unity event callbacks.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -305,5 +351,62 @@ public class LeapServiceProvider : LeapProvider {
...
+      Camera.onPreCull -= LateUpdateHandTransforms;
+      Camera.onPreCull += LateUpdateHandTransforms;
+      _transformArray[0] = Matrix4x4.identity;
```
Amarcolina: This logic to reset the global shader constant can be placed in it's own method to reduce duplication.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -305,5 +351,62 @@ public class LeapServiceProvider : LeapProvider {
...
+      _transformArray[1] = Matrix4x4.identity;
+      Shader.SetGlobalMatrixArray("_handTransforms", _transformArray);
```
Amarcolina: The name of the matrix array should be a public constant defined at the top instead of hardcoded.  Especially since Unity does not throw any errors when you set properties that don't exist.

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -177,16 +203,23 @@ public class LeapServiceProvider : LeapProvider {
       }
 #endif
 
+      if (_updateHandInPrecull!=_prevUpdateHandInPrecull && !_updateHandInPrecull) {
```
Amarcolina: You could simplify this logic to !_updateHandInPrecull && _prevUpdateHandInPrecull

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
Amarcolina: Fair enough, no harm

Assets/LeapMotion/Scripts/LeapServiceProvider.cs
Amarcolina: Ah that's right, that makes sense.

PR #444, opened 2016-10-04
Documentation comments for UIInput scripts

jo3w4rd:
Added some comments for the docs.


jo3w4rd:
Mainly they are because the classes don't show up in the docs without correctly formatted comments.

PR #449, opened 2016-10-11
Feature expose position joints

protodeep:
- [x] In the Rigged_Hands_AutoRig_Example.unity scene, run the autorig script on the either pair of hands.  In the RiggedHand component for each hand, toggle both SetLeapEditorPose and DeformPositionsInFingers.


Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
```diff
@@ -14,7 +14,7 @@ public class LeapHandsAutoRig : MonoBehaviour {
     [Tooltip("Set to True if each finger has an extra trasform between palm and base of the finger.")] 
     public bool UseMetaCarpals;
     [Tooltip("When True, hands will be put into a Leap editor pose near the LeapServiceProvider's transform.  When False, the hands will be returned to their Start Pose if it has been saved.")]
-    public bool SetEditorLeapPose = false;
+    private bool SetEditorLeapPose = false;
```
Amarcolina: This should be removed completely rather than just made private.

protodeep:
OK done, thanks for the catch.

PR #451, opened 2016-10-12
Support for auto-rigging one hand at a time

protodeep:
- [ ] In the Hands/Examples/Scenes/Rigged_Hands_AutoRig_Example.unity scene, remove one of the hand models under the HandModels transform in the hierarchy.  Add the autorig component, set UseMetaCarpals to True, the hit the Auto Rig button.  Either hand should run fine.

Also 

- re-setting Update When Offscreen to True in the LoPoly_Hand prefabs.

- Removing pushing values for model vectors from the autorig's OnValidate() and putting it into a new, inspector button-triggered method, PushVectorValues(). This allows the values to be set individually and not be overwritten by the autorig script

### DIFF
--- a/Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
+++ b/Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
@@ -61,8 +61,6 @@ namespace Leap.Unity {
         }
       }
       AutoRigByName();
-      RiggedHand_L.StoreJointsStartPose();
-      RiggedHand_R.StoreJointsStartPose();
     }
     /**Allows a start pose for the rigged hands to be created and stored anytime. */
     [ContextMenu("StoreStartPose")]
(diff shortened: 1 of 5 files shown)


Assets/LeapMotionModules/Hands/Scripts/LeapHandsAutoRig.cs
```diff
@@ -91,50 +89,59 @@ public class LeapHandsAutoRig : MonoBehaviour {
...
 
+        modelFingerPointing_L = RiggedHand_L.modelFingerPointing;
+        modelPalmFacing_L = RiggedHand_L.modelPalmFacing;
+
+        RiggedHand_L.StoreJointsStartPose();
+      }
       Transform Hand_R = null;
       foreach (Transform t in transform) {
         if (RightHandStrings.Any(w => t.name.ToLower().Contains(w))) {
+          Debug.Log("Hand_R = " + t.name);
```
jo3w4rd: I frown upon Debug.Log statements in non-error cases.

PR #457, opened 2016-11-10
Adds thread safe queue

Amarcolina:
This PR adds the ProduceConsumeBuffer data structure.  This is a type of lock-free, wait-free, thread-safe queue that supports a single producer and a single consumer.  Only one thread (the producer) is allowed to call TryEnqueue, and only one thread (the consumer) is allowed to call TryDequeue.  

To test:
 - [ ] Make sure tests pass

### DIFF
--- a/Assets/LeapMotion/Scripts/DataStructures/Editor/Tests/ProduceConsumeBufferTest.cs
+++ b/Assets/LeapMotion/Scripts/DataStructures/Editor/Tests/ProduceConsumeBufferTest.cs
@@ -37,7 +37,7 @@ namespace Leap.Unity.Tests {
           TestStruct s;
           s.index = i;
           s.name = i.ToString();
-          while (!buffer.TryPush(ref s)) { }
+          while (!buffer.TryEnqueue(ref s)) { }
         }
       } catch (Exception e) {
         Assert.Fail(e.Message);
@@ -48,7 +48,7 @@ namespace Leap.Unity.Tests {
       try {
         for (int i = 0; i < buffer.Capacity; i++) {
           TestStruct s;
-          while (!buffer.TryPop(out s)) { }
+          while (!buffer.TryDequeue(out s)) { }
 
           Assert.That(s.index, Is.EqualTo(i));
           Assert.That(s.name, Is.EqualTo(i.ToString()));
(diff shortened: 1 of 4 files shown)


Assets/LeapMotion/Scripts/DataStructures/Editor/Tests/ProduceConsumeBufferTest.cs
```diff
@@ -0,0 +1,66 @@
...
+      Thread producer = new Thread(new ThreadStart(producerThread));
+
+      consumer.Start();
+      producer.Start();
+
+      consumer.Join();
+      producer.Join();
+    }
+
+    private void consumerThread() {
```
Sloftin891: what happens if a consumer or producer thread stalls during buffer use? what happens if a thread dies while writing data (part of a string or array)?

Assets/LeapMotion/Scripts/DataStructures/ProduceConsumeBuffer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+  public class ProduceConsumeBuffer<T> {
+    private T[] _buffer;
+    private uint _bufferMask;
+    private uint _head, _tail;
+
+    /// <summary>
+    /// Constructs a new produce consumer buffer of a given capacity.  This capacity
+    /// is fixed and cannot be changed after the buffer is created.  This capacity
+    /// must be a power of two.
```
Sloftin891: Maybe specify why this has to be a power of 2 (buffer mask)?

Assets/LeapMotion/Scripts/DataStructures/Editor/Tests/ProduceConsumeBufferTest.cs
Amarcolina: If a consumer or producer stalls, it is the same as that thread deciding to discontinue writing or reading, which is a safe and legitimate thing for a thread to do.  If a producer stalls, for example, the only thing that will happen is that no more data will be written into the buffer.  It is the consumers responsibility in this case to not get themselves stuck in an infinite loop waiting for data that may never arrive.

Amarcolina: Since the increment of the head or tail index is done *after* the write has been finished, and since the assignment of the new value is atomic, the data structure cannot ever be left in an invalid state.

PR #458, opened 2016-11-28
Small UI Module Housecleaning

jselstad:
This is the version getting sent to QA

Just deleting some unused files and fixing an error in the main UI Module Scene caused by duplicate hand names.


jselstad:
Heyyyyyyy let's get this in; there aren't any code changes.  I just changed a tickbox in the example scene and deleted some unused assets.

PR #460, opened 2016-11-30
Fix Internal Visualizer

jselstad:
Add a ton of duplicate code to do a separate but parallel temporal warp for images (so hands and images can be temporally matched).

Includes grabbing seperate timestamps for images...

### DIFF
--- a/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
+++ b/Assets/LeapMotion/Scripts/LeapServiceProvider.cs
@@ -73,6 +73,9 @@ namespace Leap.Unity {
 
     protected Matrix4x4[] _transformArray = new Matrix4x4[2];
 
+    [NonSerialized]
+    public long imageTimeStamp = 0;
+
     public override Frame CurrentFrame {
       get {
         return _transformedUpdateFrame;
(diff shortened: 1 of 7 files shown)


Assets/LeapMotion/Scripts/LeapServiceProvider.cs
```diff
@@ -220,7 +223,7 @@ public class LeapServiceProvider : LeapProvider {
       } else {
         leap_controller_.Frame(_untransformedUpdateFrame);
       }
-
+      imageTimeStamp = leap_controller_.FrameTimestamp();
```
Amarcolina: Why is FrameTimestamp considered to be the image timestamp?  In LeapImageRetriever we request images based on the timestamp of CurrentFrame.

Assets/LeapMotion/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -157,7 +157,7 @@ protected struct TransformData {
         return false;
       }
 
-      TransformData past = transformAtTime(leapTime - warpingAdjustment * 1000);
+      TransformData past = transformAtTime((leapTime - warpingAdjustment * 1000)+20000);
```
Amarcolina: Where is this new number coming from?

jselstad:
Closing this since it's served its purpose as a PR (showing the diff and whatnot).

PR #463, opened 2016-12-01
Clean up Service Provider for cheap Late Latching/Android

jselstad:
Update the LeapC and LeapCSharp .dlls with better ones (that can extrapolate and get late latching data cheaply).

Clean up the Service Provider and refactor the late-latching section


Amarcolina:
Making a note that we will need to get 32 bit dlls from somewhere

PR #474, opened 2016-12-14
Adding basic package builder

Amarcolina:
I figured since this package manager exists, we might as well use it to help standardize our package building process.  Note that this makes NO CHANGES to our stance on how packages are presented to the user.  All it does is give us a big button we can use to build our packages automatically :)

I'd still like a number of eyes on this though, just to be sure

To test:
 - [x] Make sure you can still build EXE's and that there are no errors
 - [x] Make sure that the build menu appears up in the menu bar and allows you to build packages
 - [x] Make sure you can build all packages.  The first time you build it will prompt you to select a build destination for each package.
 - [x] Make sure you are able to build a single package by the build menu
 - [x] Make sure you are able to build a single package via the build button on the package itself
 - [ ] Make sure you are able to create new packages, and that the inspector interface functions as expected
 - [x] Make sure that the dependencies defined by the packages are correct


jselstad:
Does the package manager belong in the LeapMotionModules folder or in the root Assets folder?

Assets/LeapMotionModules/Package/Editor/PackageDefinitionEditor.cs
```diff
@@ -0,0 +1,211 @@
...
+      string relativePath = Uri.UnescapeDataString(relativeUri.ToString());
+
+      if (toUri.Scheme.Equals("file", StringComparison.InvariantCultureIgnoreCase)) {
+        relativePath = relativePath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+      }
+
+      return relativePath;
+    }
+
+    private void generateBuildMenuScript() {
```
jselstad: It tickles my brain that the menu script generator is shorter than the generated script...

Amarcolina:
I guess it could go in either location, but I like to think of it still as a module of functionality that could theoretically be exported to other locations (even though I'm guessing we won't have any plans to release this)

Assets/LeapMotionModules/Package/Editor/PackageDefinitionEditor.cs
```diff
@@ -0,0 +1,211 @@
...
+        lines.Add("    public static void Build_" + guid + "() {");
+        lines.Add("      PackageDefinition.BuildPackage(\"" + guid + "\");");
+        lines.Add("    }");
+      }
+
+      lines.Add("  }");
+      lines.Add("}");
+      lines.Add("");
+
+      File.WriteAllLines("Assets/LeapMotionModules/Package/Editor/PackageDefinitionBuildMenuItems.cs", lines.ToArray());
```
jselstad: Not critical, just curious; any reason why List<string> over StringBuilder?

nickjbenson:
Building InteractionEngine as a module worked and I assumed everything was fine, but I noticed the .unitypackage failed to contain IE.ldat (which was specified as a dependency in the package), so the resulting IE Example project (+ Core module) didn't work and the unitymodule would have been a dud. I think this is bad behavior for a magic button, so I propose that if any files don't exist that are marked as dependencies, it should fail to build the unitypackage and raise a red Error message in the Console.

Amarcolina:
Nice catch!  I added in a check, if any dependent files are missing, it will pop up an error to alert the user and it will fail to produce a .unitypackage.  I opted for a popup-dialog instead of a console error to make it more visible to the user that something bad has happened.  If the user didn't notice that a build failed, they might go to the build folder like normal and use an old build.

protodeep:
Alright, ran it through its paces here.  Merging.

PR #477, opened 2016-12-15
Add MinValue/MaxValue support for Vector2 and Vector3

nickjbenson:
Because sometimes I want my Vector2s to have minimum nonzero X and Y values, and Vector3s for good measure because why not, and MaxValue too because also why not.

### DIFF
--- a/Assets/LeapMotion/Scripts/Attributes/MaxValue.cs
+++ b/Assets/LeapMotion/Scripts/Attributes/MaxValue.cs
@@ -19,13 +19,19 @@ namespace Leap.Unity.Attributes {
         property.floatValue = Mathf.Min(maxValue, property.floatValue);
       } else if (property.propertyType == SerializedPropertyType.Integer) {
         property.intValue = Mathf.Min((int)maxValue, property.intValue);
-      }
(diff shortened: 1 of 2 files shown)


Assets/LeapMotion/Scripts/Attributes/MaxValue.cs
```diff
@@ -19,13 +19,19 @@ public class MaxValue : CombinablePropertyAttribute, IPropertyConstrainer {
         property.floatValue = Mathf.Min(maxValue, property.floatValue);
       } else if (property.propertyType == SerializedPropertyType.Integer) {
         property.intValue = Mathf.Min((int)maxValue, property.intValue);
-      }
+      } else if (property.propertyType == SerializedPropertyType.Vector2) {
+        property.vector2Value = new Vector2(Mathf.Min(minValue, property.vector2Value.x), Mathf.Min(minValue, property.vector2Value.y));
```
Amarcolina: `property.vector2Value = Vector2.Min(property.vector2Value, Vector2.one * minValue);`
Is a little bit smaller, and can be more easily expanded to Vector3 (and why no love for Vector4? D:)

PR #479, opened 2017-01-13
Add MovingReferenceFrame Option to UIInputModule

jselstad:
Add MovingReferenceFrame Option to UIInputModule under the Advanced Section; check it to support moving reference frames in your UI (at the slight additional cost of retransforming the leap hand data for the EventSystem "Process" callback).

Also ensure that your event system is childed to your moving reference frame to ensure the visual cursors do not lag behind either.


jselstad:
Wrong target branch

PR #480, opened 2017-01-13
Add option to retransform Leap Frames for UI Input

jselstad:
Add MovingReferenceFrame Option to UIInputModule under the Advanced Section; check it to support moving reference frames in your UI (at the slight additional cost of retransforming the leap hand data for the EventSystem "Process" callback).

Also ensure that your event system is childed to your moving reference frame to ensure the visual cursors do not lag behind either.


jo3w4rd:
Tested both UI and Mobile UI example scenes with a motor script and they behaved properly

PR #481, opened 2017-01-13
Line, Circle, Rect, Pivot constraints + example scene

nickjbenson:
John's on JointConstraint logic (two angular DoF, a conical range of motion, optional twist min/max).
PivotConstraint renamed to HingeConstraint, fyi.


Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/FingerType.cs
```diff
@@ -0,0 +1,38 @@
+﻿
+namespace Leap.Unity {
+
+  public static class FingerType {
```
Amarcolina: This whole class seems to be unused?  Furthermore, it seems to only mirror the Leap.Finger.FingerType enum, what is the reason it exists?

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/Hands.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Collections.Generic;
+
+namespace Leap.Unity {
+
+  public static class Hands {
+
+    private static LeapServiceProvider s_provider;
+
+    static Hands() {
+      s_provider = GameObject.FindObjectOfType<LeapServiceProvider>();
```
Amarcolina: This logic will fail under any case where the provider is destroyed and recreated.  The variable s_provider will still reference the old destroyed provider, with nothing able to cause it to be re-initialized to point to the new provider.  Probably the most common case of this would be when switching scenes.

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/Hands.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Returns the average of the index finger and thumb tip positions.
+    /// </summary>
+    public static Vector3 GetPinchPosition(this Hand hand) {
+      return (hand.Fingers[(int)Finger.FingerType.TYPE_INDEX].TipPosition + hand.Fingers[(int)Finger.FingerType.TYPE_THUMB].TipPosition).ToVector3() * 0.5F;
+    }
+
+    /// <summary>
+    /// Returns the fingertip position of the argument finger type, e.g. Leap.FingerType.Index. See Leap.FingerType.
+    /// </summary>
+    public static Vector3 GetTipPosition(this Hand hand, int fingerType) {
```
Amarcolina: I think this method could be replaced with something like Hand.Finger(Finger.FingerType type), that way it could be used in a larger number of cases.  For this case it would just be:

hand.Finger(type).TipPosition;

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/Hands.cs
```diff
@@ -0,0 +1,117 @@
...
+      return hand.Fingers[fingerType].TipPosition.ToVector3();
+    }
+
+    public static bool IsFacing(this Vector3 facingVector, Vector3 fromWorldPosition, Vector3 towardsWorldPosition, float maxOffsetAngleAllowed) {
```
Amarcolina: Would be nice to have a default value for maxOffsetAngleAllowed

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/Hands.cs
```diff
@@ -0,0 +1,117 @@
...
+      Vector3 actualVectorTowardsWorldPosition = (towardsWorldPosition - fromWorldPosition).normalized;
+      return Vector3.Angle(facingVector, actualVectorTowardsWorldPosition) <= maxOffsetAngleAllowed;
+    }
+
+  }
+
+
+
+}
```
Amarcolina: Git apparently thinks you are the devil for no newline at end of file.

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/Util.cs
```diff
@@ -0,0 +1,22 @@
...
+  public static class Util {
+
+    /// <summary>
+    /// Similar to Unity's Transform.LookAt(), but resolves the forward vector of this
+    /// Transform to point away from the argument Transform.
+    /// 
+    /// Useful for billboarding Quads and UI elements whose forward vectors should match
+    /// rather than oppose the Main Camera.
+    /// </summary>
+    public static void LookAwayFrom(this Transform thisTransform, Transform transform) {
```
Amarcolina: Should have the up vector as an optional param.

Assets/LeapMotionModules/UI/LeapMotionHandsExtentionAPI/ValueMappingExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+  public static Vector3 Map(this Vector3 value, float valueMin, float valueMax, float resultMin, float resultMax) {
+    return new Vector3(value.x.Map(valueMin, valueMax, resultMin, resultMax),
+                       value.y.Map(valueMin, valueMax, resultMin, resultMax),
+                       value.z.Map(valueMin, valueMax, resultMin, resultMax));
+  }
+
+  /// <summary>
+  /// Clamps the value between 0 and 1.
+  /// </summary>
+  public static float Clamp01(this float value) {
```
Amarcolina: Adding new functionality is one thing, making it so that you don't need to type Mathf. is a little more debatable :P  I would argue any existing functionality should not be re-routed through extension methods, but would probably be something good to talk about with the rest of the team too.

Assets/LeapMotionModules/UI/Scripts/BoxDepenetrator.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class BoxDepenetrator : MonoBehaviour {
+
+  public SphereCollider sphere;
+
+  [Header("Automatic if null (this)")]
```
Amarcolina: What does this header mean?

Assets/LeapMotionModules/UI/Scripts/BoxDepenetrator.cs
```diff
@@ -0,0 +1,51 @@
...
+  public struct DepenetrationRay {
+    public Vector3 position;
+    public Vector3 direction;
+  }
+  
+  private DepenetrationRay _depenetrationRay;
+  private bool _requiresDepenetrationThisFrame = false;
+
+  void Start() {
+    if (box == null) box = GetComponent<BoxCollider>();
```
Amarcolina: You can use the AutoFind attribute in order to automatically fill the box field with sibling boxColliders.

Assets/LeapMotionModules/UI/Scripts/BoxDepenetrator.cs
```diff
@@ -0,0 +1,51 @@
...
+  [Header("Automatic if null (this)")]
+  public BoxCollider box;
+
+  public struct DepenetrationRay {
+    public Vector3 position;
+    public Vector3 direction;
+  }
+  
+  private DepenetrationRay _depenetrationRay;
+  private bool _requiresDepenetrationThisFrame = false;
```
Amarcolina: Looks like the only reason this variable is not a local variable to update is that it is used in the Gizmos function.  I would recommend having a *separate* variable for use in the gizmos, something like 'didDoDepenetrationLastFrame' or something, and keep the 'requiresDepenetration' logic completely local to the Update function.

Assets/LeapMotionModules/UI/Scripts/ColliderUtil.cs
```diff
@@ -0,0 +1,281 @@
+﻿using UnityEngine;
+using System.Collections;
+using System.Collections.Generic;
+
+public static class ColliderUtil {
```
Amarcolina: All of these methods should be UpperCamelCase

Assets/LeapMotionModules/UI/Scripts/ColliderUtil.cs
```diff
@@ -0,0 +1,281 @@
...
+   */
+  public static Vector3 getPointAtDistance(Vector3 point, Vector3 center, float distance) {
+    return (point - center).normalized * distance + center;
+  }
+
+
+  /* Given a collider and a position relative to the colliders transform, return whether or not the 
+   * position lies within the collider.  Can also provide the allowed distance tolerance
+   */
+  public static bool isInsideCollider(this Collider collider, Vector3 localPosition, float tolerance = 0.0f) {
```
Amarcolina: There is a concept called tolerance here, is this different than the concept called extrude in some other functions?

Assets/LeapMotionModules/UI/Scripts/ColliderUtil.cs
```diff
@@ -0,0 +1,281 @@
...
+    }
+    BoxCollider boxCollider = collider as BoxCollider;
+    if (boxCollider != null) {
+      return boxCollider.isPointInside(localPosition, tolerance);
+    }
+    CapsuleCollider capsuleCollider = collider as CapsuleCollider;
+    if (capsuleCollider != null) {
+      return capsuleCollider.isPointInside(localPosition, tolerance);
+    }
+    Debug.LogWarning("ColliderUtil does not work for Mesh Colliders!");
```
Amarcolina: This warning should probably be an error, and also should not assume that the collider is a mesh collider without checking.  Probably should just use the type of the object instead of hardcoding the name.

Assets/LeapMotionModules/UI/Scripts/ColliderUtil.cs
```diff
@@ -0,0 +1,281 @@
...
+    return Vector3.zero;
+  }
+
+  /* Given a GameObject and a position in global space, return whether or not the position lies within
+   * one of the colliders on the GameObject or one of it's children.  Can also provide the allowed 
+   * distance tolerance
+   */
+  public static bool isInsideColliders(GameObject obj, Vector3 globalPosition, float tolerance = 0.0f) {
+    obj.GetComponentsInChildren<Collider>(false, _colliderList);
```
Amarcolina: I would recommend allowing this method to take in a list of colliders instead of a gameObject, that way we can avoid doing a GetComponent every frame

Assets/LeapMotionModules/UI/Scripts/ColliderUtil.cs
```diff
@@ -0,0 +1,281 @@
...
+      }
+    }
+    return false;
+  }
+
+  /* Given a GameObject and a position in global space, return a point on the surface of one of the
+   * colliders on the GameObject or one of it's children that is closest to the given position
+   */
+  public static Vector3 closestPointOnSurfaces(GameObject obj, Vector3 globalPosition, float extrude = 0.0f) {
+    obj.GetComponentsInChildren<Collider>(false, _colliderList);
```
Amarcolina: Same idea here, list of colliders instead of a gameObject for efficiency (and greater flexibility!), this method will currently count trigger colliders!

Assets/LeapMotionModules/UI/Scripts/Constraints/Abstract/ConstraintBase.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Leap.UI.Constraints {
+
+  public enum Axis   { X, Y, Z }
+  public enum Anchor { Beginning, Center, End }
+
+  public static class ConstraintExtensions {
```
Amarcolina: This looks like it belongs in a generic util class somewhere, not in constraints.

Amarcolina: Oh I see, Axis is a constraint concept only.  Carry on!

Assets/LeapMotionModules/UI/Scripts/Constraints/Abstract/Linear2Constraint.cs
```diff
@@ -0,0 +1,43 @@
...
+  /// <summary>
+  /// Two-dimensional linear constraint.
+  /// </summary>
+  public abstract class Linear2Constraint : PositionConstraint {
+
+    [ConstraintDebugging]
+    [SerializeField]
+    [Range(0, 1)]
+    [Disable] // (Readonly)
+    protected Vector2 _uvCoords = Vector2.zero;
```
Amarcolina: What does _uvCoords represent?

Assets/LeapMotionModules/UI/Scripts/Constraints/HingeConstraint.cs
```diff
@@ -0,0 +1,196 @@
...
+        else {
+          return Mathf.Clamp(angle, MinAngle, MaxAngle);
+        }
+      }
+    }
+
+    private float ComponentOf(Vector3 vector, Axis axis) {
+      switch (axis) {
+        case Axis.X:
+          return (vector.y == 180F && vector.z == 180F ? 180F : 0F) - vector.x * (vector.y == 180F && vector.z == 180F ? 1 : -1F);
```
Amarcolina: This looks like a standard component accessor until I get to Axis.X, what is this logic for?

Assets/LeapMotionModules/UI/Scripts/Constraints/LineConstraint.cs
```diff
@@ -0,0 +1,116 @@
...
+            return this.transform.localScale.y;
+          case Axis.Z: default:
+            return this.transform.localScale.z;
+        }
+      }
+    }
+
+    #region Gizmos
+
+    private const float GIZMO_THICKNESS = 0.01F;
```
Amarcolina: Poor gizmo thickness, no method can take them as an argument so they are left alone.

Assets/LeapMotionModules/UI/Scripts/Constraints/RectConstraint.cs
```diff
@@ -0,0 +1,105 @@
+﻿using Leap.Unity.Attributes;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Leap.UI.Constraints {
+
+  public class RectConstraint : Linear2Constraint {
```
Amarcolina: What is the difference between a Linear2Constraint and a RectConstraint?  What other types of constraints could extend from Linear2Constraint that wouldn't be a RectConstraint?

Assets/LeapMotionModules/UI/Scripts/MinimalBody.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Leap.Unity;
+using UnityEngine;
+
+[ExecuteBefore(typeof(SpringBase))]
```
Amarcolina: We should make a note somewhere to fix the execution order attributes, since they can bug out sometimes still.

Assets/LeapMotionModules/UI/Scripts/SpringBase.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class SpringBase : MonoBehaviour {
```
Amarcolina: Spring base? o.O

Amarcolina:
Also it seems that the extension methods exists inside of the UI folder, shouldn't they exist inside LeapMotion or something?

nickjbenson:
1) Re: ExtensionsAPI things: I only included these because the rest of the things didn't compile without them. They're rough as you can tell, but what I really need to do is separate them out and commit them as a separate PR to develop. Look for that today, and in the mean time, I'll submit another change that makes the extension APIs here in a UI-dedicated-namespace (just for the sake of compiling), to be replaced by the proper extensions once they're approved. So expect those extension comments to be addressed in a new PR targeted for UnityModules/develop.

2) Will go through the non-Extension comments normally. We will most definitely need Execution Order attributes (or to specify execution order some other way, since constraints, springs, and minimalbodies are going to need to execute in a specific order every frame).

3) ColliderUtil is jselstad. ¯\_(ツ)_/¯

jselstad:
I thought Alex wrote ColliderUtil?

Amarcolina:
Oh snap this is my code lol

Amarcolina:
I'll push changes to that file haha.

Assets/LeapMotionModules/UI/Scripts/Constraints/Abstract/Linear2Constraint.cs
nickjbenson: Two dimensional coordinates from 0 to 1. In other words, the same thing UV coordinates usually represent, which is why I named them uvCoords.

Assets/LeapMotionModules/UI/Scripts/Constraints/HingeConstraint.cs
nickjbenson: Because euler angles sometimes return the angle you're looking for but other times it returns the opposite of the angle you're looking for with -180 as the Y and Z components. However, I've removed this getter because it was for an older, less-reliable constraint calculation method, as well as other properties I wrote but that were never used.

jselstad:
Proposing that we close this with great prejudice.

PR #483, opened 2017-01-13
Adding new garbage-free queries 

Amarcolina:
These queries mirror the queries that exist in the System.Linq.Enumerable class.  They produce no garbage (assuming your data structure has an enumerator that is a struct) and are just as fast or faster than normal linq!  Still missing a bunch of ops, but we can add them as we need them.

@nickjbenson 
@jselstad 
@Sloftin891 

To test:
 - [x] Review code to make sure you understand it and it makes logical sense
 - [x] Make sure all unit tests pass


nickjbenson:
Looks good!

Amarcolina:
You gotta approve your review :O

Amarcolina:
Cool, all good to merge in then !

PR #487, opened 2017-01-16
Extensions API + add'l minor conveniences

nickjbenson:
The in-progress UI module already makes use of some of these, but I wanted to have them here in their own PR.

- Added Hands.cs to the Utils folder. Not sure where this should live but maybe not there... I also added a fix that makes changing scenes work, but it's probably not the most elegant solution for having a static LeapRig / ServiceProvider accessor. **This is the primary extensions API part.**
- Added DeltaBuffer.cs, RingBuffer.cs, DelayBuffer.cs to DataStructures. DeltaBuffers have been very convenient for calculating N+1th derivatives for, e.g., gesture recognition. RingBuffers are generic constant-size buffers intended for continuous streams of data. DelayBuffers delay continuous streams of data, and were used to improve throw directionality in Kame.
DeltaBuffer usage:
```
private DeltaBuffer _positionBuffer = new DeltaBuffer(5);
void Update() {
	_positionBuffer.Add(this.transform.position, Time.time);
	if (_positionBuffer.isFull) {
		// 5 frames later...
		Vector3 velocity = _positionBuffer.Delta();
		// do things with velocity information. Even... Add it to another DeltaBuffer!
	}

	// tracking a hand? This is useful:
	if (!hand.isTracked) {
		_positionBuffer.Clear();
	}
}
```
- Added a FingerType static class to the Leap namespace. This is a temporary convenience mapping from "Leap.FingerType.Foo" to "Leap.Finger.FingerType.TYPE_FOO", potentially to be changed to an enum when we modify the LeapCSharp DLL and UnityModule references to Finger.FingerType.TYPE_FOO.
- Merged my Util.cs (LookAwayFrom()) to Scripts/Util/Utils.cs and added some regions for organization.
- Added Scripts/Utils/ValueMappingExtensions.cs, for Map functions.
- Added MinValue and MaxValue attribute support for Vector2, Vector3, Vector4.
- Added .vscode/ to .gitignore (VS Code >> MonoDevelop for OS X Unity dev)


Assets/LeapMotion/Scripts/Utils/Hands.cs
```diff
@@ -0,0 +1,169 @@
...
+namespace Leap.Unity {
+
+  public static class Hands {
+
+    private static LeapServiceProvider s_provider;
+    private static GameObject s_leapRig;
+
+    static Hands() {
+      InitStatic();
+      SceneManager.activeSceneChanged += InitStaticOnNewScene;
```
Amarcolina: So I think that this can catch some of the cases, but it still doesn't catch the general case of swapping out the provider for a different one.  It also would not work if the provider is spawned 1 frame after the scene changes (like how can happen with blocks due to the networking).  

I think the only failsafe way would be some kind of getter that automatically checks to see if the current reference is null, and if so go searching for a new one.-

Assets/LeapMotion/Scripts/Utils/Hands.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    private static void InitStatic() {
+      s_provider = GameObject.FindObjectOfType<LeapServiceProvider>();
+      s_leapRig = s_provider.transform.parent.parent.parent.gameObject;
+    }
+
+    /// <summary>
+    /// Static convenience accessor for the Leap Rig GameObject.
+    /// </summary>
+    public static GameObject Rig {
```
Amarcolina: Is this the camera object or it's parent?  It seems really dangerous to rely on the provider being a certain amount down the hierarchy.

Assets/LeapMotion/Scripts/Utils/Hands.cs
```diff
@@ -0,0 +1,169 @@
...
+    public static Finger Pinky(this Hand hand) {
+      return hand.Fingers[(int)Leap.Finger.FingerType.TYPE_PINKY];
+    }
+
+    /// <summary>
+    /// Returns the direction the Hand's palm is facing.
+    /// For the other two axis directions, see ThumbAxis and FingerAxis.
+    /// </summary>
+    public static Vector3 PalmAxis(this Hand hand) {
+      return -hand.Basis.yBasis.ToVector3();
```
Amarcolina: Isn't this the same as PalmNormal?

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -5,12 +5,23 @@
...
   public static class Utils {
+    
+    /// <summary>
+    /// Similar to Unity's Transform.LookAt(), but resolves the forward vector of this
+    /// Transform to point away from the argument Transform.
+    /// 
+    /// Useful for billboarding Quads and UI elements whose forward vectors should match
+    /// rather than oppose the Main Camera.
+    /// </summary>
+    public static void LookAwayFrom(this Transform thisTransform, Transform transform) {
```
Amarcolina: You should be able to specify what direction is up as an optional parameter.

Assets/LeapMotion/Scripts/Utils/Hands.cs
nickjbenson: Yes. But this completes the HandAxis trifecta. ThumbNormal and FingerNormal would be misleading; e.g. one would assume the ThumbNormal is the normal of the thumb (which it almost never is).

Assets/LeapMotion/Scripts/Utils/Hands.cs
nickjbenson: It's the GameObject with Rig in the name -- the camera's parent. This is the thing you would move if you wanted to translate the player's position in space. I agree, it's really dangerous and we shouldn't assume it's going to be in a given place down the hierarchy, but we should have a static accessor for the VR Camera's parent. How can we reliably get it? Camera.main.transform.parent?

Amarcolina: Possibly search for a camera as a parent of the provider, then get it's parent?  I think that would be at least safer than using multiple parent accessor.  If we cannot find a reliable and safe way of getting it, we should not include it here.

Assets/LeapMotion/Scripts/Utils/Hands.cs
Amarcolina: But I think having multiple methods which return the exact same thing is confusing as well.  I could easily see a beginner trying to figure out what the different is between PalmNormal and PalmAxis.  I'm not sure what the correct naming is, but I definitely think that we should not be duplicating methods for the purpose of matching names for certain situations.

Assets/LeapMotion/Scripts/Utils/ValueMappingExtensions.cs
```diff
@@ -0,0 +1,125 @@
+﻿using UnityEngine;
+
+namespace Leap.Unity {
+
+  public static class ValueMappingExtensions {
```
Amarcolina: Well it's an extension, so being verbose is fine :D you never actually see the name.

Assets/LeapMotion/Scripts/Utils/Hands.cs
jselstad: imo it should step up the hierarchy until it finds an object with a Camera, and then return that one's parent.

And then cache it because that many GetComponents() every frame will annihilate your performance!

Assets/LeapMotion/Scripts/Utils/Hands.cs
nickjbenson: I think the duplication is fine, if we call out in the PalmAxis() doc comment and say "Hand axes are aligned by the normal of the palm, so PalmAxis() equivalent to PalmNormal()." It's less mental overhead than trying to remember "FingerAxis, ThumbAxis, Palm..._Normal_"

Assets/LeapMotion/Scripts/Utils/FingerType.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public static Finger.FingerType Pinky {
+      get {
+        return Finger.FingerType.TYPE_PINKY;
+      }
+    }
+
+  }
+
+}
```
nickjbenson: No, it turns `Finger.FingerType.TYPE_PINKY` into `FingerType.Pinky`(Leap namespace, not a member of Finger).
Dunno about autocomplete, but if you're typing FingerType, it should only show you the one in your namespace, so you'd only be seeing both if you were currently inside the Finger class (which is very rare).
The core idea is just to allow this behavior before the DLL changes but, I can also just remove this and we can just get the DLL changes in (for FingerType.Foo) and none of this FingerType stuff matters anyway.

Assets/LeapMotion/Scripts/Utils/ValueMappingExtensions.cs
nickjbenson: At least in my workflow, if it's the first time I'm using it in a file, I just type myFloat.Map, Ctrl+., Enter (auto-adds "using ValueMappingExtensions;"), then continue typing the function arguments. Or if you're a manual "using" typer, autocomplete will save the day as you write "using ValueM".

Amarcolina: Well ValueMappingExtensions is the name of the class anyway, so no need for long using statements, its all in Leap.Unity :)

Assets/LeapMotion/Scripts/Utils/Hands.cs
Amarcolina: I agree that FingerAxis, ThumbAxis, PalmNormal would not be good either, but maybe we could come up with some different names for ThumbAxis and FingerAxis that would make sense?

Assets/LeapMotion/Scripts/Utils/ValueMappingExtensions.cs
nickjbenson: Woops! Of course.

Assets/LeapMotion/Scripts/Utils/FingerType.cs
nickjbenson: Deleting this in the next commit; we should just change the DLL.

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: Added.

Assets/LeapMotion/Scripts/Utils/Hands.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// <summary>
+    /// Returns a confidence value from 0 to 1 indicating how strongly the Hand is making a fist.
+    /// </summary>
+    public static float FistStrength(this Hand hand) {
+      return (Vector3.Dot(hand.Fingers[1].Direction.ToVector3(), -hand.DistalAxis() )
+            + Vector3.Dot(hand.Fingers[2].Direction.ToVector3(), -hand.DistalAxis() )
+            + Vector3.Dot(hand.Fingers[3].Direction.ToVector3(), -hand.DistalAxis() )
+            + Vector3.Dot(hand.Fingers[4].Direction.ToVector3(), -hand.DistalAxis() )
+            + Vector3.Dot(hand.Fingers[0].Direction.ToVector3(),  hand.RadialAxis() )
```
jselstad: Shouldn't this be -hand.RadialAxis()?

Also why do they still have the parentheses?

Assets/LeapMotion/Scripts/Utils/Hands.cs
nickjbenson: Replaced Rig and Provider getters that check for null and try to become non-null, or throw out a warning and just return null. I also check for null whenever I use those getters in that file.

Assets/LeapMotion/Scripts/DataStructures/DeltaBuffer.cs
```diff
@@ -0,0 +1,152 @@
+﻿using UnityEngine;
+using System.Collections;
+
+/// <summary> Allows you to add to a capped-size ring buffer of Vector3s and,
+/// when full, compute the buffer's average change over time. </summary>
+public class DeltaBuffer {
+
+  private RingBuffer<Vector3> _buffer;
```
Amarcolina: Looks like the code could be simplified if this was just a single buffer of an internal pair-struct instead of two buffers.

Assets/LeapMotion/Scripts/DataStructures/DeltaBuffer.cs
```diff
@@ -0,0 +1,152 @@
...
+      deltaPerTimeSum += (Get(i + 1) - Get(i)) / (GetTime(i + 1) - GetTime(i));
+    }
+    return deltaPerTimeSum / (length - 1);
+  }
+
+}
+
+/// <summary> Allows you to add to a capped-size ring buffer of floats and,
+/// when full, compute the buffer's average change over time. </summary>
+public class DeltaFloatBuffer {
```
Amarcolina: I would really really want to find a way to not have this be a line-by-line copy of DeltaBuffer but with Vector3 replaced with float.  Possibly a generic base class with all of the mathematics pushed into abstract methods, such that DeltaVector3Buffer and DeltaFloatBuffer are subclasses that only need to implement the basic math.

Assets/LeapMotion/Scripts/DataStructures/RingBuffer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+  private T[] arr;
+  private int firstIdx = 0;
+  private int lastIdx = 0;
+
+  public RingBuffer(int bufferSize) {
+    arr = new T[bufferSize];
+  }
+
+  public int Length {
```
Amarcolina: You have a getter here that uses UpperCamelCase but isFull uses lowerCamelCase :(

Assets/LeapMotion/Scripts/Utils/Utils.cs
Amarcolina: You can use a default value if you don't want to create an entire second method.

Assets/LeapMotionModules/DetectionExamples/Scripts/LeapRTS.cs
```diff
@@ -57,39 +37,49 @@ public enum RotationMethod {
...
-//        enabled = false;
-//      }
-
       GameObject pinchControl = new GameObject("RTS Anchor");
       _anchor = pinchControl.transform;
       _anchor.transform.parent = transform.parent;
       transform.parent = _anchor;
     }
 
+    private Hand _lHand;
```
Amarcolina: These Hand vars dont seem to need to be defined as fields, they could be local vars to Update right?

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: No you can't, Vector3.up is not a compile-time constant, nor is new Vector3(0F, 1F, 0F)

Amarcolina: Oh snap you are right!  Carry on :-3

Amarcolina:
Woohoo, all looks good to me!

PR #490, opened 2017-01-18
Add ConnectionMonitor class

### LINKED ISSUE #20: PlugInLeapNotice prefab is broken
3 things:
1. On the LeapC side, IsServiceConnected isn't working because of a bug in LeapC. This has been fixed in platform. -- the fix should be in the assets by now.
2. I am getting null object errors when the prefab is in the scene, so it seems that something might be missing from the prefab?  -- further investigation reveals that at some point, some code was changed to expect a UI.Image instead of GUITexture, but other scripts still expect GUITexture.
3. We shouldn't need to check for a connection every frame. Probably a coroutine that runs every couple of seconds is more efficient.

jo3w4rd:
The ConnectionMonitor class replaces the now deprecated DisconnectionNotice class, which requires a Canvas and performed several operations rather inefficiently (i.e. every frame, even when no state change was occurring). ConnectionMonitor uses a coroutine to monitor the connection and can use any sprite object as the notification image. Also simplified a few other aspects, like not creating its own controller object.

Also removed the DisconnectionNotice prefab, which was broken by changes made to DisconnectionNotice for VRVisualizer, and the embedded disconnection images -- since Orion doesn't support embedded devices. 

Note that I did not create a new prefab. If Frontend Team wants one, Frontend Team can make one.

Fixes #20 (second oldest active issue!)


Amarcolina:
Looks wonderful!

PR #493, opened 2017-01-19
Adding zip and select many

Amarcolina:
Adds zip and select many to the list of ops in the gLINQ pool.

To test
 - [x] Verify unit tests pass.


nickjbenson:
Looks good

PR #496, opened 2017-01-21
Adding fill commands

Amarcolina:
Adding these fill commands because it's about time!  I need this basically every day and I just end up writing for loops forever.  So now they are in finally uhg.


nickjbenson:
Looks good.

PR #543, opened 2017-02-02
Moved SingleLayer and ReadonlyList to core out of IE

Amarcolina:
Both SingleLayer and ReadonlyList are useful, but for some reason exist inside of the IE instead of core.  This PR moves them into core so they can be used by everybody!

To test:
 - [ ] No compile errors due to move!
 - [ ] Now you know they both exist!


nickjbenson:
Lovely. Thanks for the extra docs!

PR #548, opened 2017-02-03
Implement Hand Post Processes

jselstad:
In the Hand Pool, you'll now see a Unity Event called "Hand Post Processes" where you can add functions that will modify that hand group right before it gets displayed, without modifying the frame or generating garbage.

How is it implemented?   Sneakily...

An Example Hand Post Process function looks like this:
```
 public void setHeldHand(Hand inHand) {
    inHand.SetTransform(transform.position, transform.rotation);
  }
```

- [x] Verify that it doesn't generate garbage
- [x] Verify that there aren't side-effects
- [x] Verify that its totally awesome


Amarcolina:
I'm not going to list it as a blocker to merging this PR, but just making a note that we are absolutely going to want a way to do a post process on multiple models at the same time without doing the post process multiple times, since copying a hand is so expensive.

Assets/LeapMotion/Scripts/HandPool.cs
```diff
@@ -51,6 +51,11 @@ public class ModelGroup {
       public List<IHandModel> modelsCheckedOut;
       public bool IsEnabled = true;
       public bool CanDuplicate;
+
+      [System.Serializable]
+      public class HandEvent : UnityEvent<Hand> { }
```
Amarcolina: Would be great to move this somewhere else (maybe inside the HandUtil.cs file?) so that it can more easily be references by other classes wanting to do the same thing.  For example there is a class InteractionHandEvents that defines the exact same event type, would be great to refactor it to use this new class you created from core.

Assets/LeapMotion/Scripts/HandRepresentation.cs
```diff
@@ -14,12 +14,15 @@ public abstract class HandRepresentation
     public Chirality RepChirality { get; protected set;}
     public ModelType RepType { get; protected set;}
     public Hand MostRecentHand { get; protected set; }
+    public Hand PostProcessHand { get; set; }
+    public HandPool.ModelGroup Group { get; set; }
```
Amarcolina: We talked about this in the issue about this abstraction, but just re-iterating here.  HandRepresentation is part of the underlying abstraction, and HandPool is an implementation of that abstraction, thus HandRepresentation is not technically allowed to know about HandPool or ModelGroup.

However, since this has been a hassle for a long while, we should resolve the abstraction decision first, and then this should become more straightforward.

jselstad:
Alex's comments are addressed in this next PR I have up.

PR #556, opened 2017-02-06
Adding misc. features and utilities

Amarcolina:
- CustomEditorBase now allows the script field to be hidden
 - CustomEditorBase now allows easy creation of horizontal fields in normal inspectors.
 - Renamed FillExtensions to ListAndArrayExtensions because it is just amassing a bunch of small utils for lists and arrays
 - Added variants of FillEach that take the index as an argument into the generator delegate
 - Added RemoveUnordered when you want to remove an element from a list in O(1) and order is not important.
 - Added RemoveAtMany for when you want to remove a bunch of elements from a list all at once.  Is O(N + M) instead of the O(N*M) that comes from calling RemoveAt over and over again.
 - Assed InsertMany for when you want to insert a bunch of elements into a list all at once.  Is O(N + M) instead of the O(N*M) that comes from calling Insert over and over again.
 - Added some asserts for asserting runtime vs edit time.


Assets/LeapMotion/Scripts/Query/Editor/ListAndArrayExtensionTests.cs
```diff
@@ -0,0 +1,50 @@
...
+      toRemove.Add(99);
+
+      doRemoveAtManyTest(toRemove);
+    }
+
+    private void doRemoveAtManyTest(List<int> toRemove) {
+      List<int> listA = new List<int>().FillEach(100, i => i);
+      List<int> listB = new List<int>(listA);
+
+      var stopwatch = new System.Diagnostics.Stopwatch();
```
Amarcolina: Oh snap forgot to remove it!

PR #562, opened 2017-02-10
Gui Renderer MVP

Amarcolina:
This is the MVP release for the leap gui rendering system.  The goals of this system are to:
 - Define a reasonable but powerful definition for non-rectangular spaces
 - Allow the user to place gui elements within these spaces, as well as easily allowing switching between spaces.
 - Run at a high-enough performance to be mobile-viable
 - Be extensible enough to support any future optimizations or features needed.

The current features of the system are:
 - Cylindrical and Spherical space definitions (not all transformation functions present currently, but base functions are).  
 - Feature based rendering method for highest modularity and flexibility
 - A baked renderer that bakes the entire gui into a single mesh.
 - A dynamic renderer that renders each element as a different mesh (but ensures dynamic batching is possible)
 - A mesh feature, which allows elements to have a mesh based graphical representation (with plugs for procedural mesh definitions as well)
 - A texture feature, which allows elements to map arbitrary textures onto meshes, including automatic texture packing into an atlas.
 - A sprite feature, which allows sprite metadata to be taken advantage of, as well as using the existing unity atlas bake pipeline for sprites.
 - A tint feature, which allows elements to be tinted dynamically at runtime for low cost.
 - A blend shape feature, which allows elements to be deformed dynamically at runtime for low cost.
 - Serialization that mirrors unity's default, so that prefabs and duplication just work as you would expect.

I have tried to cover up or disable features that are not yet implemented, but I have no doubt there are bugs, errors, or cases where the system does something it should not!  If it is critical to the MVP that it is fixed, post it here and I will address, but if not, open an issue and i will fix it in a later pr.

ALSO, this is currently 5.5 only, there are some issues with 5.6 that I will fix in a later pr.


Assets/LeapMotionModules/ElementRenderer/Resources/DynamicRenderer.cginc
```diff
@@ -0,0 +1,207 @@
...
+
+float4 _LeapGuiCurved_ElementParameters[ELEMENT_MAX];
+float4x4 _LeapGui_LocalToWorld;
+
+void ApplyGuiWarping(inout float4 anchorSpaceVert, int elementId) {
+  float4 parameters = _LeapGuiCurved_ElementParameters[elementId];
+
+  Cylindrical_LocalToWorld(anchorSpaceVert.xyz, parameters);
+
+  anchorSpaceVert = mul(_LeapGui_LocalToWorld, vert);
```
jselstad: This line is yelling at me when I try to switch to a dynamic renderer 😨

`Shader error in 'LeapGui/Defaults/DynamicTransparent': 'mul': no matching 2 parameter intrinsic function; Possible intrinsic functions are: mul( *lots of bullshit* )  at Assets/LeapMotionModules/ElementRenderer/Resources/DynamicRenderer.cginc(28) (on d3d11)`

`Shader error in 'LeapGui/Defaults/DynamicTransparent': undeclared identifier 'vert' at Assets/LeapMotionModules/ElementRenderer/Resources/DynamicRenderer.cginc(28) (on d3d11)`

Have you tried switching your build target back to PC, away from Android recently?

Assets/LeapMotionModules/ElementRenderer/Scripts/Spaces/LeapGuiCylindricalSpace.cs
```diff
@@ -0,0 +1,149 @@
...
+      angleOffset = 0,
+      heightOffset = 0,
+      radiusOffset = radius,
+      radiansPerMeter = 1.0f / radius
+    });
+  }
+
+  public override void RefreshElementData(Transform root) {
+    //TODO!
+    //refreshElementData(root, root);
```
jselstad: 😨

Assets/LeapMotionModules/ElementRenderer/Scripts/Spaces/LeapGuiSphericalSpace.cs
```diff
@@ -0,0 +1,179 @@
...
+      angleXOffset = 0,
+      angleYOffset = 0,
+      radiusOffset = radius,
+      radiansPerMeter = 1.0f / radius
+    });
+  }
+
+  public override void RefreshElementData(Transform root) {
+    //TODO!
+    //refreshElementData(root, root);
```
jselstad: 😨

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
```diff
@@ -0,0 +1,159 @@
...
+
+    base.buildTopology(meshData);
+  }
+
+  protected override void postProcessMesh() {
+    base.postProcessMesh();
+
+    //For the baked renderer, the mesh really never accurately represents it's visual position 
+    //or size, so just disable culling entirely by making the bound gigantic.
+    _currMesh.bounds = new Bounds(Vector3.zero, Vector3.one * 100000);
```
jselstad: 🤔

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Tint/LeapGuiTintFeature.cs
```diff
@@ -0,0 +1,16 @@
...
+[AddComponentMenu("")]
+[LeapGuiTag("Tint")]
+public class LeapGuiTintFeature : LeapGuiFeature<LeapGuiTintData> {
+  public const string FEATURE_NAME = LeapGui.FEATURE_PREFIX + "TINTING";
+
+#if UNITY_EDITOR
+  public override void DrawFeatureEditor(Rect rect, bool isActive, bool isFocused) { }
+
+  public override float GetEditorHeight() {
+    return 0;
```
jselstad: 😨

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Sprite/LeapGuiSpriteFeature.cs
```diff
@@ -0,0 +1,60 @@
+﻿using UnityEngine;
+using UnityEngine.Rendering;
+#if UNITY_EDITOR
+using UnityEditor;
+using UnityEditor.Sprites;
+#endif
+
```
jselstad: Doesn't check for an existing Texture feature 😨 

(and visa versa ( 😨 ))

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 5.5.0f3
+m_EditorVersion: 5.5.1f1
```
Amarcolina: Is 5.51f1 bad? D: It's the most recent version of unity.

Amarcolina: Hmm I cannot replicate this, could you paste the error?

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
Amarcolina: Yeah I wish there was an easier way to disable culling :/  As far as I can tell this is the solution people keep coming back to, at least until Unity adds some sort of api to do it properly/directly...

Assets/LeapMotionModules/ElementRenderer/Resources/DynamicRenderer.cginc
Amarcolina: Ah, nice catch.  Only triggered when using a cylindrical space.

Assets/LeapMotionModules/ElementRenderer/Scripts/Spaces/LeapGuiCylindricalSpace.cs
Amarcolina: I should have made a note, this method is a stub for a more efficient means of refreshing element space data, but there is no logic in LeapGui for handling this, so it just calls BuildElementData every frame to rebuild everything.  In another PR I will add the more efficient means of updating element data.

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Tint/LeapGuiTintFeature.cs
Amarcolina: Intentional, the Tint feature has no feature options, so it's height is zero :D

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Sprite/LeapGuiSpriteFeature.cs
Amarcolina: Mmmmm yeah looks like there should be some additional validation.  It's legal to have both a sprite feature and a texture feature at the same time, but not if they both share the same UV channel.

Assets/LeapMotionModules/ElementRenderer/Resources/SphericalSpace.cginc
```diff
@@ -0,0 +1,21 @@
+
+float _LeapGuiRadialSpace_Radius;
+
+// x : degree X offset
+// y : degree Y offset
+// y : degrees per meter
```
Amarcolina: Nice catch, fixed.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
```diff
@@ -0,0 +1,159 @@
...
+  private List<Vector4> _curved_elementParameters = new List<Vector4>();
+
+  #endregion
+
+  protected override void OnValidate() {
+    base.OnValidate();
+
+    //Currently full motion is not supported
+    if (_motionType == MotionType.Full) {
+      _motionType = MotionType.Translation;
```
Amarcolina: Just that it's not supported?  Or is there a logic error I'm missing.

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Sprite/LeapGuiSpriteFeature.cs
Amarcolina: Ok this should be fixed now.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
jselstad: Just wanted to think about whether or not we're incurring some equally massive cost with the massive bounding box...

Probably not, but you never know with Unity...

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
jselstad: Just odd that we have the option but not the ability to set to it...

I understand if you put it there so you wouldn't forget about it later, but often times I find it's the opposite (placeholder UI elements, by virtue of existing, tend to not get followup implementations)...

so idk just a personal observation

ProjectSettings/ProjectVersion.txt
jselstad:
```
KeyNotFoundException: The given key was not present in the dictionary.
System.Collections.Generic.Dictionary`2[UnityEngine.Transform,LeapGuiSphericalSpace+Transformer].get_Item (UnityEngine.Transform key) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/Dictionary.cs:150)
LeapGuiSphericalSpace.GetTransformer (UnityEngine.Transform anchor) (at Assets/LeapMotionModules/ElementRenderer/Scripts/Spaces/LeapGuiSphericalSpace.cs:39)
LeapGuiBakedRenderer.OnUpdateRenderer () (at Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs:77)
LeapGui.doLateUpdateRuntime () (at Assets/LeapMotionModules/ElementRenderer/Scripts/LeapGui.cs:333)
LeapGui.LateUpdate () (at Assets/LeapMotionModules/ElementRenderer/Scripts/LeapGui.cs:281)
```

To reproduce, just open the FeatureTest scene, hit play, and disable the Anchor of Constant Size component on the Anchor object.

jselstad: idk just that it changed :p

Assets/LeapMotionModules/ElementRenderer/Scripts/Spaces/LeapGuiCylindricalSpace.cs
jselstad: Okay, was just curious if this had anything to do with disabling the anchors.

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Tint/LeapGuiTintFeature.cs
jselstad: Ah, it wasn't obvious to me that the features here magically propagated to the elements (where they do have a Tint Color parameter).

Hmm, probably my first major stumble with the architecture of the system and how features are enabled or disabled...

ProjectSettings/ProjectVersion.txt
Amarcolina: Ooooh at runtime, sorry I missed that part D:  Yes there is no support for changing the hierarchy at runtime, which includes enabling/disabling the anchors of constant size.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
Amarcolina: We could probably have some sort of way to have the user define a bounding box if it becomes a problem, kinda like skinned mesh renderers, which have a similar problem.  We can add that in another PR if it ends up being an issue.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
Amarcolina: That makes sense to me, I'll go ahead and remove it from the dropdown for now, there already is a ticket for it.

ProjectSettings/ProjectVersion.txt
Amarcolina: I'm going to keep it :D

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Tint/LeapGuiTintFeature.cs
Amarcolina: The feature class controls the global settings, and is visible on the LeapGui component.  The data class controls the per-element settings, and is visible on the LeapGuiElement component.  

Whenever a gui element becomes a child of a gui, the gui manually goes through elements and makes sure that they have the data needed to support the features the gui has.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiBakedRenderer.cs
Amarcolina: Ok removed.

ProjectSettings/ProjectVersion.txt
jselstad: Maybe throw a more informative assert?

Lest other wayward devs meet the same fate...

Amarcolina: Ok, the system should now throw an assert in most cases.  The one thing it will not detect is if an anchor is created or added at runtime, since it would require extra processing overhead to detect such a case.

Assets/LeapMotionModules/ElementRenderer/Scripts/SupportInfo.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+
+[Serializable]
+public struct SupportInfo {
```
nickjbenson: When I first saw this file in the root scripts folder it struck me as as "Support" as in "Help" rather than "Do I support this", and it stuck out. Maybe something like FeatureSupportInfo?

Assets/LeapMotionModules/ElementRenderer/Scripts/Utility/GuiRectUtil.cs
```diff
@@ -0,0 +1,46 @@
+﻿using UnityEngine;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+public static class GuIRectUtil {
```
nickjbenson: Capitalization

Assets/LeapMotionModules/ElementRenderer/Scripts/Utility/LeapGuiComponentBase.cs
```diff
@@ -0,0 +1,20 @@
+﻿using UnityEngine;
+
+[ExecuteInEditMode]
+public class LeapGuiComponentBase<AttatchedComponent> : MonoBehaviour
```
nickjbenson: Spelling

Assets/LeapMotionModules/ElementRenderer/Resources/BakedRenderer.cginc
```diff
@@ -0,0 +1,218 @@
+#include "Assets/LeapMotionModules/ElementRenderer/Resources/LeapGui.cginc"
```
Amarcolina: Actually I'm not 100% sure.  I think so, I think that if they are not, then Unity will fail to compile the shader during export.  I will have to verify and get back to you.

Assets/LeapMotionModules/ElementRenderer/Resources/LeapGui.cginc
```diff
@@ -0,0 +1,3 @@
+#include "UnityCG.cginc"
+
+#define ELEMENT_MAX 32
```
Amarcolina: 3 line file for 2 reasons.  First, it gives us a good place to put functions that are common (even though it isn't done yet), but it also gives us a central place to set the element max (and also set it manually through the regex method later!)

Amarcolina: And yeah, currently a max of 32, but that is an arbitrary number that is GPU side only.  It can be increased here without any need to change anything CPU side (for now)

Assets/LeapMotionModules/ElementRenderer/Scripts/Editor/LeapGuiElementEditor.cs
```diff
@@ -0,0 +1,105 @@
...
+using UnityEditor;
+using Leap.Unity;
+using Leap.Unity.Query;
+
+[CanEditMultipleObjects]
+[CustomEditor(typeof(LeapGuiElementData), editorForChildClasses: true, isFallback = true)]
+public class DefaultFeatureDataEditor : CustomEditorBase {
+  protected override void OnEnable() {
+    base.OnEnable();
+    dontShowScriptField();
```
Amarcolina: Ah that is right, I was just following the style that already existed in CustomEditorBase, but that file was created a long time ago and needs to be updated to the new style guide !

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/BlendShape/Editor/LeapGuiBlendShapeDataEditor.cs
```diff
@@ -0,0 +1,19 @@
...
+
+[CanEditMultipleObjects]
+[CustomEditor(typeof(LeapGuiBlendShapeData))]
+public class LeapGuiBlendShapeDataEditor : CustomEditorBase {
+  protected override void OnEnable() {
+    base.OnEnable();
+
+    if (target == null) return;
+
+    dontShowScriptField();
```
Amarcolina: I also agree that it is annoying, and the symptom is usually a custom editor script that is not properly drawing it.  One of the things CustomEditorBase does is make sure it is drawn correctly.

However, in this case the custom editor is being embeded into the inspector for LeapGuiElement,  and I thought that it would be annoying and visually noise to have a script field for every enabled feature, so I disabled.

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Mesh/ProceduralMeshSource.cs
```diff
@@ -0,0 +1,16 @@
...
+using UnityEngine.Rendering;
+
+[RequireComponent(typeof(LeapGuiElement))]
+public abstract class ProceduralMeshSource : MonoBehaviour {
+
+  // Just to make sure all procedural mesh sources can be enabled/disabled
+  protected virtual void Start() { }
+
+  /// <summary>
+  /// Tried so generate a procedural mesh for the given mesh feature.
```
Amarcolina: Ah yeah that is a little unintuitive.  The system assumed that all uvs map to a texture, and so when the texture gets atlased, the uvs need to be remapped.  This flag allows you to enabled/disable that feature of remapping.  Maybe changing the name to include the word 'atlas' would help too.

Assets/LeapMotionModules/ElementRenderer/Scripts/SupportInfo.cs
Amarcolina: It was originally called that, but Feature is a specific thing in this system, and I wanted to communicate that this class was for more than just supporting features, you could also give renderer support, or space support.

Assets/LeapMotionModules/ElementRenderer/Scripts/Utility/GuiRectUtil.cs
Amarcolina: Oh snap good eye!

Assets/LeapMotionModules/ElementRenderer/Scripts/Utility/LeapGuiComponentBase.cs
Amarcolina: Oh snap good catch thanks.

Assets/LeapMotionModules/ElementRenderer/Scripts/Renderers/LeapGuiDynamicRenderer.cs
```diff
@@ -0,0 +1,127 @@
...
+  public override void OnUpdateRenderer() {
+    base.OnUpdateRenderer();
+
+    if (gui.space is LeapGuiRectSpace) {
+      using (new ProfilerSample("Draw Meshes")) {
+        for (int i = 0; i < gui.elements.Count; i++) {
+          Graphics.DrawMesh(_meshes[i], gui.elements[i].transform.localToWorldMatrix, _material, 0);
+        }
+      }
+    } else if (gui.space is LeapGuiRadialSpaceBase) {
```
Amarcolina: 👾

Assets/LeapMotionModules/ElementRenderer/Scripts/Editor/LeapGuiElementEditor.cs
Amarcolina: Hmm since it touches so many non-gui files, I think I'll change this in a different PR into develop instead of on this PR which is already pretty big.

nickjbenson:
One more request: The Unity Editor is leaking memory when I rotate the Anchor object in the example scene. It doesn't seem like it's as bad as it was before but is there a way we can prevent that entirely?

PR #575, opened 2017-02-15
Adding ability to modify maximum element amount

Amarcolina:
This adds the ability to change the maximum element count, as well as the ability for the gui to emit warnings whenever you are over the limit.

To test:
 - [ ] Visit the Edit->Preferences->Leap Gui dialog and verify that you can change the maximum element count
 - [ ] Verify that the LeapGui inspector tells you when you have too many elements
 - [ ] Verify that a LeapGuiElement inspector tells you whenever it is not able to fit into the current element count.

### DIFF
--- a/Assets/LeapMotionModules/ElementRenderer/Resources/LeapGui.cginc
+++ b/Assets/LeapMotionModules/ElementRenderer/Resources/LeapGui.cginc
@@ -1,3 +1,5 @@
 #include "UnityCG.cginc"
 
+// This line is auto-generated, do not modify!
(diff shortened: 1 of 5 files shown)


nickjbenson:
Warning says "Visit the preferences to change the maximum element count" -- could we reword to "Visit Edit->Preferences to change the maximum element count" -- lower-case preferences might sound like the generic form of the word and people might look for it on the component instead of the Edit menu

Amarcolina:
Good idea, made the change

PR #584, opened 2017-02-17
Compatibility with 5.6

Amarcolina:
This upgrades core so that it is compatible with 5.6, while also maintaining compatibility with 5.5!  

To test: 
 - [x] Verify that there are no compile errors or compile warnings on either 5.5 or 5.6
 - [x] Verify that opening the project in either 5.5 or 5.6 does not trigger a 'script update' due to outdated scripts.
 - [ ] Verify that our regular scenes still work as expected


jselstad:
The AR example scene does not display in the left eye in 5.6b6.

Possibly need to redo the shaders.

Amarcolina:
Unfortunately this looks like a problem with the beta and not our shaders :(  When I just have an empty non-leap scene with 2 cameras, one rendering to the right eye and one rendering to the left eye, I get the same issue.

jselstad:
Known issue, fixed in release: https://issuetracker.unity3d.com/issues/vr-when-having-two-cameras-in-a-scene-only-the-right-cameras-view-is-rendered?_ga=1.109245126.237557156.1485895563

PR #585, opened 2017-02-17
Better procedural panels

Amarcolina:
This PR adds a lot more features to the procedural panels, including:
 - The panels can now optionally use a RectTransform to define the size and position of the panel
 - The panel resolution can be expressed in one of two ways:  Total vertices, or vertices per meter.
 - The panel automatically prevents the resolution from getting so high that mesh errors are generated.
 - The panel disables the nine-slice option if nine slicing is not available.  
 - The panel has a button to add a RectTransform if one does not exist.  When creating, it will ensure that the new RectTransform matches the current size.

The mesh feature editor will also now disable it's mesh and channel fields if it is currently being driven by a procedural mesh component.


Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Mesh/ProceduralMeshes/ProceduralPanel.cs
```diff
@@ -4,35 +4,80 @@
 using Leap.Unity.Attributes;
 
 public class ProceduralPanel : ProceduralMeshSource {
```
jselstad: In keeping with the naming convention, it might be good to add "LeapGui" or "LeapElement" as a prefix to the procedural panel; keeps it from being confused with other non-Leap Element Renderer Procedural Panel/Mesh Generators

Assets/LeapMotionModules/ElementRenderer/Scripts/Features/Mesh/ProceduralMeshes/ProceduralPanel.cs
```diff
@@ -118,4 +175,9 @@ public class ProceduralPanel : ProceduralMeshSource {
       return (dv / (vertCount - 1.0f)) * size;
     }
   }
+
+  public enum ResolutionType {
+    Vertices,
+    VerticesPerMeter
```
jselstad: Vertices per meter was a little unintuitive until I recognized that it was vertices per _rectilinear_ meter (ideally it would be "vertices per radian/degree" but I recognize this makes it dependent on the base space and its position within it so vertices per rectilinear meter is a fine approximation)

PR #594, opened 2017-03-03
Fix procedural panels

Amarcolina:
This pr includes some fixes and improvements to the procedural panel:
 - You can now use the panel without using a sprite, nine slicing will automatically be disabled with a helpful tooltip 
 - Fixed a bug where the panel would assume that the mesh used channel 0 for it's uvs
 - With multiple texture or sprites are available, a dropdown will appear to allow you to select which feature you want to use as a source.
 - Fixed longstanding bug where editor objects were not cleaned up properly, leading to seemingly random null reference errors.

To test:
 - [ ] Play around with the panel component, it should work as before
 - [ ] Add a sprite and a texture feature, note how there is now a dropdown that lets you select which one you want to use as a source (shader only displays uv0 right now so anything but that might be weird, but dropdown works!)
 - [ ] Make sure that selecting multiple panels at the same time doesn't break the editor


nickjbenson:
Merging. We'll bring in a fix later involving autogen-ing standard and unlit shaders so that property names / UVs are correctly coupled in the shader.

Issue #563, opened 2017-02-11
Allow texture atlases to be compressed


Amarcolina:
Currently texture atlases are created using the Texture2D function Pack(), which always turns the texture into format ARGB32, which is not ideal for many cases.  There currently seems to be no way to compress a texture to a custom format without turning it into an asset and importing it??  I hope there is a way to get around this that ISN'T writing out PNGs to file so they can be re-imported by Unity.

Amarcolina:
Cool, there is a method EditorUtility.CompressTexture which will allow us to compress the atlas automatically!

PR #595, opened 2017-03-04
Allowed non-readable textures to be read without reimporting!

Amarcolina:
This is totally a hack to read unreadable textures at edit time.  Basically blit them onto a rendertexture and then read the data back off of that rendertexture into a readable texture.  Unity is very silly.

To test: 
 - [ ] Verify you can use any texture (that is not compressed) in a texture field!  Also verify that this does not change the import settings of the texture.


Assets/LeapMotionModules/ElementRenderer/Scripts/Utility/PackUtil.cs
```diff
@@ -96,31 +97,26 @@ public class Settings {
...
 
     Texture2D bordered;
-    BorderKey key = new BorderKey() { texture = source, border = border };
-    if (!_cachedBorderedTextures.TryGetValue(key, out bordered)) {
-      source.EnsureReadWriteEnabled();
-      bordered = source.GetBordered(border);
-      _cachedBorderedTextures[key] = bordered;
+    ProcessKey key = new ProcessKey() { texture = source, border = border };
+    if (!_cachedProcessedTextures.TryGetValue(key, out bordered)) {
+      var tempReadable = source.GetReadableTexture();
```
nickjbenson: So I'm not certain but it looks like you're always doing this blit hack on every texture being selected. Is that true? We should only be re-blitting textures if they're not readable. Otherwise we're hanging the editor unnecessarily when the user scrolls over a big texture in their editor texture list. (I experienced this behavior and want to make sure it isn't because we're blitting every time.) It's also really problematic to blit every time because you're only ever using ARGB32 and sometimes I want different texture compression.

Amarcolina: Makes sense, I'll add in a clause that allows GetReadableTexture to return the texture itself if it is readable.  As for the ARGB32 issue, currently that is a known issue (#563), because Texture2D.Pack changes the texture format to argb32 no matter what the input textures are :(

ajohnston33:
Non-readable textures are in a compressed format.  They wouldn't want to do this for free because it would silently eat draw calls.

Amarcolina:
Yeah that's true, but it would be nice if they provided some sort of api at *edit time*, where all the information is available anyway.

PR #599, opened 2017-03-06
Only rebuild when dirty

Amarcolina:
This PR adds even greater efficiency to the gui renderer.  Specifically do NOT do a rebuild of the entire gui system unless there has been a change that will actually affect the system.  This is done through a combination of per-element dirty flags, specific callbacks, and a hash of the entire hierarchy state.

To test, verify that all of these changes are visualized right away:
 - [x] Adding/Removing/Changing a feature
 - [x] Switching the space
 - [x] Switching the renderer
 - [x] Moving a gui element (or any transform that is part of the gui)
 - [x] Making a change to a gui element
 - [ ] Making a change to a procedural mesh component

ALSO: Try changing something else in the scene like the strength of a light, and use the profiler to verify that the mesh is not being baked!


jselstad:
Looks good.

PR #602, opened 2017-03-10
Adding edit time only attribute

Amarcolina:
Adding a super basic attribute that allows us to tag fields that should only be editable at edit time.

To test:
 - [ ] Make sure it works as expected.


nickjbenson:
Cool. I've needed this before. mergymerge

PR #604, opened 2017-03-11
Feature gizmo capsules

nickjbenson:
Add capsule drawing for Runtime Gizmos!

To test, create a scene with some capsule colliders and add a runtime collider gizmo component, and make sure the capsules look right. (You need a runtime gizmo manager to see them, too.)

### DIFF
--- a/Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
+++ b/Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
@@ -524,11 +524,68 @@ namespace Leap.Unity.RuntimeGizmos {
     /// <summary>
     /// Draws a wire gizmo circle at the given position, with the given normal and radius.
     /// </summary>
-    public void DrawWireCirlce(Vector3 center, Vector3 direction, float radius) {
+    public void DrawWireCircle(Vector3 center, Vector3 direction, float radius) {
       PushMatrix();
       matrix = Matrix4x4.TRS(center, Quaternion.LookRotation(direction), new Vector3(1, 1, 0)) * matrix;
       DrawWireSphere(Vector3.zero, radius);
       PopMatrix();
+    }
+
+    private void DrawLineWireCircle(Vector3 center, Vector3 normal, float radius) {
+      DrawWireArc(center, normal, radius, 26, 26);
+    }
+
+    private void DrawWireArc(Vector3 center, Vector3 normal, float radius, int numCircleSegments = 26, int numSegmentsToDraw = 26) {
+      normal = normal.normalized;
+      Vector3 radiusVector = Vector3.Slerp(normal, -normal, 0.5F) * radius;
+      Vector3 nextVector;
+      for (int i = 0; i < numSegmentsToDraw; i++) {
+        nextVector = Quaternion.AngleAxis(360F / numCircleSegments, normal) * radiusVector;
+        DrawLine(center + radiusVector, center + nextVector);
+        radiusVector = nextVector;
+      }
+    }
+
+    /// <summary>
+    /// Draws a wire gizmo capsule at the given position, with the given start and end points and radius.
+    /// </summary>
+    public void DrawWireCapsule(Vector3 start, Vector3 end, float radius) {
+      Vector3 up = (end - start).normalized * radius;
+      Vector3 forward = Vector3.Slerp(up, -up, 0.5F);
+      Vector3 right = Vector3.Cross(up, forward).normalized * radius;
+
+      float height = (start - end).magnitude;
+      float sideLength = Mathf.Max(0, (height * 0.5F) - radius);
+      Vector3 middle = (end + start) * 0.5F;
+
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -524,13 +524,60 @@ public class RuntimeGizmoDrawer {
...
+      DrawWireArc(start, forward, -right, radius, 26, 13);
+      DrawWireArc(end, right, -forward, radius, 26, 13);
+      DrawWireArc(end, forward, right, radius, 26, 13);
+    }
+
+    private void DrawLineWireCircle(Vector3 center, Vector3 normal, float radius) {
+      DrawWireArc(center, normal, Vector3.Slerp(normal, -normal, 0.5F), radius, 26, 26);
+    }
+
+    private void DrawWireArc(Vector3 center, Vector3 normal, Vector3 radialStartDirection, float radius, int numCircleSegments = 26, int numSegmentsToDraw = 26) {
```
Amarcolina: I think that instead of passing in two segment arguments, it would be better to pass in first a 'percent filled' argument that defaults to 1.0f, followed by the 'numSegments' argument that has some default that can draw reasonably accurate circles.  

In this way, a user can draw a specific percentage of a circle without needing to do math, or know what the default number of segments are.  They could also up the resolution of the circle without needing to change the number of segments to draw.

Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -524,13 +524,60 @@ public class RuntimeGizmoDrawer {
...
+      normal = normal.normalized;
```
Amarcolina: Also, this is a super tiny thing, but I think that the default segment count should be divisible by 4 so that by default you can represent 25%, 50%, and 75% filled perfectly XD

Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -524,13 +524,60 @@ public class RuntimeGizmoDrawer {
     /// <summary>
     /// Draws a wire gizmo circle at the given position, with the given normal and radius.
     /// </summary>
-    public void DrawWireCirlce(Vector3 center, Vector3 direction, float radius) {
+    public void DrawWireCircle(Vector3 center, Vector3 direction, float radius) {
       PushMatrix();
       matrix = Matrix4x4.TRS(center, Quaternion.LookRotation(direction), new Vector3(1, 1, 0)) * matrix;
```
Amarcolina: Could you include this fix in this PR as well?  This is the matrix fix to address the issue where the circle appears stuck to the XY plane:

```    public void DrawWireCirlce(Vector3 center, Vector3 direction, float radius) {
      PushMatrix();
      matrix = Matrix4x4.TRS(center, Quaternion.LookRotation(direction), Vector3.one) *
               matrix * 
               Matrix4x4.TRS(Vector3.zero, Quaternion.identity, new Vector3(1, 1, 0));
      DrawWireSphere(Vector3.zero, radius);
      PopMatrix();
    }```

File: Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs [hunk 4]
```diff
@@ -524,13 +524,60 @@ public class RuntimeGizmoDrawer {
...
+    public void DrawWireCapsule(Vector3 start, Vector3 end, float radius) {
+      Vector3 up = (end - start).normalized * radius;
+      Vector3 forward = Vector3.Slerp(up, -up, 0.5F);
+      Vector3 right = Vector3.Cross(up, forward).normalized * radius;
+
+      float height = (start - end).magnitude;
+      float distToCap = Mathf.Max(0, (height * 0.5F) - radius);
+      Vector3 middle = (end + start) * 0.5F;
+
+      start = middle + ((start - middle).normalized * distToCap);
```
Amarcolina: Might be controversial, but I think that the 'start' and 'end' params should be the actual points that define the capsule, and not the 'top' and 'bottom' concepts.  Maybe there could be a method overload that allows you to choose the method?

Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: Sure thing!

Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: So I actually agree with you on this but when I render using the function as-is the capsules look correct by that interpretation you're using...

Assets/LeapMotion/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: le done

PR #609, opened 2017-03-15
Feature textures and assets

Amarcolina:
Pretty big PR with some big features:
 - Rewrote PackUtil, turning it into AtlasBuilder!  Now supports GPU acceleration at every step for faster atlas building, as well as atlasing secondary textures.
 - All renderers now support multiple texture features.
 - Now supports changing the texture format for generated atlas to allow compression.
 - New asset import method supports much larger textures/meshes without holding them in the scene.  This means faster scene loading and smaller file sizes.
 - Textures are not atlased automatically, since that is actually too expensive when using a bunch of large textures!  Instead the user needs to specifically trigger a refresh and wait for the process to finish.
 - Various small bugs fixed.

I did *not* commit the assets to the repo, so when you visit the test scene they should auto-generate in a folder next to the scene.  The textures will require an explicit refresh action that you can perform on the gui component.  

To test the handling of large textures, you can download the albedo / glossiness / normals from [here](http://opengameart.org/content/4096-scifi-hex-tiles-pbr-texture) and assign them to _MainTex, _MetallicGlossMap, and _BumpMap.  Make sure to use the physically based shader!


nickjbenson:
Just gave it a quick once-over from the editor functionality, all of that works the way I expect. Only thing I noticed was that I couldn't tell which Texture elements corresponded to which Texture features, so I had to guess they would be index-aligned. Would be nice if the texture PropertyName could be in the Texture slot in the element renderer so that's not guesswork.

Will review the code over the next hour at the airport, and merge or have specific code comments later today.

nickjbenson:
OK! Looks good

PR #613, opened 2017-03-17
Interaction Module (IE v3)

nickjbenson:
Hot on the heels of John's IE v2.9 PR, which removed the need for external DLLs or an ldat in the Interaction Engine by moving soft contact into Unity C# and providing the C# heuristic grab classifier...

This PR is a big refactor of the Interaction Engine to provide an easier interface to the end-user for building more complex interactions. Most of the function logic is copy-paste (or rewrite-verbatim), with some variables and methods renamed for clarity and style consistency. However, certain fundamental paradigms have changed:

- InteractionBrushHands are now InteractionHands, and are _not_ MonoBehaviours. They are persistent objects that are constructed and updated manually by the InteractionManager, and are given Hand objects per-FixedUpdate by the InteractionManager.
- 'Hover' is now a first-class feature of the Interaction Engine, joining Contact (Brush+Soft) and Grasping.
- InteractionBehaviour now consists almost entirely of logistics that provide the user with a robust callback API for getting information about interactions as they are happening. InteractionBehaviours provide public OnHoverBegin/Stay/End as well as OnGraspBegin/Hold/End and OnContactBegin/Stay/End callbacks, including per-object variants, and other useful hooks/accessors like OnGraspedMovement (for overriding the result of Kabsch'd poses) and isHovered/isPrimaryHovered.
- Controllers are still present, but are hidden in public accessors rather than being floated up into the inspector.

However, there are still a couple of things missing:
- Throw Controllers are not yet implemented.
- Suspension Controllers are not yet implemented. If hands stop tracking during, e.g., a grasp, the grasped object will hover in place and some null reference errors will print to the console. When the hand becomes tracked again, normal functionality resumes. So it's not _the worst_ behavior ever but it obviously needs fixing.
- Please help me fill out what is missing as a part of the review process for this PR. Monday 3/20 I will be updating tickets to match the current state of IEv3 and get to re-adding missing features.


Assets/LeapMotion/Scripts/DataStructures/RingBuffer.cs
```diff
@@ -1,68 +1,110 @@
...
+
+    public void SetLatest(T t) {
+      Set(Length - 1, t);
+    }
+
+    public RingBufferEnumerator GetEnumerator() {
+      return new RingBufferEnumerator(this);
+    }
+
+    public class RingBufferEnumerator : IEnumerator<T>, System.Collections.IEnumerator {
```
Amarcolina: If you make RingBufferEnumerator a struct you can use foreach without creating garbage

Assets/LeapMotion/Scripts/Utils/HandUtils.cs
```diff
@@ -71,7 +71,6 @@ public class HandEvent : UnityEvent<Hand> { }
     public static Hand Left {
       get {
         if (Provider == null) return null;
-        if (Provider.CurrentFrame == null) return null;
```
Amarcolina: Why were these null checks removed?

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+      //For each probe (fingertip)
+      for (int j = 0; j < classifier.probes.Length; j++) {
+        //Calculate how extended the finger is
+        float tempCurl = Vector3.Dot(_hand.Fingers[j].Direction.ToVector3(), (j != 0) ? _hand.Direction.ToVector3() : (_hand.IsLeft ? 1f : -1f) * _hand.Basis.xBasis.ToVector3());
+
+        //Determine if this probe is intersecting an object
+        Array.Clear(collidingCandidates, 0, 10);
+        Physics.OverlapSphereNonAlloc(_hand.Fingers[j].TipPosition.ToVector3(), j == 0 ? 0.015f : 0.01f, collidingCandidates);
+        bool collidingWithObject = false;
+        foreach(Collider col in collidingCandidates) {
```
Amarcolina: More efficient to use the return value of OverlapSphereNonAlloc to do a for loop, this also removes the need for the internal null check.

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+          if (col!=null && col.attachedRigidbody != null) {
+            collidingWithObject = (col.attachedRigidbody == classifier.body) ? true : collidingWithObject;
```
Amarcolina:
```
if (col!=null && col.attachedRigidbody != null && col.attachedRigidbody == classifier.body) {
  collidingWithObject = true;
  break;
}
```
This change allows you to break out of the loop early, since all you need to know is if at least one obj is colliding.

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+          }
+        }
+
+        //Nullify above findings if fingers are extended
+        bool tempIsInside = collidingWithObject && (tempCurl < 0.65f) && (tempCurl > -0.1f);
```
Amarcolina: If collidingWithObject is no longer used, why not just modify collidingWithObject itself?

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        //Probes go inside when they intersect, probes come out when they uncurl
+        if (!classifier.probes[j].isInside) {
+          classifier.probes[j].isInside = tempIsInside;
+          classifier.probes[j].curl = tempCurl + (j==0?0.15f:0f);
```
Amarcolina: Magic number is magic!  Should be moved to a constant to give clarity on what it means.

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      //If thumb and one other finger is "inside" the object, it's a grab!
+      classifier.isGrabbing = (classifier.probes[0].isInside && (classifier.probes[1].isInside ||
+                                                                 classifier.probes[2].isInside ||
+                                                                 classifier.probes[3].isInside ||
+                                                                 classifier.probes[4].isInside));
+      //If grabbing within 10 frames of releasing, discard grab.
+      //Suppresses spurious regrabs and makes throws work better.
+      if (classifier.coolDown <= 10) {
```
Amarcolina: I would definitely want this to be time based instead of frame based.  With our work on mobile we are going to be having different frame-rates across different applications, both for rendering and for physics.

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+        if (classifier.isGrabbing) {
+          classifier.isGrabbing = false;
+        }
+        classifier.coolDown++;
+      }
+    }
+
+    public void UpdateBehaviour(IInteractionBehaviour behaviour) {
+      GrabClassifier classifier;
+      if (!(_hand.IsLeft?leftGrabClassifiers: rightGrabClassifiers).TryGetValue(behaviour, out classifier)) {
```
Amarcolina: Would be nice to save the result of this ternary expression so it doesn't need to be duplicated.

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+        classifier = new GrabClassifier(behaviour);
```
Amarcolina: Would also be nice to pool these classifiers

Assets/LeapMotionModules/InteractionEngine/Scripts/ControllerImplementations/HeuristicGrabClassifier.cs
```diff
@@ -0,0 +1,142 @@
...
+
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+namespace Leap.Unity.Interaction {
+  public class HeuristicGrabClassifier {
+    Collider[] collidingCandidates = new Collider[10];
+    public InteractionManager _manager;
+    private Hand _hand;
```
Amarcolina: Be very careful when keeping references to hands around, they can change out from underneath you because we pool them on the provider side.  If you need a reference to the entire hand, you should create a new hand and copy the data over.

Assets/LeapMotionModules/InteractionEngine/Scripts/Editor/InteractionManagerEditor.cs
```diff
@@ -29,6 +29,9 @@ public class InteractionManagerEditor : CustomEditorBase<InteractionManager> {
                                 "_interactionNoClipLayer",
                                 "_brushLayer");
 
+      specifyConditionalDrawing(() => serializedObject.FindProperty("_graspingEnabled").boolValue,
```
Amarcolina: You can actually just write
`specifyConditionalDrawing("_graspingEnabled", "_twoHandedGrasping")`

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionBehaviours/InteractionBrushBone.cs
```diff
@@ -17,50 +19,23 @@ public class InteractionBrushBone : MonoBehaviour {
...
-    public bool updateTriggering() {
-      if (_dislocatedCounter < DISLOCATED_BRUSH_COOLDOWN) {
-        if (++_dislocatedCounter == DISLOCATED_BRUSH_COOLDOWN) {
-          capsuleCollider.isTrigger = false;
-          return false;
-        }
-        return true;
-      }
-      return false;
+    public void DisableColliderTemporarily(float seconds) {
```
Amarcolina: For safety you should store the reference to the currently running coroutine so you can stop it if this method gets called while it is still running.  If you don't stop it, you might get unexpected results.

Assets/LeapMotionModules/InteractionEngine/Scripts/Utility/BrushHandPostProcess.cs
```diff
@@ -0,0 +1,18 @@
+﻿using UnityEngine;
+using Leap.Unity.Interaction;
+using Leap;
+
+public class BrushHandPostProcess : MonoBehaviour {
```
Amarcolina: What is this?  I can't find a reference to it in any scene, nor in any script.

Assets/LeapMotionModules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -63,6 +62,10 @@ public partial class InteractionManager : MonoBehaviour {
     [SerializeField]
```
Amarcolina: This file still has ldat path in it, which is no longer needed.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+    [Header("Interactions")]
+    public bool enableHovering = true;
+    public bool enableContact  = true;
+    public bool enableGrasping = true;
+
+    // TODO: Hide these settings behind a dropdown, they probably don't really ever need to be changed
+    [Header("Interaction Settings")]
+    [DisableIf("enableHovering", isEqualTo: false)]
+    public float hoverActivationRadius = 0.5F;
+    [DisableIf("contactOrGraspingEnabled", isEqualTo: false)]
```
Amarcolina: You can just use 2 DisableIf attributes instead.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Leap.Unity.UI.Interaction {
+
+  public partial class InteractionManager : MonoBehaviour, IRuntimeGizmoComponent {
+
+    [Header("Interactions")]
+    public bool enableHovering = true;
```
Amarcolina: Eventually going to need to hide this behind properties so that you can take proper action if they are messed with at runtime.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+    /// that has had its Awake() method called in the current scene. If an
+    /// InteractionBehaviourBase does not have a non-null interactionManager by the
+    /// time it has Start() called, it will default to using the InteractionManager
+    /// referenced here.
+    /// 
+    /// If you have multiple InteractionManagers in your scene, you should be sure to
+    /// assign InteractionBehaviours' managers appropriately. If you instantiate an
+    /// InteractionBehaviour at runtime, you should assign its InteractionManager
+    /// right after you instantiate it. </remarks>
+    public static InteractionManager singleton { get; set; }
```
Amarcolina: Would recommend adding in a Find in the get case.  If manager A is created, then manager B is created, then manager A is deleted, singleton will return null even though there is an active manager in the scene.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+
+    void Awake() {
+      if (InteractionManager.singleton == null) singleton = this;
```
Amarcolina: `singleton = singleton ?? this; `

This is potentially the most important change in the entire world.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+
+      Provider = Hands.Provider;
+
+      _interactionHands[0] = new InteractionHand(this, () => { return Hands.FixedLeft; }, Chirality.Left, WorldHoverActivationRadius, WorldTouchActivationRadius);
```
Amarcolina: By passing in these radii at construction time they will miss changes made after init

Amarcolina: Ah I see they are regularly copied over in Update.  Might it be better for the hands to have a reference to the manager so they can just ask questions / access data whenever they need it?

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+
+      if (_autoGenerateLayers) {
+        AutoGenerateLayers();
+        AutoSetupCollisionLayers();
+      }
+    }
+
+    void Start() {
+      if (Provider == null) {
+        Debug.LogError("[InteractionManager] No LeapServiceProvider found.");
```
Amarcolina: would be good to disable the manager in this case so you don't get error spew.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+using UnityEngine;
+
+namespace Leap.Unity.UI.Interaction {
+
+  public abstract class InteractionBehaviourBase : MonoBehaviour {
+
+    public InteractionManager interactionManager;
+
+    protected Rigidbody _body;
+    public Rigidbody Rigidbody { get { return _body; } }
```
Amarcolina: would be nice to name this rigidbody instead, to follow property naming conventions.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+        interactionManager.RegisterInteractionBehaviour(this);
+      }
+    }
+
+    protected virtual void Start() {
+      if (interactionManager == null) {
+        interactionManager = InteractionManager.singleton;
+
+        if (interactionManager == null) {
+          Debug.LogError("Interaction Behaviours require an Interaction Manager. Please ensure you have an InteractionManager in your scene.");
```
Amarcolina: Would be good to disable behavior in this case to prevent spew.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+    protected virtual void Awake() {
+      _body = GetComponent<Rigidbody>();
+      _body.maxAngularVelocity = 100F;
+
+      if (interactionManager != null) {
+        interactionManager.RegisterInteractionBehaviour(this);
+      }
+    }
+
+    protected virtual void Start() {
```
Amarcolina: This should probably be done in OnEnable instead of start so that you can register/unregister by enabling/disabling behaviors

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+using UnityEngine;
+
+namespace Leap.Unity.UI.Interaction {
+
+  public class InteractionHand {
+
+    public InteractionManager interactionManager;
+
+    private Func<Hand> _handAccessor;
+    private Hand _hand;
```
Amarcolina: You need to be careful when holding onto references to hands.  They can change out from underneath you due to the pooling done by the provider.  If you need the hand to stick around past fixedUpdate, make a copy of it.  If you don't, I would recommend passing the hand around via method arguments to make it very clear that this hand is not a long-term-field.

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+      };
+
+      foreach (var interactionObj in hoverCandidates) {
+        results = CheckHoverForElement(hand, interactionObj, results);
+      }
+
+      return results;
+    }
+
+    private HoverCheckResults CheckHoverForElement(Hand hand, InteractionBehaviourBase hoverable, HoverCheckResults curResults) {
```
Amarcolina: Looks like this method is returning the argument, but modified.  A better way would be to pass in curResults as a ref param, which makes it more efficient, but also more obvious what it is doing.

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+        Leap.Vector distalFingerBoneTip = _hand.Fingers[i].bones[3].NextJoint;
+        Leap.Vector medialFingerBoneDirection = _hand.Fingers[i].bones[2].Direction;
+        Leap.Vector distalDirection = _hand.Basis.zBasis;
+        var fingerUsage = (((medialFingerBoneDirection.x * distalDirection.x)
+                         + (medialFingerBoneDirection.y * distalDirection.y)
+                         + (medialFingerBoneDirection.z * distalDirection.z))
+                         * _intentionFingerWeights[i])
+                         .Map(-0.3F, 0.9F, 0, 1);
+        usageSum += fingerUsage;
+        averagePosition = new Leap.Vector(averagePosition.x + distalFingerBoneTip.x * fingerUsage,
```
Amarcolina: Any reason you are doing the component math manually instead of just doing
`averagePosition + distalFingerBoneTip * fingerUsage`
?

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+      // Distal Intention Ray: Add punch vector for when finger usage is low (closed fist).
+      // Vector3 distalAxis = _hand.DistalAxis();
+      float punchWeight = usageSum.Map(0F, 1F, 1F, 0F);
+      averagePosition = new Leap.Vector(averagePosition.x + _hand.Fingers[2].bones[1].PrevJoint.x * punchWeight,
+                                         averagePosition.y + _hand.Fingers[2].bones[1].PrevJoint.y * punchWeight,
+                                         averagePosition.z + _hand.Fingers[2].bones[1].PrevJoint.z * punchWeight);
+      //averageDirection = new Leap.Vector(averageDirection.x + distalAxis.x * punchWeight,
+      //                                   averageDirection.y + distalAxis.y * punchWeight,
+      //                                   averageDirection.z + distalAxis.z * punchWeight);
+      usageSum += punchWeight;
```
Amarcolina:
```
newSum = usageSum + punchWeight
punchWeight = 1 - usageSum        (map (0-1) to (1-0))
newSum = usageSum + (1 - usageSum)
newSum = 1???
```

Assets/LeapMotionModules/UI/Interaction/ActivityManager.cs
```diff
@@ -0,0 +1,78 @@
...
+      get { return _activeBehaviours; }
+    }
+
+    public ActivityManager(InteractionManager manager, float activationRadius) {
+      this.manager = manager;
+      this.activationRadius = activationRadius;
+    }
+
+    public void FixedUpdateHand(Hand hand) {
+      int count = GetSphereColliderResults(hand, _colliderResultsBuffer, out _colliderResultsBuffer);
```
Amarcolina: Instead of passing in the array twice, once as an out modifier, you can just use the ref modifier for the same effect

Assets/LeapMotionModules/UI/Interaction/ActivityManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    private int GetSphereColliderResults(Hand hand, Collider[] resultsBuffer_in, out Collider[] resultsBuffer_out) {
+      resultsBuffer_out = resultsBuffer_in;
+      if (hand == null) return 0;
+
+      int overlapCount = 0;
+      while (true) {
+        overlapCount = Physics.OverlapSphereNonAlloc(hand.PalmPosition.ToVector3(),
+                                                         activationRadius * 100,
+                                                         resultsBuffer_in,
+                                                         ~0,
```
Amarcolina: Isn't this able to constrain the search based on the layer specified in the manager?

Assets/LeapMotionModules/UI/Interaction/ActivityManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private int GetSphereColliderResults(Hand hand, Collider[] resultsBuffer_in, out Collider[] resultsBuffer_out) {
+      resultsBuffer_out = resultsBuffer_in;
+      if (hand == null) return 0;
+
+      int overlapCount = 0;
+      while (true) {
+        overlapCount = Physics.OverlapSphereNonAlloc(hand.PalmPosition.ToVector3(),
+                                                         activationRadius * 100,
```
Amarcolina: Why is there a *100 here?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+    #region Hovering
+
+    // Logistics for providing per-object (instead of per-hand) Hover callbacks.
+    private Hand  _closestHoveringHand = null;
+    private float _closestHoveringHandDistance = float.PositiveInfinity;
+    private Hand  _closestJustStoppedHoveringHand = null; // Provided for OnObjectHoverEnd.
+    private int   _hoveringHandsCountLastFrame = 0;
+    private int   _hoveringHandsCount = 0;
+
+    private Hand _publicClosestHoveringHand = null;
```
Amarcolina: Why the duplicate variable for the public getter?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+          _closestHoveringHand = hand;
+          _closestHoveringHandDistance = handDistance;
+        }
+      }
+    }
+
+    public override void HoverEnd(Hand hand) {
+      _hoveringHandsCount--;
+      if (_hoveringHandsCount == 0) {
+        _closestJustStoppedHoveringHand = hand;
```
Amarcolina: Isn't this just the last hand to stop hovering?  What makes it the closest?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+      FixedUpdatePrimaryHovering();
+      FixedUpdateContact();
+      FixedUpdateGrasping();
+
+      FixedUpdateCollisionMode();
+    }
+
+    // TODO: Currently this gets the distance from the point to this transform, but this will
+    // need to incorporate distance to the rigidbody.
+    private float GetInteractionDistanceToPoint(Vector3 point) {
```
Amarcolina: Would be nice if this was protected virtual to allow you to specify a custom metric as well.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+
+    private void InitGrasping() {
+      _moveObjectWhenGrasped__WasEnabledLastFrame = moveObjectWhenGrasped;
+
+      _kinematicHoldingMovement = new KinematicGraspedMovement();
+      _nonKinematicHoldingMovement = new NonKinematicGraspedMovement();
+    }
+
+    private int _graspCountLastFrame = 0;
+    private List<InteractionHand> _graspingIntHandsCache = new List<InteractionHand>(4);
```
Amarcolina: Is this really a cache?  It just seems to be a list of currently grasping hands

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+      _moveObjectWhenGrasped__WasEnabledLastFrame = moveObjectWhenGrasped;
+
+      _graspingHandsCache.Clear();
+      foreach (var hand in _graspingIntHandsCache.Query().Select(intHand => intHand.GetLeapHand())) {
+        _graspingHandsCache.Add(hand);
+      }
+
+      if (_graspCount > 0) {
+        if (_graspCountLastFrame == 0) {
+          OnObjectGraspBegin();
```
Amarcolina: Any reason this isn't dispatched within the OnGraspBegin callback?  Will this cause a frame delay?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+        }
+        else {
+          OnObjectGraspHold();
+        }
+
+        if (_graspCount > 1) {
+          if (_graspCountLastFrame <= 1) {
+            OnMultiGraspBegin(_graspingHandsCache);
```
Amarcolina: Same for this, it seems like it would be better to dispatch the event from within OnGraspBegin

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+          _closestPrimaryHoveringHand = hand;
+          _closestPrimaryHoveringHandDistance = handDistance;
+        }
+      }
+    }
+
+    public override void PrimaryHoverEnd(Hand hand) {
+      _primaryHoveringHandsCount--;
+      if (_primaryHoveringHandsCount == 0) {
+        _closestJustStoppedPrimaryHoveringHand = hand;
```
Amarcolina: Seems like you could just dispatch OnObjectPrimaryHoverEnd right here?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// <summary> Return the distance the interaction object is from the given world position. </summary>
+    public abstract float GetDistance(Vector3 worldPosition);
+
+    #region Hovering
+
+    /// <summary> Called per-hand when that hand produces a non-zero hover score for this object. </summary>
+    public abstract void HoverBegin(Hand hand);
+
+    /// <summary> Called per-hand when that hand maintains a non-zero hover score for each frame beyond the first. </summary>
+    public abstract void HoverStay(Hand hand);
```
Amarcolina: For most of these callbacks, I would like to discuss using a HoverStay(List<Hand> hands) for all hands instead of HoverStay(Hand) for each hand one at a time.  I think this would be positive for a number of reasons:

 - If I want to do an action for every frame that an object is hovered, I just put my logic in HoverStay, and it works!
 - If I want to do logic that is based around all currently hovering hands, I also can just use HoverStay, and it's easy
 - I think that it would reduce the number of fields in this class, since a fair number of them are simply temporary state fields used during Stay callbacks to do a calculation like closest hand.

Also, I think it would be useful for the Stay callback of a hand/event to be called on the *same* frame that the Begin callback happened.  If I want to do logic for every frame that an object is hovered, I want to be able to put that logic in HoverStay without getting a 1 frame delay.

Assets/LeapMotion/Scripts/Utils/HandUtils.cs
nickjbenson: John assured me that Provider never has a null CurrentFrame, so I'm committing to that :P is it not true?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+          OnObjectHoverBegin(_closestHoveringHand);
+        }
+        else {
+          _publicClosestHoveringHand = _closestHoveringHand;
+          OnObjectHoverStay(_closestHoveringHand);
+        }
+      }
+      else if (_hoveringHandsCountLastFrame > 0) {
+        _closestHoveringHand = null;
+        OnObjectHoverEnd(_closestJustStoppedHoveringHand);
```
Amarcolina: Another general lifecycle discussion.  I would argue that Stay should be dispatched after all Begin/End callbacks are dispatched.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+    public override void ContactEnd(Hand hand) {
+      _contactingHandsCount -= 1;
+      if (_contactingHandsCount == 0) {
+        _closestJustStoppedContactingHand = hand;
+      }
+
+      OnContactEnd(hand);
+    }
+
+    public void ObjectContactBegin(Hand hand) {
```
Amarcolina: Should this be protected and virtual?

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
Amarcolina: I would argue that instead of using the Hands class as the accessor, that the manager should be specifically hooked up to a provider to allow cases for multiple providers to be handled unambiguously.

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+            Vector3 boneCenter = bone.Center.ToVector3();
+
+            // Generate and fill softContacts with SoftContacts that are intersecting a sphere
+            // at boneCenter, with radius softContactBoneRadius.
+            bool sphereIntersecting;
+            sphereIntersecting = PhysicsUtility.GenerateSphereContacts(boneCenter, _softContactBoneRadius,
+                                                                       (boneCenter - _previousBoneCenters[boneArrayIndex])
+                                                                         / Time.fixedDeltaTime,
+                                                                       1 << interactionManager.InteractionLayer,
+                                                                       ref interactionManager._softContacts,
```
Amarcolina: Instead of giving the manager a public interface so hands can talk back to it, how about the hands can return some results as a result of being updated, that the manager can then act on?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    public override void GraspHold(Hand hand) {
+      if (moveObjectWhenGrasped) {
+        Vector3 origPosition = Rigidbody.position; Quaternion origRotation = Rigidbody.rotation;
+        Vector3 newPosition; Quaternion newRotation;
+        GraspedPoseController.GetGraspedPosition(out newPosition, out newRotation);
+
+        IGraspedMovementController holdingMovementController = Rigidbody.isKinematic ?
+                                                                 (IGraspedMovementController)_kinematicHoldingMovement
```
Amarcolina: These casts are not needed

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+       && (_body.constraints & RigidbodyConstraints.FreezePositionY) > 0
+       && (_body.constraints & RigidbodyConstraints.FreezePositionZ) > 0) {
+        _isPositionLocked = true;
+        return;
+      }
+      else {
+        Joint[] joints = _body.GetComponents<Joint>();
+        foreach (var joint in joints) {
+          FixedJoint fixedJoint = joint as FixedJoint;
+          if (fixedJoint != null) {
```
Amarcolina: can just use
`joint is FixedJoint`
since you don't actually need the reference to the fixedJoint object

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+            _isPositionLocked = true; return;
```
Amarcolina: Would argue that setting it to false at the top and then setting it to true after would be slightly better.  That way if someone comes in here and makes a change and forgets to add a return, it doesn't suddenly break, it just gets slightly slower.

Also I think it would be good to put these returns on their own line.

Assets/LeapMotionModules/UI/Interaction/KabschGraspedPose.cs
```diff
@@ -0,0 +1,193 @@
...
+
+  public class KabschGraspedPose : IGraspedPoseController {
+    public const int NUM_FINGERS = 5;
+    public const int NUM_BONES = 4;
+
+    public enum SolveMethod {
+      SixDegreeSolve,
+      PivotAroundOrigin
+    }
+    public SolveMethod _solveMethod;
```
Amarcolina: Currently this is overwritten no matter what it is initialized to, so it probably should not be a public field.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+        _collisionMode = desiredCollisionMode;
+        FixedUpdateLayer();
+      }
+
+      Assert.IsTrue((_collisionMode == CollisionMode.Grasped) == isGrasped);
+    }
+
+    protected void FixedUpdateLayer() {
+      int layer;
+      if (_collisionMode != CollisionMode.Normal) {
```
Amarcolina: since this is an enum, a switch statement with a default case that throws an exception would be great to enforce correct behavior (or noisy failing!) if a new mode is added

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,692 @@
...
+      _moveObjectWhenGrasped__WasEnabledLastFrame = moveObjectWhenGrasped;
+
+      _kinematicHoldingMovement = new KinematicGraspedMovement();
+      _nonKinematicHoldingMovement = new NonKinematicGraspedMovement();
+    }
+
+    private int _graspCountLastFrame = 0;
+    private List<InteractionHand> _graspingIntHandsCache = new List<InteractionHand>(4);
+    private List<Hand> _graspingHandsCache = new List<Hand>(4);
+    private List<Hand> _lastFrameGraspingHandsCache = new List<Hand>(4); // for OnMultiGraspEnd
```
Amarcolina: Also you cannot keep references to hands around for the next frame, since they might have changed out from underneath you due to the provider performing object pooling.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+    [SerializeField]
+    [HideInInspector]
+    #pragma warning disable 0414
+    // Variable used only for inspector rendering.
+    private bool _interactionManagerIsNull = true;
+    #pragma warning restore 0414
+
+    [Header("Interaction Settings")]
+    [DisableIf("_interactionManagerIsNull", isEqualTo: true)]
+    public bool ignoreHover = false; // TODO:
```
Amarcolina: For me, checking a box when I want to disable a feature is very unintuitive.  Maybe enableHover instead?

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    [SerializeField]
+    [HideInInspector]
+    #pragma warning disable 0414
+    // Variable used only for inspector rendering.
+    private bool _interactionManagerIsNull = true;
+    #pragma warning restore 0414
+
+    [Header("Interaction Settings")]
+    [DisableIf("_interactionManagerIsNull", isEqualTo: true)]
```
Amarcolina: You still want to be able to edit these fields even if a manager does not exist.  You might want to edit an interaction behaviour that is on a prefab, or before you create the manager.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+    [SerializeField]
+    protected bool _autoGenerateLayers = true;
+
+    [Tooltip("Layer to use for auto-generation.  The generated interaction layers will have the same collision settings as this layer.")]
+    [SerializeField]
+    protected SingleLayer _templateLayer = 0;
+
+    [Tooltip("The layer containing interaction objects.")]
+    [SerializeField]
+    protected SingleLayer _interactionLayer = 0;
```
Amarcolina: These layers should be hidden if autoGenerateLaters is enabled, since their serialized value doesn't matter and will be overwritten.  Similarly, the template layer should be hidden if _autoGenerateLayers is false, since it isn't used and isn't important.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+
+    void LateUpdate() {
+      OnGraphicalUpdate();
+    }
+
+    #region Object Registration
+
+
+    public void RegisterInteractionBehaviour(InteractionBehaviourBase interactionObj) {
+      _interactionBehaviours.Add(interactionObj);
```
Amarcolina: There should be some verification that an object is not registered multiple times.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+      _rigidbodyRegistry[interactionObj.Rigidbody] = interactionObj;
```
Amarcolina: Also looks like there is no way currently for an object to unregister itself, nor can it know if it is currently registered.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// <summary>
+    /// A scale that can be used to appropriately transform distances that otherwise expect
+    /// one Unity unit to correspond to one meter.
+    /// </summary>
+    public float SimulationScale { get { return _providerScale; } }
+
+    private InteractionHand[] _interactionHands = new InteractionHand[2];
+    private HashSet<InteractionBehaviourBase> _interactionBehaviours = new HashSet<InteractionBehaviourBase>();
+    private Dictionary<Rigidbody, InteractionBehaviourBase> _rigidbodyRegistry = new Dictionary<Rigidbody, InteractionBehaviourBase>();
```
Amarcolina: For cases where you want a variable to be publicly gettable but not settable, you can use this syntax

`public Type fieldName { get; private set; }`

This won't work for serialized fields unfortunately because unity cannot serialize properties.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+    [SerializeField]
+    protected SingleLayer _graspedObjectLayer = 0;
+
+    [Tooltip("The layer containing the colliders for the bones of the hand.")]
+    [SerializeField]
+    protected SingleLayer _contactBoneLayer = 0;
+
+    [SerializeField]
+    #pragma warning disable 0414
+    private bool _showDebugOptions = false;
```
Amarcolina: For cases where you want to hide options inside of a toggle, you can wrap them in a struct.  Unity will keep them organized underneath a nice foldout.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
Amarcolina: Ah I see, you have addressed that above in the Provider property.  So it looks like what really should be passed in here is something to get the left hand from the Provider property.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+      _interactionHands[0] = new InteractionHand(this, () => { return Hands.FixedLeft; }, Chirality.Left, WorldHoverActivationRadius, WorldTouchActivationRadius);
+      _interactionHands[1] = new InteractionHand(this, () => { return Hands.FixedRight; }, Chirality.Right, WorldHoverActivationRadius, WorldTouchActivationRadius);
+
+      if (_autoGenerateLayers) {
+        AutoGenerateLayers();
+        AutoSetupCollisionLayers();
+      }
+    }
+
+    void Start() {
```
Amarcolina: Eventually going to need correct enable/disable handling for when a manager is enabled/disabled at runtime.  This will include things like letting all currently grasped objects know they have been released.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+      if (!_interactionBehaviours.Contains(interactionObj)) {
+        Debug.LogError("ReleaseObjectFromGrasp was called, but the interaction object " + interactionObj.transform.name + " is not registered"
+          + " with this InteractionManager.");
+        return false;
+      }
+
+      foreach (var hand in _interactionHands) {
+        if (hand.IsGrasping(interactionObj)) {
+          hand.ReleaseGrasp();
+          return true;
```
Amarcolina: For cases where more than one hand is holding onto an object, you don't want to bail out at the first hand found.

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+    public bool allowMultiGrasp = false;
+
+    /// <summary>
+    /// Called by the InteractionManager every FixedUpdate, after
+    /// InteractionHands have FixedUpdated.
+    /// 
+    /// InteractionBehaviour uses this to fire per-object interaction events.
+    /// </summary>
+    public abstract void FixedUpdateObject();
```
Amarcolina: I would say that calling this something like OnPostSolve or something could help make it clearer that this is called after the manager does it's thing.  Otherwise when looking at code I see a method called FixedUpdateObject and wondering how it might be different from the normal FixedUpdate callback

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    protected virtual void Awake() {
+      _body = GetComponent<Rigidbody>();
+      _body.maxAngularVelocity = 100F;
```
Amarcolina: Probably should be a constant defined somewhere.  Also we should discuss if we want this value to be reset to default if the interaction behaviour is disabled/deleted

Assets/LeapMotionModules/UI/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,130 @@
...
+      _body = GetComponent<Rigidbody>();
+    }
+
+    /// <summary> Return the distance the interaction object is from the given world position. </summary>
+    public abstract float GetDistance(Vector3 worldPosition);
+
+    #region Hovering
+
+    /// <summary> Called per-hand when that hand produces a non-zero hover score for this object. </summary>
+    public abstract void HoverBegin(Hand hand);
```
Amarcolina: Does this mean that objects that are currently being grasped can get hover callbacks?  Can they get hover callbacks from the hand that is currently grasping them?  If so, is there a guarantee that the hand currently grasping the object has it's primary hover set to be this object as well?

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+    }
+
+    public Hand GetLeapHand() {
+      return _hand;
+    }
+
+    public void FixedUpdateHand(bool doHovering, bool doContact, bool doGrasping) {
+      _hand = _handAccessor();
+
+      if (doHovering) FixedUpdateHovering();
```
Amarcolina: I would argue it makes sense to add more strict ordering to the ordering.  Currently it looks like callbacks for objects are dispatched in 'hand order'.  So you receive all possible callbacks from one hand, then all possible callbacks from the next hand.  This can result in strange cases where my grasped state is not the same between my first and second hover stay callback during the same frame!

I would propose ordering by concept instead.  So something like, first all hover callbacks, then all touch callbacks, then all grasp callbacks.

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+      }
+    }
+
+    public void Grasp(InteractionBehaviourBase interactionObj) {
+      interactionObj.GraspBegin(_hand);
+      _graspedObject = interactionObj;
+    }
+
+    public void ReleaseGrasp() {
+      if (_graspedObject == null) return; // Nothing to release.
```
Amarcolina: I think that it should be an error to tell a hand to release if it is not currently releasing.  Either that, or make this a Try method so that it is obvious that the method can have no effect.

Assets/LeapMotionModules/UI/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,293 @@
...
+      if (hand.Handedness() == Chirality.Left) {
+        return _interactionHands[0];
+      }
+      else {
+        return _interactionHands[1];
+      }
+    }
+
+    /// <summary> Returns true if the object was released from a grasped hand, or false if the object was not held in the first place. </summary>
+    public bool ReleaseObjectFromGrasp(InteractionBehaviourBase interactionObj) {
```
Amarcolina: Would argue for renaming this TryReleaseObjectFromGrasp to make it more obvious that the method can fail.

Assets/LeapMotionModules/UI/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,780 @@
...
+    private void FixedUpdateGrasping() {
+      _grabClassifier.FixedUpdate();
+
+      if (_graspedObject != null) {
+        _graspedObject.GraspHold(_hand);
+      }
+    }
+
+    public void Grasp(InteractionBehaviourBase interactionObj) {
+      interactionObj.GraspBegin(_hand);
```
Amarcolina: Should do verification here to make sure that the hand isn't already grasping an object.

PR #617, opened 2017-03-21
Fix Multiple Post Processes

jselstad:
There was a subtle bug in the structure of post processes that caused them to apply to all hands of the same model type; this PR ensures post processes apply to hands on a per-model-group basis.

An Example Hand Post Process function (for quick testing) looks like this:
```
 public void setHeldHand(Hand inHand) {
    inHand.SetTransform(transform.position, transform.rotation);
  }
```


Amarcolina:
Looks good to me!

PR #623, opened 2017-03-24
Add tween to core

Amarcolina:
This adds our age-old tween library to core, since it has proven to be quite useful in a bunch of places.  This includes adding a new lightweight data structure, the Pool class.  It also adds a small utility function, Util.DoubleCapacity, which doubles the length of an array.

### DIFF
--- a/Assets/LeapMotion/Scripts/DataStructures/Pool.cs
+++ b/Assets/LeapMotion/Scripts/DataStructures/Pool.cs
@@ -6,7 +6,7 @@ namespace Leap.Unity {
   /// <summary>
   /// Implement this interface to recieve a callback whenever your object is
   /// spawned from a pool.  You do not recieve a callback during recycle because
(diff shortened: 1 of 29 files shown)


Assets/LeapMotion/Scripts/DataStructures/Pool.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Leap.Unity {
+
+  public interface IPoolable {
```
nickjbenson: It looks like the way to implement an IPoolable<T> is pretty simple: Implement OnSpawn(), use Pool<T>.Spawn() to get a T, and then when a given T is ready to be recycled, just call Pool<T>.Recycle(theT). There aren't any other gotchas to think about, right? If so, (1) that's super sweet, (2) could you put a doc comment here for that so that an implementer is told that explicitly?

PR #626, opened 2017-03-24
Discussion PR: Fancier Pinch Position

nickjbenson:
Pull the PR branch and check out the InteractionEngine example scene. (This has nothing to do with the interaction engine, I just needed a scene.) Press play and look at the gizmo indicating your pinch position.

The relevant method is in PinchGizmoTest.cs, in InteractionEngine/Examples/Scenes.

The core idea is, I think it'd be good to have a heuristic for "pinch position" that is more accurate than any index-thumb average when those fingers are far apart. Optionally, the method I am using to calculate the pinch position can be set to heuristic-only, providing a rigid approximation of the pinch position on a hand. This could be useful in a context where drift due to index-thumb finger movement would be highly undesired.

(1) Is it too expensive?
(2) Is it unnecessary?
If it's not too crazy, I'd like to replace GetPinchPosition() in HandUtil.cs with this fancier calculation. I think it's valuable to have a better estimate of where the hand _will_ pinch, even if it isn't pinching. (And would like to use it in Lines; so if it's too fancy I'll just keep it bespoke for that use-case.)

(Secondarily in this PR you'll see I've renamed the HandUtil.cs "Index()," "Pinky()," etc. methods to "GetIndex(), GetPinky()", because functions in general should be verb phrases, to distinguish them from properties, which "Index" and "Pinky" reeeaally feel like they should be as you type them out)

(If approved, I'll submit additional changes that incorporate replaces the appropriate method in HandUtil.cs and clean up the dirt from the discussion-related items in the commit)

[UPDATE]
OK! Everything is pushed to make this an actual PR. Please review as normal.


nickjbenson:
Ready to merge now! (Upon Alex's approval)

PR #627, opened 2017-03-24
Fix Flicking, Fix Scaling, Make Grab Less Sticky

jselstad:
-Add a Heuristic that allows Flicking to work more often
-Reduce the stickiness value slightly
-Ensure that a part of the object has to be close to the hand to trigger a grab (to help prevent spurious unreasonable looking grabs from taking place)
-Allow each of the scale-aware grab heuristics to derive scaling from the interaction manager

- [x] - Ensure you're still able to grab and hold reliably, but that it's not _too_ sticky


nickjbenson:
Looks good to me. My time with it revealed no unwanted holding when I wanted to release; there were a couple of times where it released when I wasn't intending, but only a couple, over ~50 grasp-release cycles. In general, it felt reliable. Code changes look good.

Amarcolina:
Everything looks good to me as well, just waiting on check off from QA and I think we are good!

mmenicutch:
Everything is looking good on my end as well. I say go ahead and merge it.

PR #630, opened 2017-03-27
Fixed single layer editor tooltip

Amarcolina:
Unity for some reason doesn't pass tooltips directly into property drawers -_- use reflection to grab onto it manually.


nickjbenson:
It woooorks

PR #631, opened 2017-03-27
Added OnChange attribute

Amarcolina:
Tag a property with OnChange to get an edit time callback whenever it is modified by the user.  The callback can be in the form of a method invocation (with or without the new value passed as an argument), or in the form of invoking a property setter with the new value.


nickjbenson:
Looks good!

PR #632, opened 2017-03-28
Optimize Grab Classifier

jselstad:
Was querying the scene once per object rather than once per hand; this optimization was present in earlier grab classifiers but got removed due to some devious gittery.

- [ ] - Ensure everything still works


nickjbenson:
Looks good to me, and example scene still works the way I expect

nickjbenson:
Scratch that, just realized you got rid of GrapProbe.position in the DLL.

PR #633, opened 2017-03-28
Interaction Module (IEv3)

nickjbenson:
OK, let's try this again.

Recap of the good stuff:
- API redesign: Lots of callbacks! Falling-off-a-log easy to get hover information, contact callbacks, grasp callbacks and grasped behaviour overriding
- InteractionHands are persistent and do a lot of player/number-of-hands-agnostic work; this makes various things nice and allows us to support many players if we want.

Things to be changed later:
- Have InteractionHands return result objects instead of giving Manager a public interface (this is not a bad idea but again a good-sized change)
- Manager allowing multiple providers; at the top level abstract I am OK with the current state that the Manager assumes one player. It was important that lower levels of abstraction didn't care how many players/hands there were, and that should be the case currently, so it is easy to change later.

Things being worked on now / to be done before merge:
- Testing to make sure the big callback refactor didn't break everything


Assets/LeapMotionModules/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    public override void GraspHold(Hand hand) {
+      if (moveObjectWhenGrasped) {
+        Vector3 origPosition = rigidbody.position; Quaternion origRotation = rigidbody.rotation;
+        Vector3 newPosition; Quaternion newRotation;
+        GraspedPoseController.GetGraspedPosition(out newPosition, out newRotation);
+
+        IGraspedMovementController holdingMovementController = rigidbody.isKinematic ?
+                                                                 (IGraspedMovementController)_kinematicHoldingMovement
```
Amarcolina: Wow, TIL that you can't use a ternary expression in this situation without casting.  The reason I thought you could remove it was that Visual Studio was annotating that both casts could be removed, but as soon as you remove one of them it stops saying you can remove the other one XD

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,800 @@
...
+                                          averagePosition.z + distalFingerBoneTip.z * fingerUsage);
+        //averageDirection = new Leap.Vector(averageDirection.x + medialFingerBoneDirection.x * fingerUsage,
+        //                                   averageDirection.y + medialFingerBoneDirection.y * fingerUsage,
+        //                                   averageDirection.z + medialFingerBoneDirection.z * fingerUsage);
+      }
+
+      // Distal Intention Ray: Add punch vector for when finger usage is low (closed fist).
+      // Vector3 distalAxis = _hand.DistalAxis();
+      float punchWeight = usageSum.Map(0F, 1F, 1F, 0F);
+      averagePosition = new Leap.Vector(averagePosition.x + _hand.Fingers[2].bones[1].PrevJoint.x * punchWeight,
```
Amarcolina: Looks like these comments didn't get addressed, any reason you are doing the component math manually instead of just using the vector class?

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,800 @@
...
+                                         averagePosition.y + _hand.Fingers[2].bones[1].PrevJoint.y * punchWeight,
+                                         averagePosition.z + _hand.Fingers[2].bones[1].PrevJoint.z * punchWeight);
+      //averageDirection = new Leap.Vector(averageDirection.x + distalAxis.x * punchWeight,
+      //                                   averageDirection.y + distalAxis.y * punchWeight,
+      //                                   averageDirection.z + distalAxis.z * punchWeight);
+      usageSum += punchWeight;
```
Amarcolina: Also it seems that usageSum would always be 1, since punchWeight is just (1-usageSum)

Assets/LeapMotionModules/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,711 @@
...
+          _closestHoveringHand = hand;
+          _closestHoveringHandDistance = handDistance;
+        }
+      }
+    }
+
+    public override void HoverEnd(Hand hand) {
+      _hoveringHandsCount--;
+      if (_hoveringHandsCount == 0) {
+        _closestJustStoppedHoveringHand = hand;
```
Amarcolina: Same question as before, this is named closest but what makes it the closest?  There are no more hovering hands, so isn't this just the most recent hovering hand?

Assets/LeapMotionModules/Interaction/InteractionBehaviour.cs
```diff
@@ -0,0 +1,711 @@
...
+        _collisionMode = desiredCollisionMode;
+        FixedUpdateLayer();
+      }
+
+      Assert.IsTrue((_collisionMode == CollisionMode.Grasped) == isGrasped);
+    }
+
+    protected void FixedUpdateLayer() {
+      int layer;
+      if (_collisionMode != CollisionMode.Normal) {
```
Amarcolina: Same comment as before, really prefer switch statements with a noisy default clause whenever dealing with enums.

Assets/LeapMotionModules/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,118 @@
...
+
+  public abstract class InteractionBehaviourBase : MonoBehaviour {
+
+    public InteractionManager interactionManager;
+
+    protected Rigidbody _body;
+    public new Rigidbody rigidbody { get { return _body; } }
+
+    [Header("Interaction Settings")]
+    public bool allowHover = true;
```
Amarcolina: Small thing, but it seems strange that it is called allowHover here, but enableHover in the manager.

Assets/LeapMotionModules/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,310 @@
...
+    void OnValidate() {
+      if (!Application.isPlaying && _autoGenerateLayers) {
+        AutoGenerateLayers();
+      }
+    }
+
+    private InteractionHand[] _interactionHands = new InteractionHand[2];
+    private HashSet<InteractionBehaviourBase> _interactionBehaviours = new HashSet<InteractionBehaviourBase>();
+    private Dictionary<Rigidbody, InteractionBehaviourBase> _rigidbodyRegistry = new Dictionary<Rigidbody, InteractionBehaviourBase>();
+    public Dictionary<Rigidbody, InteractionBehaviourBase> RigidbodyRegistry { get { return _rigidbodyRegistry; } }
```
Amarcolina: public Type name {get; private set;} would be nice .

Assets/LeapMotionModules/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,310 @@
...
+    public bool ReleaseObjectFromGrasp(InteractionBehaviourBase interactionObj) {
+      if (!_interactionBehaviours.Contains(interactionObj)) {
+        Debug.LogError("ReleaseObjectFromGrasp was called, but the interaction object " + interactionObj.transform.name + " is not registered"
+          + " with this InteractionManager.");
+        return false;
+      }
+
+      foreach (var hand in _interactionHands) {
+        if (hand.IsGrasping(interactionObj)) {
+          hand.ReleaseGrasp();
```
Amarcolina: Same as before, when two hands are holding onto an object you don't want to bail on the first hand.

Assets/LeapMotionModules/Interaction/InteractionBehaviourBase.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// Called by the InteractionManager every FixedUpdate, after
+    /// InteractionHands have FixedUpdated.
+    /// 
+    /// InteractionBehaviour uses this to fire per-object interaction events.
+    /// </summary>
+    public abstract void FixedUpdateObject();
+
+    protected virtual void Awake() {
+      _body = GetComponent<Rigidbody>();
+      _body.maxAngularVelocity = 100F;
```
Amarcolina: This should be a constant (public too!)

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -0,0 +1,800 @@
...
+      if (_contactInitialized && interactionManager.enableContact) {
+        EnableSoftContact();
+      }
+
+      _grabClassifier.NotifyGraspReleased(_graspedObject);
+      _graspedObject.GraspEnd(_hand);
+      _graspedObject = null;
+    }
+
+    public bool ReleaseObject(InteractionBehaviourBase interactionObj) {
```
Amarcolina: Open for discussion, but I would argue that any method that can fail should be prefixed with a Try to make it obvious it can fail.

Assets/LeapMotionModules/Interaction/InteractionBehaviour.cs
nickjbenson: Right?? weird that the compiler can't figure that situation out

Assets/LeapMotionModules/Interaction/InteractionHand.cs
nickjbenson: speed, and also lack of wanting to touch this method again until we know we want it in the first place. With John's interactivity work this entire "Intention Position" concept may well be useless

Assets/LeapMotionModules/Interaction/InteractionHand.cs
nickjbenson: Map clamps its input range.

Assets/LeapMotionModules/Interaction/InteractionBehaviourBase.cs
nickjbenson: Well the reason they were enableHover in the manager and ignoreHover originally was to emphasize that they are two different places in which you can ignore hover logic. You can have the manager ignore all hover logic, or you can override the manager's setting on a per-object basis. If they were named the same, it would be even more confusing. Come to think of it, I need to create a warning if allowHover/Contact/Grasp is checked but the corresponding enableX bool is not checked in the Manager, because the one does not override the other

Assets/LeapMotionModules/Interaction/InteractionHand.cs
Amarcolina: Ah of course, forgot about the clamping, makes sense!

Assets/LeapMotionModules/Interaction/InteractionBehaviourBase.cs
Amarcolina: Ah yeah I see, that makes sense hmmmmm.  That makes things tricky indeed, since like you say it is conceptually an override instead of an enabling.  I am fine with allow being the naming, and I agree that popping up a warning would be useful as well.

Assets/LeapMotionModules/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,650 @@
...
+            },
+            actionPerInteractionObject: (contactedObject, contactingIntHands) => {
+              contactedObject.StayGrasped(contactingIntHands);
+            });
+        }
+
+      }
+    }
+    
+    private delegate V StateChangeCheckFunc<H, T, V>(H hand, out T obj);
```
Amarcolina: Any reason these delegates are generic?  They are only used once, so you could just put the type declarations right here instead.

Assets/LeapMotionModules/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,650 @@
...
+
+        // Set brush layer to collide with nothing
+        Physics.IgnoreLayerCollision(_contactBoneLayer, i, true);
+      }
+
+      //After copy and set we enable the interaction between the brushes and interaction objects
+      Physics.IgnoreLayerCollision(_contactBoneLayer, _interactionLayer, false);
+    }
+
+    private class Pool<T> where T : new() {
```
Amarcolina: Core already has a pool definition you can use.

Assets/LeapMotionModules/Interaction/InteractionManager.cs
```diff
@@ -0,0 +1,650 @@
...
+        objHandsPair.Value.Clear();
+        s_intHandsListPool.Return(objHandsPair.Value);
+      }
+    }
+
+    /// <summary>
+    /// Checks object state per-hand, then calls an action per-object with all hand checks that reported back objects.
+    /// </summary>
+    private static void RemapMultiInteractionObjectStateChecks(
+        InteractionHand[]                                                        hands,
```
nickjbenson: ehhh just the sake of purity? It's a static transformation of methods. If it's bothersome I can change it

Assets/LeapMotionModules/Interaction/InteractionManager.cs
nickjbenson: same reason the other method was static. It was just easy to make generic. I can change it if desired

PR #634, opened 2017-03-29
Lots of graphic renderer features

Amarcolina:
- Moved space to core and gave it it's own namespace
 - Added a very basic text renderer.  Still very WIP and can throw errors, but a good start.
 - Removed lightmapping features from the baked renderer
 - More robust asset handling 
 - General bugfixes all around


nickjbenson:
Looks good to me!

PR #637, opened 2017-03-30
Broadphase; Basic Button Interaction

jselstad:
The standard Hover now takes into account spaces, so curved elements may now be be considered transparently with real ones for hovering.   If the primary hovered UI element occurs within a curved space, the interaction hand will warp to that space for interactions (this includes brush hands, soft contact, and the grab classifier).

Additionally, this PR adds a test scene with 50 buttons to stress-test the new collision system.  The buttons are based off of the "PhysicsUI.cs" buttons, but do away with the complicated hierarchy and springs.  Additionally, they only enable their own contact-ability when they are the "primary hover" (visualized by the red tinting).


Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -56,18 +58,25 @@ public class InteractionHand {
...
+        ProcessHoverCheckResults(hoverResults);
+        ProcessPrimaryHoverCheckResults(hoverResults);
+      }
 
-      hoverActivityManager.FixedUpdateHand(_hand);
-      CalculateIntentionPosition();
-      HoverCheckResults hoverResults = CheckHoverForHand(_hand, _hoverActivityManager.ActiveBehaviours);
-      ProcessHoverCheckResults(hoverResults);
-      ProcessPrimaryHoverCheckResults(hoverResults);
+      if (_hand != null && hoverResults.primaryHovered != null && hoverResults.primaryHovered.GetComponent<LeapGuiElement>()!=null) {
```
Amarcolina: Can't reference LeapGuiElement here, maybe ISpaceComponent instead?

Assets/LeapMotionModules/Interaction/Examples/Buttons/InteractionButton.cs
```diff
@@ -0,0 +1,129 @@
...
+    //[MinMax(0f, 0.1f)]
+    [Tooltip("The minimum and maximum heights the button can exist at.")]
+    public Vector2 MinMaxHeight = new Vector2(0f, 0.02f);
+    [Tooltip("The height that this button rests at; this value is a lerp in between the min and max height.")]
+    public float RestingHeight = 0.5f;
+
+    public UnityEvent OnPress = new UnityEvent();
+    public UnityEvent OnUnpress = new UnityEvent();
+
+    [HideInInspector]
```
Amarcolina: HideInInspector is for when you want a variable to remain serialized, but be invisible to the user.  Seems like what you want here might be [NonSerialized]? 

Also could alternatively use 
`public bool isDepressed {get; protected set;}`
Which would make it non serialized in addition to preventing people from messing with it.

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -76,26 +86,73 @@ private struct HoverCheckResults {
...
         checkedHand = hand
       };
 
-      foreach (var interactionObj in hoverCandidates) {
-        if (interactionObj.ignoreHover) continue;
-        CheckHoverForElement(hand, interactionObj, ref results);
+      //Loop through all the fingers (that we care about)
+      if (hand != null) {
+        float leastHandDistance = float.PositiveInfinity;
+        for (int i = 0; i < 3; i++) {
```
Amarcolina: I had to do some digging to figure out which of the fingers this referred to, it would be nice if you had a list of FingerType's as a constant somewhere that you could reference instead of using the finger indexes directly.

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -76,26 +86,73 @@ private struct HoverCheckResults {
...
+          if (!hand.Fingers[i].IsExtended) { continue; }
+          float leastFingerDistance = float.PositiveInfinity;
```
Amarcolina: Uh this variable is never modified?

Amarcolina: And neither is leastHandDistance

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -76,26 +86,73 @@ private struct HoverCheckResults {
...
     private HashSet<InteractionBehaviourBase> _hoverableCache = new HashSet<InteractionBehaviourBase>();
+    public InteractionBehaviourBase[] tempPerFingerHovered = new InteractionBehaviourBase[3];
+
     private HoverCheckResults CheckHoverForHand(Hand hand, HashSet<InteractionBehaviourBase> hoverCandidates) {
       _hoverableCache.Clear();
+      Array.Clear(tempPerFingerHovered, 0, tempPerFingerHovered.Length);
 
       HoverCheckResults results = new HoverCheckResults() {
         hovered = _hoverableCache,
+        perFingerHovered = tempPerFingerHovered,
```
Amarcolina: perFingerHovered is never used by anybody

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -56,18 +58,26 @@ public class InteractionHand {
       }
     }
 
+    private HoverCheckResults hoverResults = new HoverCheckResults();
```
Amarcolina: This is an instance field, but it is only ever used within FixedUpdateHovering.

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -76,26 +86,73 @@ private struct HoverCheckResults {
...
+      Vector3 localPalmPos = element.anchor.space.transform.InverseTransformPoint(inHand.PalmPosition.ToVector3());
+      Quaternion localPalmRot = element.anchor.space.transform.InverseTransformRotation(inHand.Rotation.ToQuaternion());
+
+      inHand.SetTransform(element.anchor.space.transform.TransformPoint(element.anchor.transformer.InverseTransformPoint(localPalmPos)),
+                          element.anchor.space.transform.TransformRotation(element.anchor.transformer.InverseTransformRotation(localPalmPos, localPalmRot)));
+    }
+
+    private void CheckHoverForElement(Vector3 position, InteractionBehaviourBase behaviour, ISpaceComponent element, int whichFinger, float leastFingerDistance, float leastHandDistance, ref HoverCheckResults curResults) {
+      //NEED BETTER DISTANCE FUNCTION
+      float dist = Vector3.SqrMagnitude(((Component)element).transform.position - transformPoint(position, element));
```
Amarcolina: this function needs to match hoverable.GetDistance or else you might get weird comparisons between objects that are inside a curved space vs objects that are not

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -218,8 +229,7 @@ private struct HoverCheckResults {
 
     private void FixedUpdateTouch() {
       _touchActivityManager.activationRadius = interactionManager.WorldTouchActivationRadius;
-
-      _touchActivityManager.FixedUpdateHand(_hand);
+      _touchActivityManager.FixedUpdateHand((_hand != null) ? _hand.PalmPosition.ToVector3() : Vector3.zero);
```
Amarcolina: Wait if the hand is null, do we want to update activity at all?  I would think updating the activity assuming the hand is at the origin might result in some strange behavior

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -657,7 +667,7 @@ private struct HoverCheckResults {
     private void FixedUpdateContactCallbacks() {
       foreach (var interactionObj in _contactBehavioursLastFrame) {
         if (!_contactBehaviours.ContainsKey(interactionObj)
-          || !_brushBoneParent.gameObject.activeSelf) {
+         || !_brushBoneParent.gameObject.activeSelf) {
```
Amarcolina: watch out, even if activeSelf returns true the object can still be inactive if a parent was deactivated.  You probably want to use activeInHierarchy instead.

Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -56,18 +58,26 @@ public class InteractionHand {
...
 
-      hoverActivityManager.FixedUpdateHand(_hand);
-      CalculateIntentionPosition();
-      HoverCheckResults hoverResults = CheckHoverForHand(_hand, _hoverActivityManager.ActiveBehaviours);
-      ProcessHoverCheckResults(hoverResults);
-      ProcessPrimaryHoverCheckResults(hoverResults);
+      ISpaceComponent space;
+      if (_hand != null && hoverResults.primaryHovered != null && (space = hoverResults.primaryHovered.GetComponent<ISpaceComponent>()) != null) {
+        //Transform bulk hand to the closest element's warped space
+        coarseInverseTransformHand(_hand, space);
```
Amarcolina: Do you need to also transform _handData here too?

Assets/LeapMotionModules/Interaction/InteractionHand.cs
Amarcolina: sounds like we should create some sort of method in the activity manager called ClearHand or ClearActive or something.  All it needs to do at that point is clear the active list right?

Amarcolina: Or I guess we could just have FixedUpdateHand check to see if the hand is null, and if so, clear the active list and return.

Assets/LeapMotionModules/Interaction/ActivityManager.cs
```diff
@@ -33,38 +32,49 @@ public class ActivityManager {
       this.activationRadius = activationRadius;
     }
 
-    public void FixedUpdateHand(Hand hand) {
-      int count = GetSphereColliderResults(hand, ref _colliderResultsBuffer);
-      UpdateActiveList(count, _colliderResultsBuffer);
-    }
+    public void FixedUpdatePosition(Vector3 palmPosition, List<LeapSpace> spaces = null) {
+      using (new ProfilerSample("Update "+ (spaces==null?"Touch" : "Hover") + " Actvity Manager")) {
```
Amarcolina: Performing a string concat creates garbage, unfortunately, which will persist even if profiling is disabled :(

jselstad:
Git shenanigans

PR #642, opened 2017-04-03
Broadphase; Basic Button Interaction

jselstad:
The standard Hover now takes into account spaces, so curved elements may now be be considered transparently with real ones for hovering. If the primary hovered UI element occurs within a curved space, the interaction hand will warp to that space for interactions (this includes brush hands, soft contact, and the grab classifier).

Additionally, this PR adds a test scene with 50 buttons to stress-test the new collision system. The buttons are based off of the "PhysicsUI.cs" buttons, but do away with the complicated hierarchy and springs. Additionally, they only enable their own contact-ability when they are the "primary hover" (visualized by the red tinting).


jselstad:
Oh boy, the above doesn't look good...

PR #648, opened 2017-04-05
Inherit features and added promp

Amarcolina:
When adding a new renderer, it will now inherit the features of the graphics that are already present.  It will take the union of all features, so if graphic A has feature X, and graphic B has feature Y, the rendering method will gain features X and Y.  

When changing the group of a graphic, pop a dialog box if the new group has a different feature set than the current group (which could result in data loss).  This dialog box can also be turned off in the preferences, but is enabled by default.

### DIFF
--- a/Assets/LeapMotion/Scripts/Utils/Utils.cs
+++ b/Assets/LeapMotion/Scripts/Utils/Utils.cs
@@ -59,6 +59,55 @@ namespace Leap.Unity {
       array = newArray;
     }
 
+    /// <summary>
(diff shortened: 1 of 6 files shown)


jselstad:
Hmm, okay, the texture features still get destroyed when you switch to a renderer without those features and switch back, but this should be enough robustness to get started.

jselstad:
I'm not seeing the dialog box though?  Switching from Baked -> Text just makes all my features red (but it keeps them, so I can switch back which is goooood).

PR #653, opened 2017-04-10
Upgrade blend shapes

Amarcolina:
- Grab blend shapes directly from the Mesh object instead of specifying a different mesh
 - Added ability to specify a custom transform offset using an actual transform
 - Changed method signature to be more efficient and better represent data being passed

To test:
 - [ ] Visit scene and mess with the blend shape sliders to see the blend shape in action!


nickjbenson:
Cool!

PR #655, opened 2017-04-12
Added ImplementsInterface attribute

nickjbenson:
A nice convenience for (sort of) being able to serialize an interface -- really just allows you to require that a MonoBehaviour dragged into an inspector field implements a desired interface.

### DIFF
--- a/Assets/LeapMotion/Scripts/Attributes/ImplementsInterface.cs
+++ b/Assets/LeapMotion/Scripts/Attributes/ImplementsInterface.cs
@@ -1,0 +1,52 @@
+﻿using UnityEngine;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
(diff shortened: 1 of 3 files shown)


Assets/LeapMotion/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -0,0 +1,52 @@
...
+    public override IEnumerable<SerializedPropertyType> SupportedTypes {
+      get {
+        yield return SerializedPropertyType.ObjectReference;
+      }
+    }
+
+#endif
+  }
+
+  public static class ImplementsInterfaceExtention {
```
Amarcolina: Probably would be good to move this to Utils.cs so that we can use it without importing Leap.Unity.Attributes

PR #658, opened 2017-04-14
Add some Capsule Collider utils to Utils.cs

nickjbenson:
Helpful for Lines app; we also deal with capsules a lot with hands so we might find use.

### DIFF
--- a/Assets/LeapMotion/Scripts/Utils/Utils.cs
+++ b/Assets/LeapMotion/Scripts/Utils/Utils.cs
@@ -167,6 +167,34 @@ namespace Leap.Unity {
 
     #endregion
 
+    #region Collider Utils
+
+    public static Vector3 GetDirection(this CapsuleCollider capsule) {
+      return capsule.direction == 0 ? Vector3.right : capsule.direction == 1 ? Vector3.up : Vector3.forward;
+    }
+
+    /// <summary>
+    /// Manipulates capsule.transform.position, capsule.transform.rotation, and capsule.height
+    /// so that the line segment defined by the capsule connects world-space points a and b.
+    /// </summary>
+    public static void SetCapsulePoints(this CapsuleCollider capsule, Vector3 a, Vector3 b) {
+      capsule.center = Vector3.zero;
+
+      capsule.transform.position = (a + b) / 2F;
+
+      Vector3 capsuleDirection = capsule.GetDirection();
+
+      Vector3 capsuleDirWorldSpace = capsule.transform.TransformDirection(capsuleDirection);
+      Quaternion necessaryRotation = Quaternion.FromToRotation(capsuleDirWorldSpace, a - capsule.transform.position);
+      capsule.transform.rotation = necessaryRotation * capsule.transform.rotation;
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -167,6 +167,34 @@ public static class Utils {
 
     #endregion
 
+    #region Collider Utils
+
+    public static Vector3 GetDirection(this CapsuleCollider capsule) {
+      return capsule.direction == 0 ? Vector3.right : capsule.direction == 1 ? Vector3.up : Vector3.forward;
```
Amarcolina: Could we use a switch instead of nested ternary statements?  I think it would really help readability.

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -167,6 +167,34 @@ public static class Utils {
...
+
+      capsule.transform.position = (a + b) / 2F;
+
+      Vector3 capsuleDirection = capsule.GetDirection();
+
+      Vector3 capsuleDirWorldSpace = capsule.transform.TransformDirection(capsuleDirection);
+      Quaternion necessaryRotation = Quaternion.FromToRotation(capsuleDirWorldSpace, a - capsule.transform.position);
+      capsule.transform.rotation = necessaryRotation * capsule.transform.rotation;
+
+      Vector3 aWorldSpace = capsule.transform.InverseTransformPoint(a);
```
Amarcolina: To get world space you want to use TransformPoint, not InverseTransformPoint.

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -167,6 +167,34 @@ public static class Utils {
...
+      float capsuleSpaceDistToA = aWorldSpace.magnitude;
```
Amarcolina: Shouldn't this be (capsule.transform.position - aWorldSpace).magnitude?

Amarcolina: Ooooh i see what's going on, aWorldSpace is not in world space, but it's not supposed to be.  You are getting the transform local distance from the position to a, makes sense!

So in that case my only suggested change here would be to rename aWorldSpace to aLocalSpace or aCapsuleSpace.

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: done

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: logic is correct, name just should have been aCapsuleSpace

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: as above response

PR #667, opened 2017-04-19
Pulled changed over from the builder repo

Amarcolina:
- Add the ability to 'defer' property drawing using the CustomEditorBase
 - Added the EnumEventTable to allow us to add a rich set of callbacks anywhere we like!
 - Improved shader-side of the dynamic renderer to support custom channels
 - Fixed a number of misc bugs
 - Made sure the thing can actually build!
 - Removed O(n^2) bottleneck in mesh refresh code

### DIFF
--- a/Assets/LeapMotion/Editor/CustomEditorBase.cs
+++ b/Assets/LeapMotion/Editor/CustomEditorBase.cs
@@ -25,6 +25,7 @@ namespace Leap.Unity {
     protected Dictionary<string, Action<SerializedProperty>> _specifiedDrawers;
     protected Dictionary<string, List<Action<SerializedProperty>>> _specifiedDecorators;
     protected Dictionary<string, List<Func<bool>>> _conditionalProperties;
+    protected List<string> _deferredProperties;
     protected bool _showScriptField = true;
 
     protected List<SerializedProperty> _modifiedProperties = new List<SerializedProperty>();
@@ -118,6 +119,19 @@ namespace Leap.Unity {
       }
     }
 
+    /// <summary>
+    /// Defer rendering of a property until the end of the inspector.  Deferred properties
+    /// are drawn in the REVERSE order they are deferred!  NOT by the order they appear in the 
+    /// serialized object!
+    /// </summary>
+    protected void deferProperty(string propertyName) {
+      if (!validateProperty(propertyName)) {
+        return;
+      }
+
+      _deferredProperties.Insert(0, propertyName);
+    }
+
     protected void drawScriptField(bool disable = true) {
       var scriptProp = serializedObject.FindProperty("m_Script");
       EditorGUI.BeginDisabledGroup(disable);
@@ -136,6 +150,7 @@ namespace Leap.Unity {
       _specifiedDrawers = new Dictionary<string, Action<SerializedProperty>>();
       _specifiedDecorators = new Dictionary<string, List<Action<SerializedProperty>>>();
       _conditionalProperties = new Dictionary<string, List<Func<bool>>>();
+      _deferredProperties = new List<string>();
     }
 
     protected bool validateProperty(string propertyName) {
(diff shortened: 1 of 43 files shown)


Assets/LeapMotionModules/Interaction/InteractionHand.cs
```diff
@@ -326,7 +326,7 @@ public struct HoverCheckResults {
 
     private const int NUM_FINGERS = 5;
     private const int BONES_PER_FINGER = 3;
-    public const float PER_BONE_MASS_MULTIPLIER = 0.2F;
+    public const float PER_BONE_MASS_MULTIPLIER = 10F;
```
jselstad: Ack this is going to completely change the feel of the interactions in the normal interaction scenes.

It's a shame there's two orders of magnitude difference in how you want to interact with UI vs normal objects...

Assets/LeapMotionModules/Interaction/Editor/InteractionBehaviourEditor.cs
```diff
@@ -0,0 +1,32 @@
...
+      base.OnEnable();
+
+      deferProperty("_eventTable");
+      specifyCustomDrawer("_eventTable", drawEventTable);
+
+      specifyConditionalDrawing(() => !target.ignoreGrasping,
+                   "_allowMultiGrasp",
+                   "_moveObjectWhenGrasped",
+                   "graspedMovementType",
+                   "graspHoldWarpingEnabled__curIgnored");
```
jselstad: I would also like to draw attention to the drawing changes I made here, @nickjbenson 

These properties had [DisableIf] attributes on them before; I assume they were there because it was the farthest you could get without writing an editor script...

Assets/LeapMotion/Editor/CustomEditorBase.cs
```diff
@@ -118,6 +119,19 @@ public class CustomEditorBase : Editor {
       }
     }
 
+    /// <summary>
+    /// Defer rendering of a property until the end of the inspector.  Deferred properties
+    /// are drawn in the REVERSE order they are deferred!  NOT by the order they appear in the
```
nickjbenson: What's the reasoning for drawing deferred properties in the _opposite_ order in which they are deferred? My intuition was initially that deferred properties would just get added in the order they are deferred to some big "deferred" group that renders last.

Assets/LeapMotion/Scripts/Utils/Editor/EnumEventTableEditor.cs
```diff
@@ -0,0 +1,126 @@
...
+
+namespace Leap.Unity {
+
+  public class EnumEventTableEditor {
+
+    private SerializedProperty _entries;
+
+    private GUIContent _iconToolbarMinus;
+    private GUIContent _eventIDName;
+    private GUIContent _addButonContent;
```
nickjbenson: spelling

Assets/LeapMotionModules/GraphicRenderer/Scripts/EditTimeApis/LeapGraphicRendererEditorApi.cs
```diff
@@ -121,32 +121,58 @@ public class EditorApi {
       }
 
       public void DoEditorUpdateLogic(bool fullRebuild, bool heavyRebuild) {
-        validateSpaceComponent();
-
-        if (fullRebuild) {
-          if (_renderer._space != null) {
-            _renderer._space.RebuildHierarchy();
-            _renderer._space.RecalculateTransformers();
+        using (new ProfilerSample("A")) {
```
nickjbenson: ?

Assets/LeapMotionModules/GraphicRenderer/Scripts/LeapGraphic.cs
```diff
@@ -8,7 +8,7 @@ namespace Leap.Unity.GraphicalRenderer {
 
   [ExecuteInEditMode]
   [DisallowMultipleComponent]
-  public abstract partial class LeapGraphic : MonoBehaviour, ISpaceComponent {
+  public abstract partial class LeapGraphic : MonoBehaviour, ISpaceComponent, IEquatable<LeapGraphic> {
```
nickjbenson: Did you experience a noticeable difference due to the overhead reduction implementing IEquatable provides when used in HashSets and Dictionaries? If so I should probably make InteractionBehaviours and any other particularly important components also implement IEquatable...

Assets/LeapMotionModules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
```diff
@@ -148,35 +148,56 @@ public partial class LeapGraphicRenderer : MonoBehaviour {
...
+
       foreach (var graphic in _tempGraphicList) {
         if (graphic.isAttachedToGroup) {
           //procede to validate
 
-          //If the graphic is anchored to the wrong anchor, detach and reattach
-          var anchor = _space == null ? null : LeapSpaceAnchor.GetAnchor(graphic.transform);
-          if (graphic.anchor != anchor) {
-            var group = graphic.attachedGroup;
+          using (new ProfilerSample("1")) {
```
nickjbenson: Will ProfilerSample names like this look confusing when actually profiling? (Might not, hard to tell the outer sample context from this diff view)

Assets/LeapMotionModules/Interaction/InteractionHand.cs
nickjbenson: go go InteractionBehaviour InteractionMassSetting.Default/UI!

Assets/LeapMotionModules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
Amarcolina: Ah my bad, these were left in from a test, they are not supposed to be here D:

Assets/LeapMotion/Scripts/Utils/Editor/EnumEventTableEditor.cs
Amarcolina: Don't blame me, blame unity!
https://bitbucket.org/Unity-Technologies/ui/src/0155c39e05ca5d7dcc97d9974256ef83bc122586/UnityEditor.UI/EventSystem/EventTriggerEditor.cs?at=5.2&fileviewer=file-view-default

Assets/LeapMotion/Editor/CustomEditorBase.cs
Amarcolina: The reasoning is to deal with tall hierarchies of objects that all want to specify deferred properties.  Imagine we have class A, B and C.  C derives from B, and B derives from A.  A defers properties A1 and A2, B defers properties B3 and B4, and C defers properties C5 and C6.

If we drew properties in order, we would see:
```
normal properties of A
normal properties of B
normal properties of C
A1
A2
B3
B4
C5
C6
```

However I think that it would be much nicer if we saw this instead:
```
normal properties of A
normal properties of B
normal properties of C
C5
C6
B3
B4
A1
A2
```

Kind of an outside-in approach.  Class A defines what comes first and last and leaves a space in between for deriving classes.

Amarcolina: Great example of this is the slider class, which contains all of it's events at the bottom in a nice order thanks to the reverse order they are added into the deferred list.

Assets/LeapMotionModules/GraphicRenderer/Scripts/LeapGraphic.cs
Amarcolina: I could have sworn that I did, but I can't replicate it anymore :( I only added it because I saw something in the profiler, and making this change made it go away, but I have no idea what it is anymore...

Assets/LeapMotion/Editor/CustomEditorBase.cs
nickjbenson: OK, that sounds fine. Could you add a line in the summary like... "This way, deriving classes can have their deferred inspector fields rendered before their parent classes' deferred inspector fields"?

PR #676, opened 2017-04-20
Add some helpful UI components to Core & Attachments

nickjbenson:
This is a low-priority PR; I just don't want these scripts to fall through the cracks.

They could be useful for Martin and Barrett though, if they look good. Otherwise, I'll probably fix them sometime next week.


Assets/LeapMotion/Scripts/Animation/TransformTweenBehaviour.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Leap.Unity.Animation {
+
+  public class TransformTweenBehaviour : MonoBehaviour {
```
Amarcolina: We should either mark this with AddComponentMenu("") as an example script, or flesh out the interface a bit more.  Things like tween time and smoothing type for example.

Amarcolina: Would also be nice (if we are trying to make it a script we want devs to use) to either add in some warning on how this script itself doesn't do anything, or we allow it to do stuff :D

Assets/LeapMotion/Scripts/Animation/TransformTweenBehaviour.cs
```diff
@@ -0,0 +1,41 @@
...
+      _tween = Tween.Persistent().Target(targetTransform)
+                                   .LocalPosition(startTransform, endTransform)
+                                   .Target(targetTransform)
+                                   .LocalRotation(startTransform, endTransform)
+                                   .Target(targetTransform)
+                                   .LocalScale(startTransform, endTransform)
+                                   .OverTime(0.33F)
+                                   .Smooth(SmoothType.Smooth);
+
+      _tween.progress = 0.0001F;
```
Amarcolina: Just so I understand, this is to force the tween state to update at the given percent (0%) but since it's lazy it won't work if you just say 0% right?

Assets/LeapMotionModules/Attachments/Scripts/Anchors/Anchor.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Leap.Unity.Attachments {
+
+  public class Anchor : MonoBehaviour, IRuntimeGizmoComponent {
+
+    public float anchorRange = 0.1F;
+
+    public bool IsWithinRange(Vector3 position) {
+      Vector3 delta = this.transform.position - position;
+      return delta.x * delta.x + delta.y * delta.y + delta.z * delta.z < anchorRange * anchorRange;
```
Amarcolina: Any reason not to use delta.sqrMagnitude instead of the long form?

Assets/LeapMotionModules/Attachments/Scripts/Anchors/AnchorGroup.cs
```diff
@@ -0,0 +1,35 @@
...
+      return anchors.Contains(anchor);
+    }
+
+    /// <summary> Returns the closest anchor to the argument position. By default, this method will
+    /// only return anchors that are enabled. </summary>
+    public Anchor FindClosestAnchor(Vector3 fromPosition, bool requireAnchorIsEnabled = true) {
+      Anchor closestAnchor = null;
+      float closestDistSqrd = float.PositiveInfinity;
+      foreach (Anchor anchor in anchors) {
+        if (requireAnchorIsEnabled && !anchor.enabled) continue;
```
Amarcolina: Probably should check for null Anchor's here, in case the user messed with the config

Assets/LeapMotionModules/Attachments/Scripts/Anchors/AnchorableBehaviour.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    private bool _attachedToAnchor = false;
+
+    void OnValidate() {
+      _isInteractionBehaviourNull = interactionBehaviour == null;
+    }
+
+    void OnEnable() {
+      if (anchorType == AnchorType.SingleAnchor && anchor == null) {
+        this.enabled = false;
```
Amarcolina: Probably would be good to throw warnings in this case.  Objects refusing to be enabled could be potentially confusing otherwise

Assets/LeapMotion/Scripts/Animation/TransformTweenBehaviour.cs
nickjbenson: Yeah. I do wish there was a better way but this has been my go-to strategy; I don't think I ever would have been burned if setting progress was never lazy FWIW

Assets/LeapMotionModules/Attachments/Scripts/Anchors/Anchor.cs
nickjbenson: this was during my "RAW SPEEEED" phase, I can change it for readability, sometimes I feel compelled to not introduce a function call :P

Assets/LeapMotion/Scripts/Animation/TransformTweenBehaviour.cs
nickjbenson: K, I fleshed out a lot of things for it so we feel comfortable with it as a real script

Assets/LeapMotionModules/Attachments/Scripts/Anchors/AnchorableBehaviour.cs
nickjbenson: Whew! Just spent a while changing how working with AnchorableBehaviours work so the inspector is less confusing to work with. Now it doesn't get enabled/disabled willy-nilly, the anchorable will just-work if it's enabled and has a valid anchor configuration, otherwise it won't do anything (and will print appropriate errors if it's being set up incorrectly).

nickjbenson:
Oh! Hold on, don't merge yet, I need to move Anchors into the IE module because they have a dependency on it :\

nickjbenson:
OK! This is ready to merge

PR #679, opened 2017-04-21
Allow users to drag/drop feature data into fields

Amarcolina:


nickjbenson:
Yeah this is fine for me.

I think it would be cool to have a tooltip somewhere that could mention "You can drag Feature Data headers into PlayMaker to edit their properties", but maybe that belong in a Getting Started guide instead.

nickjbenson:
Feel free to add some tooltip before merging or just go ahead and merge @Amarcolina

PR #687, opened 2017-04-25
Add package assets

nickjbenson:
Since I provided IE 0.4.0 and the Renderer to QA, made the relevant assets for exporting the .unitypackages.


Amarcolina:
Looks great!

PR #688, opened 2017-04-25
Import Lines changes

nickjbenson:
- InteractionHands are MonoBehaviours now (this is a smaller change than you might expect), this allows a better API when you have InteractionHand objects or you want to add special interaction scripts to individual hands (this is most of the change to InteractionHand, along with generic ActivityManager)
- Bugfix in DeltaBuffer
- Make SingleLayer IEquatable because I like cleanly using it in hashsets
- Alex-provided fix in GR/BakedUnlit.shader
- Big one: Made ActivityManager generic, and provided the appropriate setup in InteractionHand


Assets/LeapMotionModules/Interaction/ActivityManager.cs
```diff
@@ -116,13 +135,9 @@ public class ActivityManager {
...
     private void UpdateActiveList(int numResults, Collider[] results) {
       for (int i = 0; i < numResults; i++) {
-        if (results[i].attachedRigidbody != null) {
-          Rigidbody body = results[i].attachedRigidbody;
-          IInteractionBehaviour interactionObj;
-          if (body != null && manager.rigidbodyRegistry.TryGetValue(body, out interactionObj)
-              && filter != null && filter(interactionObj)) {
-            _activeBehaviours.Add(interactionObj);
-          }
+        T obj = filter(results[i]);
```
Amarcolina: Seems that there is no verification that filer is non-null?

nickjbenson: added an early return + warning.

Assets/LeapMotionModules/Interaction/ActivityManager.cs
```diff
@@ -6,99 +6,118 @@
...
+    /// this is highly inefficient.
+    /// 
+    /// See SingleLayer and use bitwise operations on their layerMasks for a convenient way to
+    /// express layer masks.
+    /// <see cref="SingleLayer"/>
+    /// </summary>
+    public int activationLayerMask = ~0;
+
+    /// <summary> If non-null, only objects with at least one Collider for which this function
+    /// returns a non-null T will be added to the active behaviours list. </summary>
```
Amarcolina: I guess the reason I was confused is that this summary seems to imply that a null filter is a valid state for the manager.  It seems like that is not the case though?

nickjbenson: Yeah that's not a valid state. And it's because there's no way to know how to convert a Collider reference to a T... I _could_ place a restriction that T has to be a MonoBehaviour and use a default filter of returning Collider.GetComponent<T>? Or even GetComponentInParent<T>? But I'd be worried about default performance in the latter case

Amarcolina: I'm totally fine with filter == null being invalid, I just think the summary is a little misleading.  
'If non-null, only objects...' seems to imply that the filter simply provides a restriction on an existing method, rather than being actually the method itself!

So an update of the docs is all I'm proposing

nickjbenson: Ahhh, that makes perfect sense. Will update!

Amarcolina:
Looks good!

PR #690, opened 2017-04-26
First Three Examples

nickjbenson:
- Tidy "Basic Example" into the first of our series of "atomic examples," explaining the core trio of interaction scripts
- Add a "Basic UI" example (replacing the current "Buttons" example) to be independent of the Graphic Renderer and be a pretty bare-bones implementation of the InteractionButton/Slider scripts.
- Organize the "handles" example into a demonstration of Interaction Callbacks, and add some explanation in the scene.
- Various bits of tidying.


Amarcolina:
Awesome, looks great to me!

PR #692, opened 2017-04-26
Remove Compatibility IE from Develop

jselstad:
This will allow the younger, newer IE to take its name place as the dominant IE.

The old IE lives in the "Compatibility-IE" branch (with a .gitignore on the new IE so changes in core can be merged back into it).


Amarcolina:
Definitely want to wait until the release of core and stuff goes out.  Do we know for sure whether or not that has happened yet?

nickjbenson:
No it hasn't happened yet, I'm in communication with Matt about it.

jselstad:
Whelp, that was a bust; let's try this again once the IE compatibility release is out.

PR #694, opened 2017-04-28
Adding editor resources class

Amarcolina:
Use to load resources stored inside a folder named EditorResources, just like the Resources class!


nickjbenson:
Crashes with f3 on Oculus currently but `develop` is known to have f3-crashing settings right now with Oculus, and it doesn't break the build when merged into `feature-leap-gui` so I'm going to merge.

PR #697, opened 2017-04-28
Adding an auto-header tool

Amarcolina:
Adding a simple tool that allows us to add/update copywrite headers for all our .cs files.  Simply go to Assets->Update Copywrite Headers to trigger the update.  Visit Assets/LeapMotionModules/AutoHeader/Editor/AutoCopywriteHeader.cs to modify the copywrite notice.

I elected to put this code in UnityModules just because I didn't think it made much sense to put it in LeapMotion and pollute our users menu space.

To test:
 - [x] Verify update process works properly
 - [x] Verify there are no scripts it has missed
 - [x] Verify code looks good


jselstad:
...Perhaps it should only affect files inside of LeapMotion and LeapMotionModules (I don't think we wrote UnityTestTools?)

Amarcolina:
Oh snap, good catch, will make that change.

protodeep:
@jselstad Yes, confirmed that "Proprietary & confidential" is correct.

PR #701, opened 2017-05-02
Removing Pepper hands models and prefabs and updating Hands_Viewer_De…

protodeep:
- Removing Pepper hands models and prefabs from Unity Modules
- Updating Hands_Viewer_Demo.unity

- [ ] Run Hands_Viewer_Demo.unity and verify there are no errors


nickjbenson:
This is fine. Note that 5.6.0f3 Oculus builds are broken, but this is a known issue with `develop` and is fixed in `feature-leap-gui`. Just want to make sure we remember this.

PR #706, opened 2017-05-04
Merge 54b36a2e... from Develop to Master

nickjbenson:
For getting a proper release hash for Core and IE.

This workflow sucks. It's a huge pain to get anything released. We should try and make it better.


testmasterbright:
@nickjbenson agreed. There are so many manual aspects of releasing. I wish we could automate a lot of it.

PR #707, opened 2017-05-04
Updating and fixing multi-select for attributes

Amarcolina:
Component references were simply ignored for multi-select situations.  Now the default and only way is to reference an array of all selected objects.  This prevents accidental creation of attributes that only support single selection in the future.  

OnEditorChange has been updated to properly dispatch callbacks to *all* selected objects instead of just throwing tantrum.  AutoFind also had to be changed, since it also used the component reference, but AutoFind was left to be a component that only works when single-selecting.  

To test:
 - [ ] Make sure both OnEditorChange and AutoFind still work properly!
 - [ ] Make sure OnEditorChange correctly dispatches callbacks to multiple selected objects


nickjbenson:
Yep! Awesome.

PR #711, opened 2017-05-08
Progress for Dynamic UI examples

nickjbenson:
- Add Example 4 (Hand UI) and Example 5 (Anchors) -- Example 6 to come in the next PR
- Examples 1-4 should have consistent textual descriptions (Example 5 text to come later)
- Added AttachmentHands.cs in Core, would like a bit of usage testing and feedback on workflow -- if it is satisfactory, I think it should replace/deprecate the Attachments Module (will look into that more for the next PR as well)
- Example 5 firmly places Anchors in the Interaction namespace; there's a lot of functionality in them now, but they'll be getting heavier changes for the next PR (will post about changes for reactivity)
- TransformTweenBehaviour got some dogfooding in Example 4, would be good to get feedback on its usage

Evaluating:
- Make sure Examples 1-5 work as-advertised
- Mess with AttachmentHands in the Editor (Example 4 and 5), make sure it's not possible to easily destroy user-created data in the editor and the workflow makes sense


Assets/LeapMotion/Scripts/Attachments/AttachmentHands.cs
```diff
@@ -0,0 +1,183 @@
...
+  /// for visuals or for logic. The AttachmentHands object will maintain two child
+  /// objects, one for each of the player's hands. Use the Inspector to customize
+  /// which points you'd like to see in the hierarchy beneath the individual
+  /// AttachmentHand objects.
+  /// </summary>
+  [ExecuteInEditMode]
+  public class AttachmentHands : MonoBehaviour {
+
+    [SerializeField]
+    [OnEditorChange("attachmentPoints")]
```
Amarcolina: Is this OnEditorChange needed?  I don't think there is ever a point where this field becomes visible in the inspector.

Assets/LeapMotion/Scripts/Attachments/AttachmentHand.cs
```diff
@@ -0,0 +1,429 @@
...
+                                                     pinkyTip);
+      if (topLevelTransform != null) {
+        topLevelTransform.SetSiblingIndex(siblingIdx++);
+      }
+    }
+
+    private static Transform[] s_hierarchyTransformsBuffer = new Transform[4];
+    /// <summary>
+    /// Tries to build a parent-child stack (index 0 is the first parent) of the argument
+    /// transforms (they might be null) and returns the number of transforms that were non-null.
```
Amarcolina: Documentation seems out of date, it returns a transform

Assets/LeapMotion/Scripts/Attachments/AttachmentHand.cs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    [ThreadStatic]
+    private static AttachmentPointBehaviour[] s_attachmentPointsBuffer = new AttachmentPointBehaviour[32];
+
+    /// <summary>
+    /// An enumerator that traverses all of the existing AttachmentPointBehaviours beneath an
+    /// AttachmentHand.
+    /// </summary>
+    public struct AttachmentPointsEnumerator {
```
Amarcolina: Because you are using a static backing array it makes it impossible to have two enumerators active at the same time.

Assets/LeapMotion/Scripts/Attachments/AttachmentHands.cs
nickjbenson: It's definitely getting called when I hit the checkboxes, because that is the only reason Undo works. ¯\_(ツ)_/¯

PR #715, opened 2017-05-10
Make buttons flip out if they're not childed

jselstad:
Also make buttons ungraspable if they're not the primary hover.

Also fix bug that you may remember from demos such as: today's front-end sync.

### DIFF
--- a/Assets/LeapMotionModules/Interaction/Scripts/UI/InteractionButton.cs
+++ b/Assets/LeapMotionModules/Interaction/Scripts/UI/InteractionButton.cs
@@ -74,6 +74,7 @@ namespace Leap.Unity.Interaction {
     private bool _handIsLeft = false;
     private bool _physicsOccurred;
     private bool initialIgnoreGrasping = false;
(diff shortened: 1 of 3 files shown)


Assets/LeapMotionModules/Interaction/Scripts/UI/InteractionButton.cs
```diff
@@ -72,14 +73,23 @@ public class InteractionButton : InteractionBehaviour {
     private Vector3 _physicsVelocity = Vector3.zero;
     private bool _handIsLeft = false;
     private bool _physicsOccurred;
+    private bool initialIgnoreGrasping = false;
+    private Quaternion initialLocalRotation;
 
     protected override void Start() {
+      if(transform == transform.root) {
+        Debug.LogError("This button has no parent!  Please ensure that it is parented to something!", this);
```
nickjbenson: I'm gonna be a stickler and say probably only one exclamation mark is necessary here (the second one can be a period and then we're not! yelling as much!)

PR #717, opened 2017-05-11
Support reading from profiling recordings

Amarcolina:
Now we can read from anything we have already defined with ProfilerSample blocks!  Simply add the string name of the block being sampled to the 'Unity Sampler Names' array and have it show up ingame on the realtime graph.  We can also add names that we have not defined, but that are built in to unity.  For example we can take a look at BehaviourUpdate directly!  

In the future we can probably rewrite realtime graph completely to rely only on the new profiler api's instead of doing manual/hacky stopwatching.  

To test:
 - [ ] Verify that graph still works the same way
 - [ ] Add a built-in name like BehaviourUpdate to the list, and make sure that works too!


nickjbenson:
Noting that this is broken in f3 but works in p3, I'm going to approve and merge.

PR #719, opened 2017-05-12
Documentation for query

Amarcolina:
- Added basic documentation to all query ops
 - Fixed some out of date documentation
 - Fixed cases where the queries could get into an invalid state after you used them once
 - Added the Last and Single operations
 - Added copywrite headers to everything

### DIFF
--- a/Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
+++ b/Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
@@ -12,39 +12,45 @@ using System.Collections.Generic;
 namespace Leap.Unity.Query {
 
   /// <summary>
-  /// The wrapper for all leap queries.  These queries are meant to mirror the queries present in the System.Linq.Enumerable
-  /// class.  These queries are meant to be functionaly identical, but allocate zero garbage, both during the generation of
-  /// the query, as well as the execution.  The speed is also aimed to be as fast or faster.
+  /// The wrapper for all leap queries.  These queries are meant to mirror the queries present in the 
+  /// System.Linq.Enumerable class.  These queries are meant to be functionaly identical, but allocate 
+  /// zero garbage, both during the generation of the query, as well as the execution.  The speed is 
+  /// also aimed to be as fast or faster.
   ///
-  /// There is some difference between using Linq and using Leap queries. One is that is you must prefix your query with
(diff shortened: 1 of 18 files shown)


Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
```diff
@@ -12,39 +12,45 @@
...
-  /// There is some difference between using Linq and using Leap queries. One is that is you must prefix your query with
-  /// a call to a Query() method if you are starting with an external data structure.  So for example if you want to query a
-  /// list, your method call would look something like this:
+  /// There is one big difference between using Linq and using Leap queries.You must prefix your query 
+  /// with a call to a Query() method if you are starting with an external data structure.  So for 
+  /// example if you want to query a list, your method call would look something like this:
   ///
   /// myList.Query().Where(someCondition).First();
-  ///
-  /// The other major difference is that you can only use a query once.  So the following will not work:
```
jselstad: Is this no longer the case?

Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
```diff
@@ -70,29 +76,76 @@ public struct Enumerator {
...
+  /// </summary>
   public interface IQueryOp<T> {
+
+    /// <summary>
+    /// Tries to get the next value in the sequence.  If this method returns true,
+    /// the next value will be placed into the out parameter t.  If this method
+    /// returns false, the sequence is at an end and t will be the default value
+    /// of T.
+    /// 
+    /// Once TryGetNext returns false, it can NEVER return true again until the
```
jselstad: Or has the warning been moved down here?

Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
Amarcolina: It is no longer the case!  You can now re-use queries as much as you want!

Assets/LeapMotion/Scripts/Query/QueryWrapper.cs
Amarcolina: This is slightly different.  In the general case, you will never interact with TryGetNext directly so you never need to worry about calling Reset().  Normally you will either use a foreach(), or a direct op like First() / ToList() / ect... all of which do not need to be manually reset.

Amarcolina:
I will add examples to each query op!

PR #720, opened 2017-05-12
Fixed stupid unity null weirdness

Amarcolina:
Since Unity objects can become null but not really be null, it can cause multiple values to be added to the hash set because actual null is considered differently than unity null.  Very annoying.

To test:
 - [ ] Create hash set of gameObjects.  Add a few gameObjects to the set.  Save scene.  Delete gameobjects.  Save scene.  Reload scene.  Visit hash set component, and verify that the number of empty elements is equal to the number of elements that were added in the first place!


nickjbenson:
Great! Look good

PR #721, opened 2017-05-13
Example Six

nickjbenson:
Featuresss:
- AnchorableBehaviour callbacks
- WorkstationBehaviourExample
- Text for Example 5
- Text and implementation for Example 6


Amarcolina:
Example 5 now has a missing prefab in it.

Amarcolina:
WorkstationPoseTest.cs also seems to be unused?  It also has an AddComponentMenu("") on it which means nobody would ever be able to add it to anything.

PR #722, opened 2017-05-16
Create world-space unwarping extension method, use for hand unwarping

nickjbenson:
Change to make the inverse transform operation we do to support curved spaces a little more readable. 
I'll be using this to make supporting controllers less of a hassle.

To test:
- [ ] Make sure it produces functionally identical results to the previous method

### DIFF
--- a/Assets/LeapMotion/Scripts/Space/ITransformer.cs
+++ b/Assets/LeapMotion/Scripts/Space/ITransformer.cs
@@ -102,14 +102,16 @@ namespace Leap.Unity.Space {
                                         Vector3 worldWarpedPosition, Quaternion worldWarpedRotation,
                                         out Vector3 worldRectilinearPosition, out Quaternion worldRectilinearRotation) {
 
-      Vector3 anchorLocalWarpedPosition = transformer.anchor.space.transform.InverseTransformPoint(worldWarpedPosition);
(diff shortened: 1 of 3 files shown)


Assets/LeapMotion/Scripts/Space/ITransformer.cs
```diff
@@ -89,4 +89,28 @@ public class IdentityTransformer : ITransformer {
...
+    /// Given a transformer and a world-space position and rotation, this method interprets that position
+    /// and rotation as being within the transformers "warped" space (e.g. cylindrical space for LeapCylindricalSpace)
+    /// and outputs the world-space position and rotation that would result if the space was no longer warped,
+    /// i.e., standard Unity rectilinear space.
+    /// </summary>
+    public static void WorldSpaceUnwarp(this ITransformer transformer,
+                                        Vector3 worldWarpedPosition, Quaternion worldWarpedRotation,
+                                        out Vector3 worldRectilinearPosition, out Quaternion worldRectilinearRotation) {
+
+      Vector3 anchorLocalWarpedPosition = transformer.anchor.space.transform.InverseTransformPoint(worldWarpedPosition);
```
Amarcolina: Only thing I would change is probably store the transform value in a variable to make each of these lines like half as long.

PR #725, opened 2017-05-19
Support Controllers in the Interaction Engine

nickjbenson:
This PR adds support for Oculus Touch in the Interaction Engine in its current state, with additional to-dos coming down the pipe before I'll request a review:

- [x] Fix bug where controller collider velocities where not being set
- [x] Test and add support for non-spherical colliders in soft contact
- [x] Investigate why relative velocity appears not to be used when throwing w/ backward motion
- [x] Allow choosing which fingertips to be primary hover points in InteractionHand inspector
- [x] Re-add non-index curl-based invalidation for primary hover (InteractionHand only)
- [x] Organize InteractionVRController inspector
- [x] Add custom, non-Unity-default controller support via optional script-only settings in InteractionVRController
- [x] Add Vive support for InteractionVRController
- [x] Fix bug when VR camera doesn't start at the origin (translated parent)
- [x] Ensure OnDisable() fires appropriate interaction-end callbacks for all InteractionControllers
- [x] Add InteractionVRController inspector workflow for adding and specifying custom interaction points and colliders
- [x] Change contact bone displacement fraction to be an absolute distance scaled by controller scale
- [x] Provide InteractionVRController inspector notifications and warnings if interaction point data is in a bad state or if overrideable, custom controller settings have been set or not set
- [x] Add functionality enabled-by-default that will disable an InteractionHand if it gets close to a VR controller
- [x] Verify examples scenes are correct and support controllers. (Example 6 chosen to remain hands-only for scene simplicity.)
- [x] Fix bug where grasp/contact/hover were still performing checks when the interaction was disabled
- [x] Make sure builds work on the relevant target platforms!


Assets/LeapMotion/Scripts/Query/DirectQueryOps.cs
```diff
@@ -361,5 +361,17 @@ public partial struct QueryWrapper<QueryType, QueryOp> where QueryOp : IQueryOp<
...
     }
+
+    public void FillHashSet(HashSet<QueryType> hashSet) {
+      hashSet.Clear();
+      AppendHashSet(hashSet);
+    }
+
+    public void AppendHashSet(HashSet<QueryType> hashSet) {
+      QueryType obj;
+      while (_op.TryGetNext(out obj)) {
```
Amarcolina: You need to make a copy of _op before you use it to allow Queries to be re-used.

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -430,6 +430,46 @@ public struct ChildrenEnumerator {
...
+    /// attached to its Rigidbody if it doesn't have one) and adds them to the provided
+    /// colliders list. Warning: The provided "colliders" List will be cleared before
+    /// use.
+    /// 
+    /// Colliders that are the children of other Rigidbody elements beneath the argument
+    /// object are ignored.
+    /// </summary>
+    public static void FindColliders<T>(GameObject obj, ref List<T> colliders) where T : Collider {
+      colliders.Clear();
+      Stack<Transform> toVisit = Pool<Stack<Transform>>.Spawn();
```
Amarcolina: A good best practice I have adopted is to do something like this when using Pool

```
var resources = Pool.Spawn();
try {
  // do stuff with resource
} finally {
  resource.Clear();
  Pool.Recycle(resource)
}
```

That way you can ensure that there is never a memory leak due to unexpected errors (or returns!) within the method.

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -430,6 +430,46 @@ public struct ChildrenEnumerator {
...
+    /// Recursively searches the hierarchy of the argument GameObject to find all of the
+    /// Colliders that are attached to the object's Rigidbody (or that _would_ be 
+    /// attached to its Rigidbody if it doesn't have one) and adds them to the provided
+    /// colliders list. Warning: The provided "colliders" List will be cleared before
+    /// use.
+    /// 
+    /// Colliders that are the children of other Rigidbody elements beneath the argument
+    /// object are ignored.
+    /// </summary>
+    public static void FindColliders<T>(GameObject obj, ref List<T> colliders) where T : Collider {
```
Amarcolina: Why is the ref keyword used here?

Assets/LeapMotion/Scripts/Utils/Utils.cs
```diff
@@ -430,6 +430,46 @@ public struct ChildrenEnumerator {
...
+          // Ignore children with Rigidbodies of their own; its own Rigidbody
+          // owns its own colliders and the colliders of its children
+          if (child.GetComponent<Rigidbody>() == null) {
+            toVisit.Push(child);
+          }
+        }
+
+        // Since we'll visit every child, all we need to do is add the colliders
+        // of every transform we visit.
+        foreach (var collider in curTransform.GetComponents<T>()) {
```
Amarcolina: If you are using the pool to get the stack without allocating, you would probably want to do something similar here to prevent allocation.  There is an overload of GetComponents where you pass in an array and it fills it with the results that doesn't allocate that you could use.

Assets/LeapMotionModules/Interaction/Scripts/Controllers/KabschGraspedPose.cs
```diff
@@ -23,122 +23,119 @@ public enum SolveMethod {
...
+        Vector3 manipulatorPosition = controller.graspManipulatorPoints[i];
 
-        for (int j = 0; j < NUM_BONES; j++) {
-          Bone.BoneType boneType = (Bone.BoneType)j;
-          Bone bone = finger.Bone(boneType);
+        newPoints.SetWorldPosition(i, manipulatorPosition);
+      }
 
-          Vector3 bonePos = bone.NextJoint.ToVector3();
+      // TODO: Delete me. Make sure InteractionHands provide correct manipulator positions
```
Amarcolina: Want to keep this in for this PR?

Assets/LeapMotionModules/Interaction/Scripts/InteractionBehaviour.cs
```diff
@@ -39,32 +43,74 @@ public class InteractionBehaviour : MonoBehaviour, IInteractionBehaviour {
...
+    public InteractionController primaryHoveringController {
+      get {
+        return _closestPrimaryHoveringController;
+      }
+    }
+
+    /// <summary>
+    /// Gets the set of all interaction controllers primarily hovering over this object.
+    /// </summary>
+    public HashSet<InteractionController> primaryHoveringControllers {
```
Amarcolina: Mean to use ReadonlyHashSet here?

Assets/LeapMotionModules/Interaction/Scripts/InteractionBehaviour.cs
```diff
@@ -39,32 +43,74 @@ public class InteractionBehaviour : MonoBehaviour, IInteractionBehaviour {
...
       }
     }
 
     /// <summary>
-    /// Gets whether this object is the primary hover for any Interaction Hand.
+    /// Gets all of the interaction controllers hovering near this object, whether they
+    /// are Leap hands or supported VR controllers.
     /// </summary>
-    public bool isPrimaryHovered { get { return _primaryHoveringHands.Count > 0; } }
+    public HashSet<InteractionController> hoveringControllers {
```
Amarcolina: ReadonlyHashSet here?

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+  public enum IgnoreHoverMode { None, Left, Right, Both }
+
+  /// <summary>
+  /// The Interaction Engine can be controlled by hands tracked by the
+  /// Leap Motion Controller, or by remote-style held controllers
+  /// such as the Oculus Touch or Vive controller.
+  /// </summary>
+  public enum ControllerType { Hand, VRController }
+
+  public abstract class InteractionController : MonoBehaviour {
```
Amarcolina: [DisallowMultipleComponent](https://docs.unity3d.com/ScriptReference/DisallowMultipleComponent.html) would probably be great for this class.

Assets/LeapMotionModules/Interaction/Scripts/InteractionBehaviour.cs
nickjbenson: good idea

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: To require explicit acknowledgement from the caller of the method that the colliders List will be modified -- and more specifically, cleared. If that's obvious enough to you I can get rid of it.

Assets/LeapMotion/Scripts/Utils/Utils.cs
nickjbenson: Aha! Clever, I will do this

Assets/LeapMotion/Scripts/Utils/Utils.cs
Amarcolina: I think that it should be obvious enough from the docs of the method.  The ref keyword also might be misleading for some, implying that you might allocate a new list inside the method and assign it to be the new colliders list.

jselstad: Seems like it's used to denote that "this list will be modified", ie filled with Colliders

Amarcolina:
When using internal private caches I have started naming the backing variable differently in order to make it less likely that it is used accidentally.  So instead of

```
private var _backing;
private var backing {get;}
```

I would recommend

```
private var _cachedBacking;
private var backing {get;}
```

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+
+      // releasedObject will be null
+      return false;
+    }
+
+    /// <summary>
+    /// Attempts to release this hand's object, but only if the argument object is the object currently
+    /// grasped by this hand. If the hand was holding the argument object, returns true, otherwise returns false.
+    /// </summary>
+    public bool ReleaseObject(IInteractionBehaviour toRelease) {
```
Amarcolina: Probably should put a null check here.  ReleaseObject(null) will return true.

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+      }
+    }
+
+    #region Grasp State Checking
+
+    /// <summary>
+    /// Called by the Interaction Manager every fixed frame.
+    /// Returns true if the hand just released an object and outputs the released object into releasedObject.
+    /// </summary>
+    public bool CheckGraspEnd(out IInteractionBehaviour releasedObject) {
```
Amarcolina: These methods are prefixed with Check, which implies that it is only a condition checl, but it does look like they do change the state of things internally.  Maybe could be renamed with a Try prefix or similar?

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+      if (s_defaultContactBoneMaterial == null) {
+        s_defaultContactBoneMaterial = new PhysicMaterial();
+      }
+      s_defaultContactBoneMaterial.hideFlags = HideFlags.HideAndDontSave;
+      s_defaultContactBoneMaterial.bounceCombine = PhysicMaterialCombine.Minimum;
+      s_defaultContactBoneMaterial.bounciness = 0F;
+    }
+
+    private bool _contactInitialized = false;
+    protected abstract ContactBone[] contactBones { get; }
```
Amarcolina: Anything abstract should be documented so that a user knows what they need to do in order to implement it.

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
jselstad: Hmm, these functions have mirrors in the scripting API (ie Physics.CheckSphere: https://docs.unity3d.com/560/Documentation/ScriptReference/Physics.CheckSphere.html ).

Canonically, they only return bools, but I like the extra telemetry...

If only there were a way to make using the out arguments optional...  is there?

nickjbenson: TryGraspEnd would be really really bad. That would encourage people to call it themselves to "try ending a grasp." The commonality between all of these methods is that they perform _the_ checks that result in state-change callbacks and modify appropriate internal state, hence the Check prefix, and also "Checking grasp end" doesn't sound very useful to someone poking around aimlessly, so they won't mess with it. I'm going to need more help coming up with an alternative if Check isn't passing muster.

FWIW I'd also like to make these methods lowerCamelCase even though they're public to make them stand out even more as methods not-to-be-messed-with.

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+      }
+
+      // Clear contact data if we lose tracking.
+      if (!isTracked && _contactBehaviours.Count > 0) {
+        _contactBehaviours.Clear();
+      }
+
+      // Disable contact bone parent if we lose tracking.
+      if (!isTracked) {
+        contactBoneParent.gameObject.SetActive(false);
```
Amarcolina: contactBoneParent is already a gameObject, so no need to call .gameObject

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
```diff
@@ -0,0 +1,1382 @@
...
+          contactBoneParent.gameObject.SetActive(true);
+        }
+      }
+      
+      // Request and store target bone positions and rotations
+      // for use during the contact update.
+      for (int i = 0; i < contactBones.Length; i++) {
+        Vector3 targetPosition;
+        Quaternion targetRotation;
+        getColliderBoneTargetPositionRotation(i, out targetPosition, out targetRotation);
```
Amarcolina: You can actually out directly into the array element :D

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
Amarcolina: Ahhh yeah that makes perfect sense.  Maybe we could do something similar to what happened with the editor functions in the graphic renderer, make an additional class for 'internal' functions.  So the interaction manager would do things like:

```
controller.internal.CheckEndingGrasp
```

Since it seems that the user should definitely never be calling these methods themselves right?

Assets/LeapMotion/Scripts/Utils/Utils.cs
Amarcolina: Yeah but the ref keyword isn't required to do that, so I think it could be confusing :P Unity has lot's of cases in its interfaces where you pass in a list (without ref) to be filled, so I don't think it's unreasonable to use that pattern.

Assets/LeapMotionModules/Interaction/Scripts/TestOculusControllerScript.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Leap.Unity.RuntimeGizmos;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.VR;
+
+public class TestOculusControllerScript : MonoBehaviour, IRuntimeGizmoComponent {
```
jselstad: I'm assuming this script is vestigial?

ProjectSettings/InputManager.asset
```diff
@@ -6,36 +6,36 @@ InputManager:
   serializedVersion: 2
   m_Axes:
   - serializedVersion: 3
-    m_Name: Horizontal
+    m_Name: LeftVRGripAxis
```
jselstad: Are our controllers reliant on these input additions?

Or do they use the direct ButtonID inputs?
https://docs.unity3d.com/Manual/OculusControllers.html

Assets/LeapMotionModules/Interaction/Scripts/InteractionHand.cs
```diff
@@ -8,947 +8,502 @@
...
-        if (_hand != null) {
-          _warpedHandData.CopyFrom(_handData);
-          if (_hoverResults.primaryHovered != null && (space = _hoverResults.primaryHovered.space) != null) {
-            //Transform bulk hand to the closest element's warped space
-            inverseTransformHand(_warpedHandData, space);
+        if (handAccessorFunc == null) {
+          if (handDataMode == HandDataMode.PlayerLeft) {
+            handAccessorFunc = (frame) => frame.Hands.Query()
+                                                     .Where(hand => hand.IsLeft)
+                                                     .FirstOrDefault();
```
Amarcolina: just a small note that you can pass predicates directly to First (and many direct ops) instead of prefixing them with a where :)

Assets/LeapMotionModules/Interaction/Scripts/IVRControllerTrackingProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+  /// 
+  /// For a reference implementation, refer to DefaultVRNodeTrackingProvider.
+  /// </summary>
+  public interface IVRControllerTrackingProvider {
+
+    /// <summary>
+    /// Returns whether or not this provider is currently tracking the controller for
+    /// which it provides data.
+    /// </summary>
+    bool GetIsTracked();
```
Amarcolina: How about an IsTracked getter instead of a method?

Assets/LeapMotionModules/Interaction/Scripts/IVRControllerTrackingProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// Sets up a subscription so that the provided Action is called whenever new
+    /// tracking data is available for this controller.
+    /// 
+    /// It is recommended that tracking providers update their controllers' tracking
+    /// data before the Interaction Manager runs every FixedUpdate to minimize latency.
+    /// </summary>
+    void Subscribe(Action<Vector3, Quaternion> onTrackingDataUpdate);
```
Amarcolina: And how about just a simple

```
event Action<Vector3, Quaternion> OnTrackingDataUpdate
```

That way you it is simpler to implement the interface?

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
nickjbenson: Ohhhh! That's a really good idea. I'll look into doing that

Amarcolina:
The interaction manager is in different places for different examples, probably would be a good idea to always have in in the same (recommended) place.

Assets/LeapMotionModules/Interaction/Scripts/InteractionController.cs
nickjbenson: fixed

PR #726, opened 2017-05-22
Fix android

protodeep:
- Moved wrist positions in LoPoly hand models to more closely match runtime joint positions
- Rebuilt LoPoly hand prefabs & Updated existing scenes
- fix: Calling setDeformPositionsInFingers() on Init() in RiggedHand to fix need to cycle DeformPositions
- LoPoly hands models, prefabs and example scenes moved into Core to support mobile development without need of another module.

- [ ] Run /LeapMotion/Scenes/Rigged_Hands_VR.unity and verify that models are working properly.  
Optional: Add CapsuleHand's to the scene to verify correct runtime joint positions.


jselstad:
Looks good to me 👍

nickjbenson:
OK!

PR #743, opened 2017-06-16
Removed unnecessary Debug.Log statements

julienkay:
Just removed some Debug.Log() statements from SkeletalFinger and SkeletalHand. They were unnecessarily logging each frame, which was bad for performance.


nickjbenson:
Assigning Barrett -- I believe this is yours? If not feel free to pass it to @Amarcolina

PR #744, opened 2017-06-17
Fix the bug for HandPool.AddNewGroup()

daleth90:
After using HandPool.AddNewGroup() at runtime, it would throw NullReferenceException when TryGetModel() being called. Defining the lists may fix, but still need to initialize the new groups for working properly.


nickjbenson:
Thanks for the suggestion @daleth90! Passing this to @Amarcolina.

Amarcolina:
Nice catch and great fix!  Gonna merge this in.

PR #745, opened 2017-06-19
Make buttons less molasses at low fixed timesteps

nickjbenson:
Swapped the exponential decay in InteractionButton for a friction-and-drag model, and tuned the coefficients to produce reasonable results at physics timesteps of 1/30, 1/45, 1/60, and 1/90 (seconds). 

Note that at a timestep of 1/30 (and to a lesser extent 1/45), buttons are still noticeably slower than their 1/60 & 1/90 counterparts. (Making the button motion truly timestep-independent is nontrivial and honestly unnecessary). If it's important to be able to make the buttons faster in lower-frequency-timestep situations, we should expose button friction & drag in the inspector for user tuning. This could be done in this hotfix but should probably be done as a v1.1 release feature.

Qs for John:
- [ ] Does the model make sense?
- [ ] Should we expose friction and drag coefficients in the inspector?

Qs for QA:
- [ ] Do buttons behave acceptably on Android now given the 1/60 timestep?
- [ ] Do buttons and sliders still behave acceptably on all UI example scenes?

### DIFF
--- a/Assets/LeapMotion/Modules/InteractionEngine/Examples/2. Basic UI/Basic UI.unity
+++ b/Assets/LeapMotion/Modules/InteractionEngine/Examples/2. Basic UI/Basic UI.unity
@@ -191,7 +191,7 @@ Rigidbody:
   m_GameObject: {fileID: 112263976}
   serializedVersion: 2
   m_Mass: 1
-  m_Drag: 40
+  m_Drag: 0
   m_AngularDrag: 0.05
   m_UseGravity: 0
   m_IsKinematic: 0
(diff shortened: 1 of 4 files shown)


testmasterbright:
@nickjbenson I second the above regarding a 1.1.0 release. What prompted this fix was an Android issue. The Android package would be going to only select OEMs so honestly, I think the above changes should be held for a future release and we cut a special build for OEMs on this occasion only. If we go this route we would need to cut the build for OEMs and then re-target this PR for develop. Thoughts?

nickjbenson:
Closed in favor of #747.

PR #748, opened 2017-06-19
Feature: Edit-time Poses

nickjbenson:
This feature addition:

- Restructures TestHandFactory to take arbitrary LeapTransforms to apply to hands after constructing them
- Provides two alternative default poses specifyable by the enum value, `TestHandPose`, PoseA (the original pose), and PoseB (a palms-in, further-out pose more suitable for building hand UIs)
- Adds an editTimePose public field to the inspector for all LeapProviders
- Modifies IHandModel to use the LeapServiceProvider's editTimePose to construct its edit-time pose

To test:
- [x] Open any Leap-enabled scene and change the editTimePose field of the LeapServiceProvider.
- [x] Verify nothing is borked!


Amarcolina:
Looks like attachment hands are stuck using only a left hand?

Issue #652, opened 2017-04-08
[Blocking Issue] Remove Interaction layer, remember existing layer instead


nickjbenson:
Note: This issue blocks the resolution of #740.

Interaction objects should not need to be on a special "Interaction" layer because there are frequent cases in which, for example, for UI, objects should be placed on a non-self-colliding layer (such as a "UI" layer). 

One option to solve the use-case is to use another Interaction-Non-Self-Colliding layer, but this isn't great because in general we should be minimizing layer use (assigning layers dynamically not well supported in Unity anyway).

- An important optimization will be to remember which layers contain Interaction objects upon manager registration, so we only have to provide those layer flags to Physics queries.
- Edit-time pop-up in behaviour's inspector: when manager.contactEnable is true, a behaviour's ignoreContact is set to false, and the behaviour's layer isn't set to collide with the Interaction Hands layer, there should be a warning visible along with a "Fix Now" button.
- Additionally, the Interaction Manager should raise an error when there aren't enough layers to perform layer auto-generation.

nickjbenson:
This fix does not address the grasped layer, which still has to be a special-snowflake layer. The solution for providing developers flexibility here is to allow specifying a target layer to use as the Grasped (InteractionNoClip) layer on a per-object basis (or in an 'InteractionMaterial').

nickjbenson:
Addressed by #752

PR #752, opened 2017-06-20
Add layer overrides feature

nickjbenson:
This PR:

- Addresses #652 -- the Interaction layer still exists, but can be overridden on a per-object basis. The Interaction-no-contact layer can also optionally be overriden on a per-object basis.
- Modifies example scenes to make use of layer overrides for UI objects
- Adds an inspector warning when no manager is configured on an interaction object (without a configured manager, we can't do any validation!)
- Provides inspector warnings / errors when there are layer configuration issues -- specifically when any object's (or the manager's) No Contact layer isn't configured properly relative to the contact bone layer
- Modifies example scenes with a helper script to prevent UI self-collision on Start()
- Addresses #755 , button-grasping instability via InteractionButton.cs by:
  - Adding **proper** support for primary hover locking (doesn't also lock hover state)
  - Adding a 10% hysteresis to Interaction Buttons so they don't have rapidly oscillating edge states

To test:
- [x] Run through the example scenes containing UI objects (2, 4, 6) and ensure buttons and sliders work as expected
- [x] Mess with layers and make sure any errors make sense in the inspector


Assets/LeapMotion/Modules/InteractionEngine/Examples/Common Example Assets/Scripts/IgnoreCollisionsInChildren.cs
```diff
@@ -0,0 +1,45 @@
...
+  /// This utility script scans through its Transform's children and tells PhysX to
+  /// ignore collisions between all pairs of Colliders it finds. This is particlarly
+  /// useful, for example, for interfaces, where buttons shouldn't collide with other
+  /// buttons.
+  /// 
+  /// This is not the recommended strategy in general: It is much more optimal for your
+  /// application to put interface objects on a layer and disable self-collision for that
+  /// layer in your Physics settings (Edit / Project Settings / Physics).
+  /// </summary>
+  [AddComponentMenu("")]
```
Amarcolina: Hmm I'm not sure we should go so far as preventing people from adding this script to their gameObjects.  I think it would be very useful to have something like this for prototyping and development before one goes all the way to create multiple layers.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/Editor/InteractionBehaviourEditor.cs
```diff
@@ -32,6 +35,105 @@ public class InteractionBehaviourEditor : CustomEditorBase<InteractionBehaviour>
...
+      // Layer Overrides
+      specifyConditionalDrawing(() => target.overrideInteractionLayer,
+                                "_interactionLayer");
+      specifyConditionalDrawing(() => target.overrideNoContactLayer,
+                                "_noContactLayer");
+      specifyCustomDecorator("_noContactLayer", drawNoContactLayerDecorator);
+    }
+
+    private void drawInteractionManagerDecorator(SerializedProperty property) {
+      if (PrefabUtility.GetPrefabType(target) == PrefabType.Prefab) {
```
Amarcolina: This editor specifies that it is allowed to edit more than one object at a time, and so you should refrain from using the 'target' keyword, since it will only allow you to make decisions based on the state of a single object.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/Editor/InteractionBehaviourEditor.cs
```diff
@@ -32,6 +35,105 @@ public class InteractionBehaviourEditor : CustomEditorBase<InteractionBehaviour>
...
+        return;
+      }
+
+      bool shouldDrawInteractionManagerNotSetWarning = false;
+      foreach (var target in targets) {
```
Amarcolina: In general, if dealing with properties that are serialized, (like manager is I believe?) we should always favor using serialized properties over using the properties directly.  We handle multiple edit targets more easily, as well as better support for undo/edit detection/on validate ect...

Amarcolina:
Auto-Fix button seems to get very tall under certain situations, but I'm not sure why...

Amarcolina:
It also seems really strange that there is an Auto-Fix button, but it can fail if there is no interaction manager in the scene.  I would instead opt to remove the auto-fix button if the manager cannot be found, but still pop the error in the inspector so they still know something is wrong.

nickjbenson:
I was concerned about repeated calls to InteractionManager.instance every editor update when there's no manager in the scene, because that means calling FindObjectOfType a lot. Maybe it's fine?

Amarcolina:
Got this stack trace when messing around in the editor

ArgumentNullException: Argument cannot be null.
Parameter name: key
System.Collections.Generic.Dictionary`2[UnityEngine.Rigidbody,Leap.Unity.Interaction.IInteractionBehaviour].set_Item (UnityEngine.Rigidbody key, IInteractionBehaviour value) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/Dictionary.cs:155)
Leap.Unity.Interaction.InteractionManager.RegisterInteractionBehaviour (IInteractionBehaviour interactionObj) (at Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionManager.cs:725)
Leap.Unity.Interaction.InteractionBehaviour.set_manager (Leap.Unity.Interaction.InteractionManager value) (at Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs:538)
Leap.Unity.Interaction.InteractionBehaviourEditor.drawSetManagerWarningBox (System.String warningMessage, MessageType messageType) (at Assets/LeapMotion/Modules/InteractionEngine/Scripts/Editor/InteractionBehaviourEditor.cs:95)
Leap.Unity.Interaction.InteractionBehaviourEditor.drawInteractionManagerDecorator (UnityEditor.SerializedProperty property) (at Assets/LeapMotion/Modules/InteractionEngine/Scripts/Editor/InteractionBehaviourEditor.cs:73)
Leap.Unity.CustomEditorBase.drawProperty (UnityEditor.SerializedProperty property) (at Assets/LeapMotion/Core/Editor/CustomEditorBase.cs:246)
Leap.Unity.CustomEditorBase.OnInspectorGUI () (at Assets/LeapMotion/Core/Editor/CustomEditorBase.cs:216)
Leap.Unity.Interaction.InteractionBehaviourEditor.OnInspectorGUI () (at Assets/LeapMotion/Modules/InteractionEngine/Scripts/Editor/InteractionBehaviourEditor.cs:142)
UnityEditor.InspectorWindow.DrawEditor (UnityEditor.Editor editor, Int32 editorIndex, Boolean rebuildOptimizedGUIBlock, System.Boolean& showImportedObjectBarNext, UnityEngine.Rect& importedObjectBarRect) (at C:/buildslave/unity/build/Editor/Mono/Inspector/InspectorWindow.cs:1229)
UnityEditor.DockArea:OnGUI()

Amarcolina:
FindObjectsOfType should be fine to do in the editor, especially since graphics are not repainted constantly.

nickjbenson:
Good catch re: stack trace, fiz shouldn't be too bad

Amarcolina:
You might need to add some editor code to prevent people from changing the layer overrides at runtime.  Or to correctly notify the manager that the layer has changed so that it can properly update it's layer mask.  Otherwise it might not correctly have a mask that covers all objects.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -830,13 +854,93 @@ public class InteractionManager : MonoBehaviour, IRuntimeGizmoComponent {
...
+      if (refreshImmediately) {
+        refreshInteractionLayerMask();
+      }
+    }
+
+    void IInternalInteractionManager.NotifyIntObjRemovedInteractionLayer(IInteractionBehaviour intObj, int layer, bool refreshImmediately) {
+      _intObjInteractionLayers[layer].Remove(intObj);
+
+      if (_intObjInteractionLayers[layer].Count == 0) {
+        _intObjInteractionLayers.Remove(layer);
```
Amarcolina: Instead of removing the hash set, why not leave it?  You can always check if it is empty in the refresh method for very little extra cost, and that way you are able to avoid the case where a specific object with a unique override layer is being enabled/disabled a whole bunch, causing a new hash set to be constructed every time it is enabled.

It's also better to keep hash sets around.  As objects are added to them, they need to periodically resize, but they keep their size once they have expanded.  If you delete the hash set when it becomes empty, it needs to go through that resizing process again, resulting in extra allocations and time.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -830,13 +854,93 @@ public class InteractionManager : MonoBehaviour, IRuntimeGizmoComponent {
...
         Physics.IgnoreLayerCollision(_interactionLayer, i, shouldIgnore);
         Physics.IgnoreLayerCollision(_interactionNoContactLayer, i, shouldIgnore);
-
-        // Set brush layer to collide with nothing
-        Physics.IgnoreLayerCollision(_contactBoneLayer, i, true);
+        Physics.IgnoreLayerCollision(_contactBoneLayer, i, shouldIgnore);
       }
 
-      //After copy and set we enable the interaction between the brushes and interaction objects
+      // Enable interactions between the contact bones and the interaction layer.
```
Amarcolina: Part of automatic generation is the guarantee that the contact bone layer collides with the interaction layer.  We should provide a warning in the same way we do with the no-contact/bone relationship.  If the configured contact bone layer does not collide with the interaction layer, we should pop a warning in the inspector.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/Internal/HeuristicGrabClassifier.cs
```diff
@@ -72,6 +70,9 @@ public class HeuristicGrabClassifier {
                                             * interactionHand.manager.SimulationScale;
           _scaledGrabParams.MAXIMUM_DISTANCE_FROM_HAND = _defaultGrabParams.MAXIMUM_DISTANCE_FROM_HAND
                                                        * interactionHand.manager.SimulationScale;
+
+          // Ensure layer mask is up-to-date.
+          _scaledGrabParams.LAYER_MASK = interactionHand.manager.GetInteractionLayerMask();
       
           // Ensure that the temporally variant variables are updated.
           // scaledGrabParams.LAYER_MASK = 1 << _manager.InteractionLayer;
```
Amarcolina: Looks like this commented line should be removed

Amarcolina:
Not sure why, but soft contact is getting triggered *waaayyy* too easily when I put an interaction object on a different interaction layer.  Here are my layers in full:

Interaction
Interaction_NoContact
Magic
Magic_NoContact
ContactBone

Interaction and Magic both collide with ContactBone
Interaction_NoContact and Magic_NoContact do not collide with ContactBone
Neither Magic nor Magic_NoContact collides with either Interaction or Interaction_NoContact

I can barely push around magic objects because soft contact is triggering so quickly.

nickjbenson:
Working on adding: 
- Show a warning if the contact bone layer does not collide with an InteractionBehaviour's interaction layer
- Show a warning if the contact bone layer in the Manager does not collide with the interaction layer (during manual layer setup).
- Show an error if a contact bone ever collides with something that is not an InteractionBehaviour

PR #756, opened 2017-06-22
Added a checkbox to manually force custom updates

Amarcolina:
Adds a checkbox that manually forces the temporal warping script to trigger manual updates in Late Update.  This is to try to handle the general case when someone is using a vr tracking solution that is not the integrated unity VR solution.  They should be able to check the checkbox, and as long as they update the transform of the head before late update, everything should be good.


nickjbenson:
@protodeep Assigning you because you wanted to verify this PR actually fixed things.
@testmasterbright When Barrett is finished this can get QA'd if you'd like but I'd recommend we just merge it to `develop` and it'll go through normal QA when we approach release to `master`

PR #758, opened 2017-06-22
Upgrade to 2017.1.0f3

Amarcolina:
Only a few changes here since it is a minor upgrade.  I deleted the lighting data assets for the UI Input Module since it is no longer a supported module.  The idea is that deleting the data allows us to switch platforms more quickly (less textures to re-import) and prevents us from getting import warnings about outdated formats.  I also updated the lighting for the VR Visualizer to prevent out of date warnings.


nickjbenson:
Hmm, heads-up that on the particles project I get Oculus and SteamVR headset instability (something like half framerate, which causes flickering on Oculus and a blank screen on SteamVR). Opening that project in 5.6.1f1 instead prevents any issues. We should double-confirm we don't have those issues generally before merging this.

nickjbenson:
I've been testing PRs in 2017.1.0 for a while now so this upgrade it satisfactory to me; it doesn't have the problems I had with Oculus in 5.6.2.

@protodeep Would you like to give this another round of testing before we approve and merge?

PR #761, opened 2017-06-26
Fix bug where rebuild of atlas was not handled

### LINKED ISSUE #760: Text renderer needs to be aware of atlas rebuilding
Need to use https://docs.unity3d.com/ScriptReference/Font-textureRebuilt.html event to get a callback whenever the texture atlas for text is rebuilt.  That way all graphics can be marked as dirty and have their meshes regenerated.  Currently this results in texture corruption.

Amarcolina:
Fixed #760

Now, whenever a rebuild of a font happens due to a change in a text graphic, it will mark _all_ graphics as dirty and rebuild them all.  This prevents the weird corruption that can happen as you change text at runtime.

To test:
 - [ ] Make sure that text still works
 - [ ] Verify that adding a whole bunch of new text at runtime doesn't break existing text.  Helps to have a lot of different sizes to trigger a rebuild.


nickjbenson:
Looks like it works!

PR #764, opened 2017-06-27
changed expected deviceString in Oculus Interaction Controller prefabs

julienkay:
The device strings that Unity's Input.GetJoystickNames() reports for the Oculus Touch Controllers for me are "oculus touch controller - left" and "oculus touch controller - right"

I'm on Unity 5.6.1f1 with Oculus 1.16. If it worked for you before, I suppose this is a change that was introduced by an update to either Unity/OVRPlugin or the Oculus Runtime.
You should probably still confirm that different machines/configurations don't report different strings, although that would be kind of odd.

### DIFF
--- a/Assets/LeapMotion/Modules/InteractionEngine/Prefabs/Interaction Controllers/VR Oculus-style Controller (Left).prefab
+++ b/Assets/LeapMotion/Modules/InteractionEngine/Prefabs/Interaction Controllers/VR Oculus-style Controller (Left).prefab
@@ -443,7 +443,7 @@ MonoBehaviour:
   _contactEnabled: 1
   _graspingEnabled: 1
   _trackingProviderType: Leap.Unity.Interaction.DefaultVRNodeTrackingProvider
-  _deviceString: oculus touch left
+  _deviceString: oculus touch controller - left
   _chirality: 0
   _hoverPoint: {fileID: 4560782703005532}
   primaryHoverPoints:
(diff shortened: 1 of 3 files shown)


nickjbenson:
Did you confirm the device strings do not work for you? We match device strings based on string-separated tokens, in this case, "oculus", "touch", and "left" or "right". We don't match the strings by literally matching the strings' contents. So the device string should be functioning correctly as-is (but please let me know if it's not actually working for you).

I created an issue to address what I believe would have been the cause of the discrepancy: #765

julienkay:
Ah, I see, I didn't catch that. I just confirmed that the code works as expected.

The real (minor) problem seems to be another one: If I do a clean start of Unity and run one of the example scenes, Input.GetJoystickNames() returns an empty string array. When i stop Play mode and start again, it returns the two touch controller strings and the InteractionVRController sets everything up as expected. That's why I thought my changes fixed the issue. I guess thats more of a Unity bug then.

Amarcolina:
Yeah julienkay, I always encounter the same Unity bug that you do, it's fairly annoying :/

nickjbenson:
Yeah.. I suspect it has to do with the fact that, before the first play, Unity hasn't requested joystick information from the relevant VR service, and if there's any sort of delay then you lose when play mode first starts. It's probably not an issue in builds because the user would already have active controllers when launching the application.

PR #771, opened 2017-07-05
Add Example 7: Moving reference frames

nickjbenson:
This PR:

- Adds Example 7, featuring a simple spaceship model and an explanation of how to work the example and what has to happen for moving reference frames to work.
- Adds scripts to support Example 7 with documentation to aid developers who dive into them.

To test:

- [ ] Make sure the button panel in Example 7 works consistently well even when the ship is moving and rotating quickly.
- [ ] The example-specific scripts should make sense.
- [ ] Example 7 should build and run without error (on PC; Android will be part of 1.1 QA).


Amarcolina:
Rigidbodies that are kinematic do not need their velocity or angular velocity set, so we should not set them.

Amarcolina:
also since setting the _transform_ instead of the rigidbody is important, we should emphasize that with some comments pointing it out.

nickjbenson:
OK, this branch is ready for testing @Amarcolina

Amarcolina:
I give this PR a big 👍

PR #776, opened 2017-07-12
Fix Visual Studio compilation!

nickjbenson:
This PR:

- Modifies the .gitignore to ignore .exe and _Data/ files so we don't have to delete test builds before committing
- **Fixes the VS build** by no longer _inexplicably_ targeting .NET 4.0 (it is still unclear why this happens)

To Test:

- [x] Pick some of our example scenes and verify they still work as expected, since this PR modifies LeapC.dll and LeapCSharp.dll
- [x] Try building the project in Visual Studio, look in awe and wonder as it compiles successfully
- [x] Try Attaching to Unity and getting play mode to stop at a Breakpoint (!!)
- [x] Profit!!!

We're on a roll here. Coming soon maybe: Supporting dynamic recompilation??!?? one can dream


Amarcolina:
Conflict here D: I touched the test files, I'll fix this...

PR #783, opened 2017-07-14
Sync Particles UnityModules-ward

nickjbenson:
Technically, this PR contains some duplicate fixes with the Various IE Fixes PR, so there'll probably be conflicts, or maybe we'll just get lucky. We should merge Various IE Fixes first, once it's resolved.

I just don't want the work here to get lost, so pardon me while I store this PR here for a little bit.


Assets/LeapMotion/Core/Scripts/DataStructures/Pool.cs
```diff
@@ -85,7 +83,29 @@ public static class Pool<T> where T : new() {
...
+      do {
+        if (_pool.Count > 0) {
+          value = _pool.Pop();
+        }
+        else {
+          value = new T();
+        }
+        count++;
+      } while (value == null && count < 100);
+      if (count == 100) {
```
Amarcolina: What's going on here with this counter?

nickjbenson: I ran into an issue where, because I had mishandled pooled objects, I had accidentally added null object to a Pool for those objects. The idea was to make sure we _really_ construct a new object and lazily clean up the pool if it contains null objects, and the counter prevents Unity from hard-locking if we somehow are unable to construct a non-null object (which could be strictly impossible?)

Amarcolina: It sounds like we should add null-checks to the recycle code then.  Wouldn't that solve the need to use a counter?

nickjbenson: Unless new T() returns null. Again, unless that's a strict impossibility, I'm concerned about having any code anywhere that could hard-lock the editor, even in the wildest scenarios... I won't push too hard for it though. I can remove the counter if you'd prefer to do the check in Recycle

Amarcolina: No new T() can never return null.  And we should definitely put null checking code into Recycle, since returning null should never be allowed.

nickjbenson: fixed now

Assets/LeapMotion/Core/Scripts/DataStructures/Pool.cs
```diff
@@ -72,11 +74,7 @@ public static class Pool<T> where T : new() {
       if (_pool == null) _pool = new Stack<T>();
 
       T value;
-      if (_pool.Count > 0) {
-        value = _pool.Pop();
-      } else {
-        value = new T();
-      }
+      value = spawnValue();
```
Amarcolina: Why you remove pooling logic? :(

nickjbenson: F.

because I'm a derp that's why. spawnValue() should be removed and this should be replaced with the actual pooling logic

PR #786, opened 2017-07-19
Adding Either and Maybe

Amarcolina:
Adds two base data structures, the Either and the Maybe.  These are concepts that exist in other languages that I have brought over for convenience/safety.  The Either represents a value that might be of one type or another, and provides methods to allow you to access them safely.  The Maybe represents a value that might or might not exist, and provides methods to allow you to access it safely.


Amarcolina:
Going to close this for now

Issue #785, opened 2017-07-19
Virtual Hand Position is Offset


clkruse:
I've been noticing that the position of the hand that is set by default in core assets is not correct. It feels like the virtual hand is higher and closer than the physical hand.

It seems like setting the focalPlaneOffset 0.07f to 0.1f, and changing the X-rotation under LMHeadMountedRig->CenterEyeAnchor->LeapSpace->LeapHandController from -90.02 to -95 helps. I don't think this is the perfect setup, but it's certainly better than the current default.

Issue #780, opened 2017-07-13
Playback Support for IE by using LeapProvider rather than LeapServiceProvider


julienkay:
Right now, Modules (i.e. the InteractionEngine) work with references to a LeapServiceProvider. I wanted to make Playback work with the InteractionEngine. So I wanted Hands using a PlaybackProvider to be able to interact with InteractionBehaviours. For this to work, I had to make a few minor modifications: I replaced the references to LeapServiceProvider in the static Hands class and the InteractionHand class with LeapProvider.

Conceptually this would make sense to me. The Interaction Engine would be independent from the implementation of the provider it gets its frames from. Be it a device, a recorded frame, frames sent through a network or whatever. I guess that abstraction is there for this reason anyway?

Now of course I don't know if there isn't another reason things have been done that way. For me, the example scenes seem to be working fine with my changes, both with recorded frames and normal use. What do you think?

nickjbenson:
I took a look at InteractionHand.cs in the latest public release and it doesn't actually require a LeapServiceProvider at all, which was by design, to address exactly this sort of situation.

Actually, if you look for a `provider` property, you should see that InteractionHands can accept a custom LeapProvider to override its default provider; this should allow you to drive a set of InteractionHands by attaching them to the output of a PlaybackProvider in an Awake() method in a custom set-up script for the InteractionHands.

Give that a shot and see if it addresses your use-case. I'll qualify that by saying PlaybackProvider + InteractionHands is a totally untested combo! :D

As for Hands.Provider, I've also made that change before in internal projects. It's worth thinking about. @Amarcolina for thoughts.

julienkay:
Yeah, you're right about InteractionHand.cs. I hought I had changed the Provider there too, but that's not the case.

I see now, that I could use the Custom HandDataMode to set my own provider. The only isue with this is however, that the InteractionControllerSwitcher.cs in line 128 still calls Hands.Left and Hands.Right every frame. This accesses the Provider property of the static Hands class (not InteractionHand), which is still a LeapServiceProvider. Accessing the property in turn does an expensive FindObjectOfType() call every frame. (Because I only have a PlaybackProvider set up in my scene so it never finds a LeapServiceProvider). Therefore OnLeftHandActive/OnLeftHandActive are never invoked. and hands not displayed.

[...]

Issue #206, opened 2016-05-07
LeapServiceProvider IsHeadMounted auto-enable when in VR


Amarcolina:
There have been a lot of issues with people forgetting to check the 'isHeadMounted' button on the service provider.  We should auto-detect whether or not VR mode is enabled and if the controller is a child of a VR camera and auto-check the box to prevent error.

Issue #160, opened 2016-04-22
Fix incorrect logic for device type


Amarcolina:
In LeapServiceProvider we are still using the incorrect serial number logic to determine device type, this should be fixed because it forces developers with devices that do not follow the pattern to check the 'override device type' checkbox, which shouldn't need to exist at all.

nickjbenson:
@jselstad @Amarcolina We should just remove the device type entirely for the next Core release, unless someone has a compelling argument for keeping it?

Amarcolina:
We need to first verify that we have a good alternative.  We originally created it because the leap API didn't provide enough data to make all of our decisions.  If we can get all the data we need from the API then we can get rid of this.

PR #793, opened 2017-07-21
Core DevX: Capsule hand, focalPlaneOffset, isHeadMounted, overrideDeviceType

nickjbenson:
Addresses #780, #160, #785, and #206

This PR:

- [x] Improves the default LMHeadMountedRig prefab by removing old unused objects in its CapsuleHand prefab
- [x] Changes the focalPlaneOffset for the controller to 10 cm, improving apparent VR vs. reality hand positions
- [x] Adds a warning when isHeadMounted is detected not to be in the correct state on Start() and chooses a good setting on Reset() automatically
- [x] Removes overrideDeviceType, always returning device type settings for the peripheral.


Assets/LeapMotion/Core/Scripts/LeapServiceProvider.cs
```diff
@@ -292,6 +267,31 @@ public enum FrameOptimizationMode {
...
+                           + "camera targeting one or both stereo eyes; You should "
+                           + "check the isHeadMounted option on the LeapServiceProvider "
+                           + "if the Leap is mounted or attached to your VR headset!",
+                             this);
+            }
+            shouldEnable = true;
+          }
+        }
+      }
+      shouldEnable = false;
```
Amarcolina: There are no other ways this function can exit right?  Which would mean this method can only ever exit with `shouldEnable` equal to false.  The line 288 does nothing.

Amarcolina: Why not just return a bool?  That way you can just return the result when you need to instead of assigning to a variable?  And then you can get rid of the overload because people can just ignore the return value.

nickjbenson: fixed

Amarcolina:
All looks good here!

PR #803, opened 2017-07-28
Moved progress bar to Core

Amarcolina:
Progress bar is a generic util and should not live inside of the Graphical Renderer namespace


Assets/LeapMotion/Core/Scripts/Utils/ProgressBar.cs
```diff
@@ -14,7 +14,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 
-namespace Leap.Unity.GraphicalRenderer {
+namespace Leap.Unity {
```
nickjbenson: My general comment for this file is that it was unclear to me getting started what class I should actually use to create the progress bar dialog. I get it now; but maybe some hints in <summary> tags will help us ramp up to proper usage faster (e.g. have ProgressBar say "if you just want an editor progress bar window, use EditorProgressView")

PR #805, opened 2017-08-02
All-in-One Warping (Collapsing the Leap Rig Hierarchy)

nickjbenson:
**Do Not Merge | PR in Progress**

Creating this PR so that this task is more on our radar. I see it as a target for Core 4.4.

The branch is sliightly old, so first we have to merge the other way, and John said he got about 3/4 of the way through with it. One big remaining task is getting temporal warping on the image pass-through functionality.


Assets/LeapMotion/Scripts/Utils/TransformHistory.cs
```diff
@@ -0,0 +1,67 @@
...
+    public struct TransformData {
+      public long time; // microseconds
+      public Vector3 position; //meters
+      public Quaternion rotation; //magic
+
+      public static TransformData Lerp(TransformData from, TransformData to, long time) {
+        if (from.time == to.time) {
+          return from;
+        }
+        float fraction = (time - from.time) / (to.time - from.time);
```
Amarcolina: This is doing integer arithmetic and then casting to floating point.  `fraction` will be 0 most of the time, and maybe 1 once in a while.  Should case to double first (for precision) and then do the mathematics, then cast to float to get a correct value.

Assets/LeapMotion/Scripts/Utils/TransformHistory.cs
```diff
@@ -0,0 +1,67 @@
...
+        };
+      }
+
+      public static TransformData GetTransformAtTime(RingBuffer<TransformData> history, long desiredTime) {
+        for (int i = history.Length - 1; i > 0; i--) {
+          if (history.Get(i).time >= desiredTime && history.Get(i - 1).time < desiredTime) {
+            return Lerp(history.Get(i - 1), history.Get(i), desiredTime);
+          }
+        }
+        return history.GetLatest();
```
Amarcolina: If the time is out of range of the history, might be a bad idea to just return the latest, since that is not necessarily the frame data closest to the desired time.  We should either return the latest or earliest (depending on desired time).  Or, if this is something that signifies an error, log it!

Assets/LeapMotion/Scripts/Utils/TransformHistory.cs
jselstad: This... might explain some things.

jselstad:
Closed in lieu of #841

PR #806, opened 2017-08-03
Fixes for Release Core 4.3.0 branch

nickjbenson:
This PR:

- [x] Fixes an InvalidOperationException when calling intObj.ReleaseFromGrasp
- [x] Prevents null reference errors with interaction objects that contain colliders on inactive GameObjects
- [x] Sets the layers of new colliders properly if they are added to an interaction object at runtime (note: still have to manually call RefreshInteractionColliders() when changing colliders at runtime)
- [x] Fixes a bug in LeapBoxGraphic that caused right-side and downward faces to render backwards
- [x] Refactored LeapPanelGraphic, LeapBoxGraphic, and LeapPanelOutlineGraphic to share code, included each graphic type in the Curved Spaces example scene
- [x] Fixes the Attach Object / Detach Object button for AnchorableBehaviour when only one AnchorableBehaviour is selected


Assets/LeapMotion/Modules/GraphicRenderer/Scripts/Graphics/ProceduralMeshGraphics/LeapSlicedGraphic.cs
```diff
@@ -0,0 +1,389 @@
...
+  /// - Are generally rectangular, supporting dynamic tesselation along the X and Y axes
+  /// - Can be intuitively controlled by an associated RectTransform
+  /// - Support nine-slicing when using a Sprite as a source for Texture data
+  /// </details>
+  [DisallowMultipleComponent]
+  public abstract class LeapSlicedGraphic : LeapMeshGraphicBase {
+
+    #region Constants
+
+    public const int MAX_VERTS_PER_AXIS = 128;
```
Amarcolina: This isn't the case in general, since for boxes your per-axis limit will be lower due to the higher dimensions.

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/Graphics/ProceduralMeshGraphics/LeapSlicedGraphic.cs
```diff
@@ -0,0 +1,389 @@
...
+      Vertices,
+      VerticesPerRectilinearMeter
+    }
+
+    public struct Vector2i {
+      public int x;
+      public int y;
+    }
+    
+    public struct RectMargins {
```
Amarcolina: Any reason you don't just use the Rect struct?

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/Graphics/ProceduralMeshGraphics/LeapSlicedGraphic.cs
nickjbenson: ?? MAX_VERTS_PER_AXIS is only used within LeapSlicedGraphic to cap the resolution passed to mesh implementations; Boxes get double this number, as they have front and back faces, and this isn't taken into account when the resolution cap is applied. (Also, their depth axis is always resolution zero.) I renamed the variable `MAX_RESOLUTION` to make this more clear.

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/Graphics/ProceduralMeshGraphics/LeapSlicedGraphic.cs
nickjbenson: Yeah, you can't set left/right/top/bottom directly in a Rect struct, and the meaning of right and bottom is different in a Rect struct.

Right and Bottom in RectMargins are measured from the right and bottom edges respectively, instead of referring to the distance along the relevant axis _to_ the right and bottom edges (In UV space, 1 - right or 1 - bottom). I added some documentation to make that more clear.

Amarcolina: Ah got it!

PR #807, opened 2017-08-04
Adding AssetFolder Property

Amarcolina:
Adding a new property type named AssetFolder.  It allows you to easily serialized references to folders that persist event when the folder is renamed or moved.  It is an editor-only feature since folder structure doesn't really exist at runtime.  The property provides:
 - Access to get/set of the folder path
 - Ability to use the folder dialog to set the folder
 - Ability to drag/drop a folder into the property inspector

This also moves the relative path code into Utils where it belongs instead of the package management code.

To test:
 - [ ] Verify that you can create a script that can serialize an AssetFolder property
 - [ ] Verify that the inspector gui for the AssetFolder behaves correctly

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/AssetFolder.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/AssetFolder.cs
@@ -1,4 +1,13 @@
-﻿using System;
+/******************************************************************************
+ * Copyright (C) Leap Motion, Inc. 2011-2017.                                 *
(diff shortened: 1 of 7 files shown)


nickjbenson:
Probably would be good to show an error message if the user picks a folder that's outside the Assets folder (if that's possible).

The other thing is we've got EnableLeapTests breaking the build. Not related to this PR but it's the first one I'm focusing on merging into develop so if you just add in the #if UNITY_EDITOR there we can merge it in and not deal with another PR

PR #819, opened 2017-08-17
Feature streaming folder

Amarcolina:
Added StreamingFolder and StreamingAsset, which use the same kind of interface as AssetFolder.  They allow you to serialize references to folders and assets located within the streaming assets folder.  They automatically return the correct full path to the files or directories they reference, both in the build and in the editor.  These references remain correct even if the files referenced get moved or renamed!

To test:
 - [x] Verify that StreamingFolder can only accept folders and StreamingAsset can only accept files
 - [x] Verify that StreamingFolder and StreamingAsset only accept things located within the StreamingAsset folder
 - [x] Verify that all of this works through both the explorer interface (the folder button) and when you drag things into the field.
 - [x] Verify that AssetFolder still works as expected

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/AssetFolderPropertyDrawer.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/AssetFolderPropertyDrawer.cs
@@ -7,6 +7,8 @@
  * between Leap Motion and you, your company or other organization.           *
  ******************************************************************************/
 
+using System.IO;
+using System.Linq;
 using UnityEngine;
 using UnityEditor;
 using Leap.Unity.Query;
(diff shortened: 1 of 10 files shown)


Assets/LeapMotion/Core/Scripts/DataStructures/Editor/AssetFolderPropertyDrawer.cs
```diff
@@ -73,5 +83,31 @@ public class AssetFolderPropertyDrawer : PropertyDrawer {
     public override float GetPropertyHeight(SerializedProperty property, GUIContent label) {
       return EditorGUIUtility.singleLineHeight;
     }
+
+    protected virtual string PrompUserForPath(string currentPath) {
```
nickjbenson: sp

Assets/LeapMotion/Core/Scripts/DataStructures/Editor/StreamingAssetPropertyDrawer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System.IO;
+using UnityEditor;
+
+namespace Leap.Unity {
+
+  [CustomPropertyDrawer(typeof(StreamingFolder), useForChildren: true)]
+  public class StreamingAssetPropertyDrawer : StreamingFolderPropertyDrawer {
+
+    protected override string PrompUserForPath(string currentPath) {
```
nickjbenson: sp

PR #823, opened 2017-08-23
Fix Button Parent Scales

jselstad:
Prior buttons assumed that the button parent transform existed in unit space (which the examples were designed to satisfy).   Breaking this assumption drastically changed the scale of the forces upon the button.

This PR adjusts the drag and spring forces to take into account the lossy scale of the button parent.

### DIFF
--- a/Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/Editor/InteractionButtonEditor.cs
+++ b/Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/Editor/InteractionButtonEditor.cs
@@ -56,6 +56,10 @@ namespace Leap.Unity.Interaction {
         }
       }
 
+      EditorGUILayout.EndHorizontal();
+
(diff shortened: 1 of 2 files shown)


nickjbenson:
Worried about the use of lossyScale. Under what conditions will this screw us over? Can we use localScale instead?

Amarcolina:
Unity docs have this to say
> If your objects are not skewed the value will be completely correct
So if we assume that the user has configured things correctly (which jon added an inspector warning to catch that case) it should be no different than calculating a correct scale for the parent

nickjbenson:
Ahhh, OK. In that case the code looks good to me. Alex, do you want to test the scene since you encountered this issue when working with IE buttons?

Amarcolina:
It's already been verified by me, I sign off and will merge!

Amarcolina:
Forgot, there was a weird thing going on with the inspector if certain warnings existed at the same time.  I posed a link to @jselstad in slack

PR #824, opened 2017-08-26
TryGrasp (also, HierarchyRecorder)

nickjbenson:
WHEW, this one was a long time coming.

Added TryGrasp() API to interaction controllers and, along with it, our first grasping unit tests!

This PR also contains the Hierarchy Recorder to support those unit tests, so this merge is a bigger deal than just a new grasp feature.

- [x] Teleport an object into the hand and call TryGrasp() to make sure that it works (there's a unit test for this!)
- [x] Make sure the hierarchy recorder is in a good enough state for develop.
- [x] Make sure the VectorHand Leap encoding code is in a good enough state for develop.


Assets/LeapMotion/Core/Scripts/Attributes/SerializedTypeNameImplementsAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+        property.stringValue = _implementingTypes[selectedIdx].FullName;
+      }
+
+      if (curSelectedIdx == -1 && _implementingTypes.Count > 0) {
+        curSelectedIdx = 0;
+        property.stringValue = _implementingTypes[curSelectedIdx].FullName;
+      }
+    }
+
+    private void refreshImplementingTypes() {
```
Amarcolina: It would be really nice if this didn't search through all assemblies every time it was drawn, that could slow down the entire inspector if there are a few of these.  Since types cannot change without a recompile, you would be safe just caching the results.

Amarcolina: In fact you could populate both this list, and the list of gui content from within the constructor of this attribute!  No caching needed, just initialization!

Assets/LeapMotion/Core/Scripts/Testing/LeapTestBase.cs
```diff
@@ -31,14 +31,32 @@ public abstract class LeapTestBase {
     /// the testing object by overriding this method.
     /// </summary>
     protected virtual void InitTest(string objectName) {
+      testObj = LoadObject(objectName);
+    }
+
+    #region Spawn Utilities
+
+    protected T Spawn<T>(T original, Vector3 position) where T : MonoBehaviour {
```
Amarcolina: It would be nice to relax the generic constraint so that T can just be of type UnityEngine.Object instead of UnityEngine.MonoBehaviour.  This is the same generic constraint that Instantiate<T> has, which allows you to not only spawn other components, but game objects as well!

Assets/LeapMotion/Core/Scripts/Utils/HandUtils.cs
```diff
@@ -327,4 +331,151 @@ public class HandEvent : UnityEvent<Hand> { }
...
+                            int id,
+                            float confidence,
+                            float grabStrength,
+                            float grabAngle,
+                            float pinchStrength,
+                            float pinchDistance,
+                            float palmWidth,
+                            bool isLeft,
+                            float timeVisible,
+                            /* Arm arm,*/
```
Amarcolina: Hm why does this not contain an initialization for the Arm?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -215,6 +215,16 @@ public static class Utils {
 
     #endregion
 
+    #region String Utils
+    public static string TrimEnd(this string str, int characters) {
+      return str.Substring(0, Mathf.Max(0, str.Length - characters));
+    }
+
+    public static string TrimStart(this string str, int characters) {
+      return str.Substring(characters);
```
Amarcolina: Would be good to have a bounds clamp here too, like in TrimEnd

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/LeapRecording.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// Loads this recording with data from the provided list of frames. These frames
+    /// reflect raw recorded Leap data. The actual frame storage scheme utilized is up to
+    /// the implementation.
+    /// 
+    /// If this recording already has frame data, a call to this method should overwrite
+    /// that data.
+    /// </summary>
+    public abstract void LoadFrames(List<Frame> frames);
```
Amarcolina: We should be specific in the description of this method whether or not the argument list can be expected to be destroyed or kept around by the recording.  Currently the raw recording just holds onto it, but that might have unexpected results if the user goes off and does something else with their list later.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/QuaternionCurves.cs
```diff
@@ -0,0 +1,78 @@
...
+                                            zCurve.Evaluate(time),
+                                            wCurve.Evaluate(time));
+      return Quaternion.Lerp(evaluated, evaluated, 1.0f);
+    }
+
+    /// <summary>
+    /// Compresses the curves in these QuaternionCurves using
+    /// AnimationCurveUtil.CompressRotations.
+    /// </summary>
+    public void Compress() {
```
Amarcolina: This probably should have some sort of argument to control the error amount ?

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorFrameCurves.cs
```diff
@@ -0,0 +1,59 @@
...
+      leftHandCurves .AddKeyframes(keyframeTime, leftHand);
+      rightHandCurves.AddKeyframes(keyframeTime, rightHand);
+    }
+
+    /// <summary>
+    /// Processes all of the current keyframe data and compresses them using
+    /// AnimationCurveUtil.
+    /// </summary>
+    public void CompressCurves() {
+      leftHandCurves .Compress();
```
Amarcolina: Has formatting gone too far???
Really tho, I had no idea that you could put whitespace between the object and the period for methods and stuff.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording.cs
```diff
@@ -0,0 +1,79 @@
...
+    private Hand _leftHand = new Hand();
+    private Hand _rightHand = new Hand();
+
+    public override float length {
+      get {
+        return _maxKeyframeTime - _minKeyframeTime;
+      }
+    }
+
+    private float _minKeyframeTime = 0f;
```
Amarcolina: Do these min and max frame times need to be serialized?  As far as I can tell they will always end up being zero.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHand.cs
```diff
@@ -0,0 +1,195 @@
...
+    public const int NUM_JOINT_POSITIONS = 25;
+
+    public bool       isLeft;
+    public Vector3    palmPos;
+    public Quaternion palmRot;
+    
+    private Vector3[]  _backingJointPositions;
+    public Vector3[] jointPositions {
+      get {
+        if (_backingJointPositions == null) {
```
Amarcolina: You could do `public readonly Vector3[] jointPositions = new Vector3[NUM_JOINT_POSITIONS]` and get the same functionality.  Can still read and write to the array, but not change it's size, and it's automatically allocated for you.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHandCurves.cs
```diff
@@ -0,0 +1,149 @@
...
+
+          palmPosCurves.AddKeyframes(time, vectorHand.palmPos);
+          palmRotCurves.AddKeyframes(time, vectorHand.palmRot);
+
+          for (int i = 0; i < VectorHand.NUM_JOINT_POSITIONS; i++) {
+            jointPositionCurves[i].AddKeyframes(time, vectorHand.jointPositions[i]);
+          }
+        }
+        finally {
+          Pool<VectorHand>.Recycle(vectorHand);
```
Amarcolina: Make sure to clear your list before you recycle it!

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHandCurves.cs
```diff
@@ -0,0 +1,149 @@
...
+          intoHand.FillTemporalData(_lastHand, Time.deltaTime);
+
+          _lastHand.CopyFrom(intoHand);
+        }
+        else {
+          _lastHand = new Hand();
+        }
+      }
+      finally {
+        Pool<VectorHand>.Recycle(vectorHand);
```
Amarcolina: Same here, always recycle!

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHand.cs
Amarcolina: Blarg I see, this is inside a struct, makes sense

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHandCurves.cs
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      return isTracked;
+    }
+    
+  }
+
+  public static class VectorHandCurvesExtensions {
+
+    public static void AddBooleanKey(this AnimationCurve curve, float time, bool value) {
```
Amarcolina: Sounds like this would fit well inside AnimationCurveUtility

ProjectSettings/EditorBuildSettings.asset
```diff
@@ -4,4 +4,10 @@
 EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
-  m_Scenes: []
+  m_Scenes:
```
Amarcolina: Make sure to clear these out

ProjectSettings/ProjectSettings.asset
```diff
@@ -511,7 +511,7 @@ PlayerSettings:
   webGLCompressionFormat: 1
   scriptingDefineSymbols:
     1: 
-    7: 
+    7: LEAP_TESTS
```
Amarcolina: Hm we need to figure out how to disable leap tests hahah, we don't want to commit with tests enabled

Assets/LeapMotion/Core/Scripts/Utils/HandUtils.cs
nickjbenson: ask @jselstad

nickjbenson: too bad @jselstad didn't add a comment explaining it
@jselstad @jselstad @jselstad

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: I think this was you tho?

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/LeapRecording.cs
nickjbenson: Good call. I feel like a copy is more appropriate. I personally like to use pooled lists all the time and would prefer not to require that the input list is persistent. Easy enough change for the raw recording implementation. What do you think?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
Amarcolina: Well

Then I will push a fix :D

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorFrameCurves.cs
nickjbenson: I'll admit this is pretty egregious formatting. Sometimes I like to do it when I'm constantly doing left-handed and right-handed operations one-after-the-other throughout a file, because then I can use vertical alignment as a very fast heuristic for whether the code throughout the file is structured correctly (faster bugspotting, in other words).

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/LeapRecording.cs
Amarcolina: I agree, I think a copy is the way to go.  And I don't think the performance overhead of a copy will every be anything to worry about, since the cost of saving a recording in the first place will be pretty large.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/VectorHandCurves.cs
nickjbenson: good call

Assets/LeapMotion/Core/Scripts/Testing/LeapTestBase.cs
nickjbenson: Added variants for GameObject and UnityEngine.Object to cover those uses

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/LeapRecording.cs
nickjbenson: pushed copying

PR #828, opened 2017-08-29
Adding ability to query 2D arrays

Amarcolina:
Can add 3d and 4d if we really need to :P It just takes one class per dimension....


Amarcolina:
looks good to me! 👍

PR #831, opened 2017-08-30
Improve leap tests and fix a bug

Amarcolina:
This PR does 2 things.  First it makes the existing leap tests more verbose when they fail so you can get a better sense of _what_ failed.  Second, it fixes a very small bug in TestHandFactory where the tip position of a finger was incorrect.  It is now assigned to be the same as the distal tip.

To test:
 - [ ] Make sure that all tests pass _except_ for the `Are Binary Equal` test (which needs LeapCSharp changes)
 - [ ] Make sure code looks all good


protodeep:
Looks good here

PR #833, opened 2017-08-31
Feature device alignment adjusments

protodeep:
- Adds and exposes properties for aligning the virtual device position and tilt of the Leap controller to LeapVRTemporalWarping.cs.

- Applies these values during TryGetWarpedTransform()

- Hides these additional sliders in the LeapVRTemporalWarping/Advanced settings under "Allow Manual Device Offset

allow_manual_device_offset

- Instructions are in an unlinked wiki page here:
[https://github.com/leapmotion/UnityModules/wiki/Manual-Leap-Motion-Device-Alignment](https://github.com/leapmotion/UnityModules/wiki/Manual-Leap-Motion-Device-Alignment)


nickjbenson:
Heads-up that the plan for this PR is currently to:

- [ ] Add default values to the Leap device offset struct based on the data we've collected in the office on average Leap-headset offsets. This means adding tilt to the Leap device offset struct and using that, in a similar way this PR already does with the manual offset.
- [ ] Set the default manual offset values to be the average forward offset and tilt we've found in the office to have the best alignment feel when holding VR controllers.

Once we get those in, I'll merge this PR to `develop` and we'll get a PR targeting `master` in so we can get that QA'd and released as `4.3.1`.

nickjbenson:
OK, LeapVRTemporalWarping has been organized, fit to current style conventions, and I've verified that the Core Leap scenes now have a 12 cm forward offset and a 5 degree tilt set as their default values.

@Amarcolina, please confirm the offsets when you enable the manual pose alignment checkbox in a few scenes (IE example scenes for example) and sanity-check that alignment looks good with controllers on your Vive. (And review as usual :) )

Assets/LeapMotion/Core/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -236,15 +261,29 @@ protected struct TransformData {
...
 
+#if UNITY_EDITOR
       if (_headTransform != null && UnityEditor.PlayerSettings.virtualRealitySupported) {
         _trackingAnchor = _headTransform.parent;
       }
-    }
 #endif
 
+      deviceOffsetYAxis = _deviceOffsetYAxis;
+      deviceOffsetZAxis = _deviceOffsetZAxis;
```
Amarcolina: These assignments do nothing right?  Because the properties just seem to be backers for the fields themselves.

Assets/LeapMotion/Core/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -376,10 +410,10 @@ protected struct TransformData {
       Vector3 pastCenterPos = _trackingAnchor.TransformPoint(past.localPosition);
       Quaternion pastCenterRot = _trackingAnchor.rotation * past.localRotation;
 
-      transform.position = Vector3.Lerp(currCenterPos, pastCenterPos, tweenPositionalWarping);
-      transform.rotation = Quaternion.Slerp(currCenterRot, pastCenterRot, tweenRotationalWarping);
+      transform.position = Vector3.Lerp(currCenterPos, pastCenterPos, _tweenPositionalWarping);
+      transform.rotation = Quaternion.Slerp(currCenterRot, pastCenterRot, _tweenRotationalWarping);
 
-      transform.position += transform.forward * deviceInfo.focalPlaneOffset;
+      transform.position += transform.forward * _deviceInfo.forwardOffset;
```
Amarcolina: Is using the forward offset needed still?  What about deviceOffsetZAxis?  There seem to be two concepts here, do we actually need them both?

Assets/LeapMotion/Core/Scripts/VR/LeapVRTemporalWarping.cs
```diff
@@ -69,127 +38,180 @@ protected struct TransformData {
...
-    public float TweenImageWarping {
-      get {
-        return tweenImageWarping;
-      }
-      set {
-        tweenImageWarping = Mathf.Clamp01(value);
-      }
-    }
+    // Manual Device Offset
+    [Tooltip("Allow manual adjustment of the Leap device's virtual offset and tilt. These "
```
Amarcolina: Hmm wanted to discuss a workflow question.  Are we ok with this boolean just showing/hiding the fields, not actually changing the way things work?  For example if someone somehow gets something into a bad state, they might not know that they need to _check_ the toggle in order to fix it.

Assets/LeapMotion/Core/Scripts/VR/LeapVRTemporalWarping.cs
nickjbenson: you're right, this is vestigial now, gotta remove that.

Assets/LeapMotion/Core/Scripts/VR/LeapVRTemporalWarping.cs
nickjbenson: We should just use deviceOffsetZAxis for everything in the LeapVRTemporalWarping script post-this-PR, I just missed it because there's a lot of code, will fix

PR #840, opened 2017-09-07
Pulling toggle from particles

Amarcolina:


Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/InteractionToggle.cs
```diff
@@ -72,11 +72,43 @@ public class InteractionToggle : InteractionButton {
...
+    public Vector3 RelaxedToggledLocalPosition {
+      get {
+        return initialLocalPosition + Vector3.back * Mathf.Lerp(minMaxHeight.x, minMaxHeight.y, toggledRestingHeight);
+      }
+    }
+
+    /// <summary>
+    /// Returns the local position of this toggle when it is able to relax into its untoggled position.
+    /// </summary>
+    public override Vector3 RelaxedLocalPosition {
```
nickjbenson: This is a compile error; you need to also pull in the InteractionButton, which includes the original RelaxedLocalPosition implementation

Amarcolina: blarg whoops.

nickjbenson:
Actually, the positions are only correct at runtime; this is because the properties require other variables set to work correctly, and those variables are only set on Start. Still approving.

PR #841, opened 2017-09-07
All-in-One Warping (Collapsing the Leap Rig Hierarchy)

jselstad:
# Ready for review!!

Updated version of #805 
- [x] Provider Frustum Gizmos need to be pushed forward by the forward offset
- [x] Temporal Warping variables should be exposed on the provider
- [x] Hand Factory needs to orient the test hand poses properly
- [x] Image Passthrough time warping needs to be fixed
- [x] All of the example scenes need to be ported to the new hierarchy structure

EDIT (nickjbenson):
- [x] Merge HandPool and LeapHandController into HandModelManager
- [x] Move HandModelManager to the `Hand Models` object in the new `Leap Rig`
- [x] Add useful tooltips to the various parts of temporal warping, hand groups, etc. in the core rig
- [x] Add nifty icons to the `LeapServiceProvider` and `LeapXRServiceProvider` scripts


Assets/LeapMotion/Core/Editor/LeapVRServiceProviderEditor.cs
```diff
@@ -23,8 +23,8 @@ public class LeapVRServiceProviderEditor : LeapServiceProviderEditor {
                              decorateAllowManualTimeAlignment);
 
       specifyConditionalDrawing(() => { return serializedObject
-                                                 .FindProperty("_disableTemporalWarping")
-                                                   .boolValue == false; },
+                                                 .FindProperty("_enableTemporalWarping")
```
nickjbenson: My "No no no" comment refers to this change (and the others in the "I don't like inverse checkboxes" commit)

nickjbenson:
This PR is now _OPEN FOR BUSINESS_ and ready for review!

Assets/LeapMotion/Core/Scripts/HandModelManager.cs
```diff
@@ -13,37 +13,159 @@
...
+  /// When a HandRepresentation is created, a HandModelBase is removed from the pool.
+  /// When a HandRepresentation is finished, its HandModelBase is returned to the pool.
+  /// 
+  /// This class was formerly known as HandPool.
+  /// </summary>
+  public class HandModelManager : MonoBehaviour {
+    
+    #region Formerly in LeapHandController
+
+    protected LeapProvider provider;
```
Amarcolina: This is a duplicate reference to a provider, there is already one (in the inspector too) named _leapProvider.

Assets/LeapMotion/Core/Scripts/XR/XRHeightOffset.cs
```diff
@@ -0,0 +1,129 @@
...
+      autoRecenterOnUserPresence = false;
+    }
+  }
+
+  private void Update() {
+    var deviceIsPresent = UnityEngine.XR.XRDevice.isPresent;
+    if (deviceIsPresent) {
+      
+      if (enableRuntimeAdjustment) {
+        if (Input.GetKeyDown(stepUpKey)) {
```
Amarcolina: I would favor using actual KeyCodes over using named input fields, since we don't know what the users input state might be like.

nickjbenson: KeyCode enums just have a terrible inspector experience.. :(

Assets/LeapMotion/Core/Scripts/XR/Editor/EyeTypeEditor.cs
```diff
@@ -12,11 +12,11 @@
 
 namespace Leap.Unity {
 
-  [CustomPropertyDrawer(typeof(EyeType))]
```
Amarcolina: This code is still commented out, was there a TODO here?

Assets/LeapMotion/Core/Scripts/XR/LeapXRCameraControl.cs
```diff
@@ -0,0 +1,89 @@
...
+using UnityEngine.Rendering;
+using System;
+
+namespace Leap.Unity {
+
+  /// <summary>
+  /// Provides a handful of VR related camera utilities, such as controlling IPD and
+  /// camera distance.
+  /// </summary>
+  public class LeapXRCameraControl : MonoBehaviour {
```
Amarcolina: Would recommend renaming this to something like `LeapEyeDislocator` or something, since that's all it's really used for now.

Assets/LeapMotion/Core/Scripts/XR/LeapXRCameraControl.cs
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+
+    [SerializeField]
+    private EyeType _eyeType = new EyeType(EyeType.OrderType.CENTER);
+
+    /// <summary>
+    /// Called during the left eye camera's OnPreRender Unity callback.
+    /// </summary>
+    public static Action<Camera> OnLeftPreRender;
```
Amarcolina: We can probably remove these since they are not used anywhere and the reason they were created is no longer needed.

Assets/LeapMotion/Core/Scripts/XR/XRHeightOffset.cs
Amarcolina: I would argue strings are even worse :P You have to type in text and you don't know if you are correct or incorrect until you press play.

Amarcolina:
I have a group visible in the inspector for all hand types?  Is this intentional?

Assets/LeapMotion/Core/Scripts/LeapDeviceInfo.cs
```diff
@@ -57,7 +57,7 @@ public struct LeapDeviceInfo {
 
       LeapDeviceInfo deviceInfo;
       deviceInfo.isEmbedded = false;
-      deviceInfo.baseline = 0.04f;
+      deviceInfo.baseline = 0.064f;
```
Amarcolina: If we are still hardcoding this, we should at least hardcode it to the value used by peripheral (0.04) since that is what developers are going to experience.

Assets/LeapMotion/Core/Scripts/LeapImageRetriever.cs
```diff
@@ -386,14 +390,14 @@ public class EyeTextureData {
     }
 
     public void ApplyCameraProjectionValues(Camera camera) {
```
Amarcolina: Looks like this file might have been in the middle of some cleanup?  Looks like ApplyCameraProjectionValues and friends can all be removed?

Assets/LeapMotion/Core/Scripts/LeapTestProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+           + "create a test-pose right hand at this transform's position and rotation."
+           + "Setting this binding to null at runtime will cause the hand to disappear "
+           + "from Frame data, as if it stopped tracking.")]
+    public Transform rightHandBasis;
+    private Hand _rightHand = null;
+    private Hand _cachedRightHand = null;
+
+    void Awake() {
+      _cachedLeftHand = TestHandFactory.MakeTestHand(isLeft: true);
+      _cachedLeftHand.TransformToUnityUnits(); // Note: sucks that this has to be here
```
Amarcolina: How about we add it as an enum or a bool switch in the Make method instead then?

Assets/LeapMotion/Core/Scripts/XR/EyeType.cs
```diff
@@ -104,4 +105,4 @@ public enum OrderType {
       _hasBegun = false;
     }
   }
-}
+}
```
Amarcolina: You _removed_ the newline at the end of the file? ಠ_ಠ

Assets/LeapMotion/Core/Scripts/XR/XRHeightOffset.cs
```diff
@@ -0,0 +1,129 @@
...
+  #endregion
+
+  #endregion
+
+  #region Unity Events
+
+  private Vector3 _startingPosition = Vector3.zero;
+
+  private void Start() {
+    _startingPosition = this.transform.position;
```
Amarcolina: Using a starting position would prevent the transform from being moved normally correct?  If this is desired behavior, we might want to consider enforcing it every frame, because otherwise people might be bitten when their transform moves back to a location whenever the user takes off their headset and puts it back on again.  (or adjusts the stationaryHeightOffset)

Assets/LeapMotion/Core/Scripts/XR/XRHeightOffset.cs
nickjbenson: When I made the change I had forgotten that the strings correspond to InputManager entries; if you could type a string and have it match to a KeyCode of the same name, that's nice and quick assuming you know the name of the key you want.

Now it's on my list to add an attribute for KeyCodes that will improve its inspector -- I think a dropdown with a properly categorized menu would be a big improvement

Assets/LeapMotion/Core/Scripts/XR/Editor/EyeTypeEditor.cs
nickjbenson: Nope. So I just removed the commenting-out for now. EyeType and Leap XR Camera Control is on the chopping block but it'll be removed later when it's easier to figure out how to do that :P

Assets/LeapMotion/Core/Scripts/XR/LeapXRCameraControl.cs
nickjbenson: Sounds good!

Assets/LeapMotion/Core/Scripts/XR/LeapXRCameraControl.cs
nickjbenson: good call

Assets/LeapMotion/Core/Scripts/LeapDeviceInfo.cs
nickjbenson: I definitely don't want to hard-code this anymore, but let's address that as part of the Image API PR, yeah?

Changing it back to 0.04f in the meantime

Assets/LeapMotion/Core/Scripts/LeapImageRetriever.cs
nickjbenson: Yep good catch. That was some of the first work I did on collapsing the hierarchy. Removed that stuff now

Assets/LeapMotion/Core/Scripts/LeapTestProvider.cs
nickjbenson: I think that hides the problem one level deeper without addressing it fully, but the "not this PR" syndrome is real..

I'll add a switch in the method call for now but will want to make more changes in a later PR.

Assets/LeapMotion/Core/Scripts/XR/EyeType.cs
nickjbenson: Not _intentionally_ >_<

Amarcolina:
If the Camera has a non-identity transform relative to it's parent in the editor, hands will not appear in the correct location.

Assets/LeapMotion/Core/Scripts/LeapXRServiceProvider.cs
```diff
@@ -217,9 +213,19 @@ public enum TemporalWarpingMode {
...
-                                      InputTracking.GetLocalRotation(XRNode.CenterEye))
-                               .Then(_trackingBaseDeltaPose);
+      // Get most recent tracked pose.
+      var trackedPose = new Pose(InputTracking.GetLocalPosition(XRNode.CenterEye),
+                                 InputTracking.GetLocalRotation(XRNode.CenterEye));
+
+      // If we don't know of any pose offset yet, account for it by finding the pose
+      // delta from the "local" tracked pose to the actual camera pose.
+      if (!_trackingBaseDeltaPose.HasValue) {
+        _trackingBaseDeltaPose = Camera.main.transform.ToWorldPose().From(trackedPose);
```
Amarcolina: We shouldn't use Camera.main, it may not even exist in the scene, or be the actual camera we are attached to.  I think _cachedCamera is the one to use?

nickjbenson: Ah sure, good catch

nickjbenson:
Merged `develop`; I also merged two fixes from LeapPaint:
- From/Then syntax was actually not multiplying in the correct order to get the intuition I intended; the tests should still pass for it. It's still a right-ward syntax due to the order that function calls are resolved (for a chain, e.g. A.Then(B).Then(C) == ((A * B) * C), but NOT (A * B * C)
- Edit-time poses were/are bugged, so there's a stop-gap fix that produces some garbage but only during edit-time. I still want to diagnose the problem in LeapServiceProvider that caused frames to accumulate offsets in hands, but in the mean time, edit-time frames behave correctly now (test is the AttachmentHands scene, since attachment hands use edit-time frames; the HandModel pipeline DOES NOT use edit-time frames, although it probably should)

nickjbenson:
Blargh, I do see the FromPoseToPose tests failing, although all the OTHER From/Then tests are still passing. I'll look at those soon.

PR #842, opened 2017-09-08
Adding curve bounds attribute

Amarcolina:
This PR adds an attribute that allows you to put specific bounds on an AnimationCurve.  This actually changes the curve interface to prevent you from entering a curve that falls outside of the bounds!  A simple variant of the attribute also exists to annotate unit bounds (0-1) for readability. 

Also added a few default curve values so that you can have your curves take on a sensible default value other than having no keys at all.

To test:
 - [ ] Make sure that the attribute functions as expected
 - [ ] Make sure that the default curves are all formed correctly


nickjbenson:
This sounds veeerrry nice. Will take a look today.

Assets/LeapMotion/Core/Scripts/Attributes/CurveBounds.cs
```diff
@@ -0,0 +1,60 @@
...
+      this.bounds = bounds;
+      this.color = color;
+    }
+
+    public CurveBounds(float width, float height) {
+      bounds = new Rect(0, 0, width, height);
+      color = Color.green;
+    }
+
+    public CurveBounds(float width, float height, Color color) {
```
nickjbenson: I'm getting a compile error when I try to do
```
  [CurveBounds(2f, 2f, Color.blue)]
  public AnimationCurve arbitraryBoundsCurve;
```
```
Assets/CurveBoundsTest.cs(8,30): error CS0181: Attribute constructor parameter has type `UnityEngine.Color', which is not a valid attribute parameter type
```

Assets/LeapMotion/Core/Scripts/Attributes/CurveBounds.cs
```diff
@@ -0,0 +1,60 @@
...
+using UnityEngine;
+
+namespace Leap.Unity.Attributes {
+
+  /// <summary>
+  /// You can use this attribute to mark that an AnimationCurve can only have
+  /// values that fall within specific bounds.  The user will be prevented from
+  /// entering a curve that lies outside of these bounds.
+  /// </summary>
+  public class CurveBounds : CombinablePropertyAttribute, IFullPropertyDrawer {
```
nickjbenson: Shouldn't Attributes be called e.g. CurveBoundsAttribute?

Amarcolina: Ah yes!  Always forget about that...

Assets/LeapMotion/Core/Scripts/Attributes/CurveBounds.cs
Amarcolina: OMG I never new this... 
> Only primitive constants or arrays of primitives can be used as attribute parameters.
This is crazy!  Ok well i guess I could either remove color or have users input RGB independently, what do you think?

PR #843, opened 2017-09-08
Sync master into develop

nickjbenson:
- [ ] We renamed the "AR" scenes to the more appropriate "VR - Infrared Viewer" scheme
- [ ] James got us a new LeapCSharp.NET DLL and matching libLeapC
- [ ] Version .txt was updated


Amarcolina:
Hm why is the source branch not master?

PR #844, opened 2017-09-08
Make anchor requirements readable, fix logic bug

nickjbenson:
JCorvinus on the dev forums pointed out an obvious issue with GetNearestValidAnchor's testAnchor checking:

```
foreach (var testAnchor in anchorsToCheck) {
  if ((requireAnchorHasSpace && (!anchor.allowMultipleObjects || anchor.anchoredObjects.Count == 0))
      || (requireAnchorActiveAndEnabled && !anchor.isActiveAndEnabled)) continue;
```

`anchor` != `testAnchor` -- also, the logic here is really unreadable, so I fixed it, and in the process I think I also fixed an error in the logic.


Amarcolina:
Love the improvement!  Looks great!

PR #849, opened 2017-09-09
Address physics stutter when physics timestep != update timestep

Amarcolina:
We are working towards allowing the IE to support physics timesteps that are not equal to the update timestep.  This is a feature that could be very useful to developers who want to save on performance by computing less physics, or for developers who are stuck on certain physics framerates for other reasons specific to their game (we are in this boat for particles!).  

This PR takes one step closer to supporting this functionality.  Currently our service provider can provide update frames, and physics frames, but until this PR the physics frames were not calculated using correct timestamps, making them jitter!  This PR uses the offset calculated in Update to calculate a correct timestamp from within FixedUpdate.  This will _increase latency_ of the frames provided to the physics system.  Later PR's hope to solve this latency with the same system that helped us out in the old IE, the rigidbody warper (maybe).

Testing:
 - To test a specific physics timestep, go to Edit->Project Settings->Time and set the `Fixed Timestep` and `Maximum Allowed Timestep` both to 1 / the desired framerate.  Use the first interaction example for testing.  The warning dialog will probably pop up and yell at you, but you can ignore it.  Once these sequence of changes are complete, we might be getting rid of it anyway.

 - Run all tests with the interaction behavior GraspMovementBehaviour set both to Kinematic and NonKinematic.  

 - [ ] Make sure interactions still feel good and cause no judder when the physics framerate is equal to 90 when on a desktop device.
 - [ ] Make sure interactions still feel good and cause no judder when the physics framerate is equal to 60 when on a mobile device.
-  [ ] Make sure interactions still feel good and cause no judder when the physics framerate is equal to 60 when on a desktop device.
 - [ ] Make sure interactions still feel good and cause no judder when the physics framerate is equal to 45 when on a mobile device.

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/LeapServiceProvider.cs
+++ b/Assets/LeapMotion/Core/Scripts/LeapServiceProvider.cs
@@ -20,9 +20,9 @@ namespace Leap.Unity {
   /**LeapServiceProvider creates a Controller and supplies Leap Hands and images */
   public class LeapServiceProvider : LeapProvider {
     /** Conversion factor for nanoseconds to seconds. */
-    protected const float NS_TO_S = 1e-6f;
+    protected const double NS_TO_S = 1e-6;
     /** Conversion factor for seconds to nanoseconds. */
-    protected const float S_TO_NS = 1e6f;
+    protected const double S_TO_NS = 1e6;
     /** Transform Array for Precull Latching **/
     protected const string HAND_ARRAY = "_LeapHandTransforms";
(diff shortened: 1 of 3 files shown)


mmenicutch:
Took a look at this with the different framerates on both PC and mobile, I didn't see any judder and the interactions still felt good at all framerates. Seems good to go for me.

nickjbenson:
Woops, John and I looked at this together so I'm happy with this as well.

PR #851, opened 2017-09-11
Re-init bones when soft contact is enabled

Amarcolina:
To prevent bones from going away to infinity under certain situations.


jselstad:
If I recall, the core thing to test with this subsystem is whether there are any jumps when letting go of objects palm-up (so they drop into your hand immediately) and throwing

jselstad:
This all feels good too (palm-up releases, throwing, and two-handed interactions tested).

PR #855, opened 2017-09-11
Fixing capsule hand scaling issues

Amarcolina:
Capsule hand meshes were not being calculated or displayed correctly at different scales.  This PR should allow the hands to be displayed correctly when the camera rig changes its scale.

To test:
 - [x] Visit the example VR scene and scale the camera rig up and down, verify that the capsule hands do not artifact in edit more or in playmode.


jselstad:
Given Nick's latest changes, this works flawlessly too!

PR #858, opened 2017-09-12
Improve enabled disabled concept for GR

Amarcolina:
Graphics now can be detached/attached to groups _independently_ of being enabled or disabled.  Disabling them will make them not visible, but will keep them attached to their group, making it _very_ cheap to show or hide a graphic.

The workflow for attaching or detaching a graphic has been upgraded to reflect this.

To test: 
 - [x] Create a new graphic renderer and add some graphics.  Make sure you can attach/detach/switch groups as expected (both with a single or multiple graphics selected)
 - [x] Make sure that when you drag a graphic into or out of a graphic renderer hierarchy everything works as expected
 - [x] Make sure that enabling or disabling a graphic works as expected for all graphic types and all renderer types, at edit time and at runtime
 - [ ] Make sure the example scenes all still function correctly


nickjbenson:
> Make sure that when you drag a graphic into or out of a graphic renderer hierarchy everything works as expected

Encountered something strange:

Create two GRs. Create a graphic in one group. Give the graphic a non-identity rotation and a non-zero local position. Move the graphic out of its GR into the root of the hierarchy. Then move the graphic into the second GR. Then press Ctrl+Z twice. The graphic will not be in the position it originally started in.

Not sure if the fix is related to the changes in this PR; but Ctrl+Z support is important so I wanted you to know about it

PR #859, opened 2017-09-13
Make anchorableBehaviours actually respect anchorRotation property

nickjbenson:
Herp.

That's all I have to say really

- [x] Verify that an anchorableBehaviour only rotates to match its anchor if anchorRotation is checked


Amarcolina:
Eyy

PR #863, opened 2017-09-13
Don't generate soft contacts for grasped objects

nickjbenson:
We should still get contact callbacks, but we shouldn't be generating nor applying soft contacts for any controllers when the relevant object is grasped by any controller.

- [ ] Make sure you can get contact callbacks from, say, your right hand, on an object held in the left hand
- [x] Make sure soft contact still feels good
- [x] Give it that ol' selstad inspection


jselstad:
>Make sure you can get contact callbacks from, say, your right hand, on an object held in the left hand
This was never going to happen anyway due to the objects being moved to the interaction no-contact layer; so we're good.

PR #864, opened 2017-09-14
Perfect alignment in AR scene without _LeapGlobalProjection shader global

nickjbenson:
This PR guts a large chunk of LeapVRCameraControl that was devoted to setting the _LeapGlobalProjection shader global that was used in the passthrough shader.

Please don't delete this branch after merging, I want to use it in the future to continue gutting LeapVRCameraControl.

The intent of this branch was to kill camera control entirely, so I will continue using the branch to continue that effort, to remove the need for _eyeType. Currently, _eyeType is used to detect from C# whether the camera is about to render a left eye or a right eye, which is used to quickly move the camera _to the virtual Leap camera position corresponding to that eye_ which produces the correct apparent depth for hands in the passthrough shader (compared to the rendered capsule hands). We should be able to cheat the effective camera position for passthrough by a simple in-shader `if` statement on unity_StereoEyeIndex (0 for left, 1 for right), provided by UnityCG.cginc.

Note that only the AR passthrough scene with Capsule Hands is verified working properly. Other scenes have old settings that are most likely bad. Also, we still need to fix image warping on this branch.

- [x] Verify that hand alignment is perfect on the peripheral (Rigel is expected not to work because it has a different baseline)


Amarcolina:
Cool!  Awesome that all of the necessary stuff is already on the GPU waiting to be used.

PR #865, opened 2017-09-14
Add physics extrapolation

Amarcolina:
This PR adds in the ability to extrapolate the physics frames that come out the service provider.  This compensates for the latency introduced by interpolated rigidbodies, which should always been used when your physics framerate is lower than your render framerate.  There are quite a bit of possible configurations, not all of which are desired, and so some extra effort is needed to make sure it is obvious how to configure things.  There is already a ticket open for documentation, but it would be good to brainstorm in-asset solutions as well.

On the service provider there is not a dropdown to select the type of extrapolation to use.  Auto simply extrapolates forward by a time equal to the physics delta time, which should correct accommodate for all interpolation.  None simply doesn't extrapolate at all.  Manual allows the user to specify the amount of time (in seconds) to extrapolate forwards in time.

Tests should be run in similar settings as #849, but in addition to smooth motion, there should also be little to zero visible latency!
 - [x] Example scenes still look fine and work
 - [x] Example scenes show no obvious signs of latency

NEW UPDATE:  Since unity has an interpolation bug we are falling back and NOT supporting framerates that are different from render rate.  Interpolation has been turned off on all scenes as a consequence.


mmenicutch:
I was starting to take a look at this and was wondering if I should just do exactly what I did with the #849 PR - test on both mobile and PC at the same framerates with kinematic on and off?

Amarcolina:
That sounds like the right plan to me!  One thing to make sure to look for is the presence of latency.  With this PR, the objects should have little to no latency _relative to the motion of the hand_

nickjbenson:
Adding a reminder here that when (if?) we do get this merged in, we should remove the Grasp Hold Warping checkbox (that does nothing because the feature has been unimplemented) from the InteractionBehaviour inspector.

mmenicutch:
I checked with a variety of interaction engine scenes at different framerates across both Oculus and Daydream and didn't see any noticeable latency, interactions all still felt good for the most part. I can second @nickjbenson about the responsiveness of the buttons on the hand UI. I'm going to look at that one specifically on daydream at 45fps as I expect it to be worse.

Amarcolina:
@mmenicutch did you get a chance to see how things looked on daydream?

mmenicutch:
Ah yes, sorry I got sick and then forgot to update when I got back. I did look at how things were on daydream and It didn't seem to be any worse or better. I was still having the same amount of trouble interacting withe the buttons as I was at the other FPS's

Amarcolina:
Trouble interacting with just the HandUI, or with all buttons?

Amarcolina:
@testmasterbright @mmenicutch 

We have rolled back the decision to support variable framerates due to being blocked by a Unity bug.  Scenes have been rolled back from using interpolation.  So we now only officially support 90FPS on PC and 60FPS on mobile.

I have updated all of the scenes to turn off interpolation, would be great to get a quick sign off on these scenes again, so we can merge this PR.

mmenicutch:
Ran across a minor issue with the anchors scene that doesn't happen in the current release version. 

The red and blue blocks are askew when placed in their anchor points and can be spun and moved, this only occurs with the red and blue blocks and doesn't seem to effect the purple one. This also doesn't appear to occur when using the Dynamic UI scene.

nickjbenson:
Pushed the fix, just needed to correctly serialize the rotation setting after a fix that made the rotation setting actually do something :)

mmenicutch:
@Amarcolina Finally got a chance to look at this! The issue is fixed and it looks good to merge!

PR #873, opened 2017-09-20
Hotfix for temporal warping default value bug

Amarcolina:
The variable `warpingAdjustment` was previously renamed to `_warpingAdjustment` without using the [FormerlySerializedAs] attribute, and as a result all of the saved data in scenes was lost.  This resulted in the adjustment value reverting back to it's default value, which was 60.  The recommended value for PC experience is 17, and so this introduced a 43 millisecond difference in the warping effects, causing significant swim when the user moved their head.

This PR makes a few changes:
 - Renamed the variable, again without using the [FormerlySerializedAs] attribute.  This is intentional so that any bad values that have been saved into developers scenes get discarded and overwritten by the new _good_ default, 17.
 - Made it so that unless `_allowManualTimeAlignment` is checked, the adjustment value will do NOTHING.  Instead the value is driven by a constant.  This is so that in the future if we want to push out a new value for the default adjustment value, we can just update the constant and all developers scenes will be automatically updated.

To test:
 - [x] Verify that code looks robust and correct.
 - [x] Verify that in _all_ VR scenes the hands appear properly anchored in the world when the user moves their hands.

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/VR/Editor/LeapTemporalWarpingEditor.cs
+++ b/Assets/LeapMotion/Core/Scripts/VR/Editor/LeapTemporalWarpingEditor.cs
@@ -18,7 +18,7 @@ namespace Leap.Unity {
       base.OnEnable();
 
       specifyConditionalDrawing("_allowManualTimeAlignment",
-                                "_warpingAdjustment",
+                                "_customWarpAdjustment",
                                 "_unlockHold",
                                 "_moreRewind",
                                 "_lessRewind");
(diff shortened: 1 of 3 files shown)


nickjbenson:
...Oops. For the record I think that was me who changed the name of the parameter.

Never did like that we had a bad initial value for the temporal warping parameter >_< I should have fixed it when I first saw that the default was no good.

nickjbenson:
Merging develop into this if you don't mind

nickjbenson:
`allowManualDeviceOffset`, when unchecked, causes the serialized device offset values to return to their default values. Do you think we should match that same behavior here for consistency's sake?

Amarcolina:
Good idea, I'll implement that.

PR #874, opened 2017-09-20
Adding utility to re-save all scenes and assets in project

Amarcolina:
This new tool simply allows us to easily re save all of the scenes and assets that exist inside of a project.  This will allow us to basically _apply_ any changes that we might have made to the objects, so that we can actually see them in git.  This will not affect the result of the scenes or assets, just simply make their on-disk representation match what they will actually be loaded as.

I did not actually do the save for this PR since it touches basically _everything_ and I don't want to create merge conflicts.

To test:
 - [ ] Go to Assets/Save all scenes and assets and make sure it works!  It should take a little bit but you should see all of the assets and scenes updated in git


Assets/LeapMotion/Testing/Scripts/Editor/ResaveUtility.cs
```diff
@@ -0,0 +1,66 @@
...
+    [MenuItem("Assets/Save All Scenes And Assets")]
+    public static void SaveAllScenesAndAssets() {
+      SaveAllScenes();
+      SaveAllAssets();
+    }
+
+    public static void SaveAllScenes() {
+      EditorSceneManager.SaveCurrentModifiedScenesIfUserWantsTo();
+
+      var paths = AssetDatabase.FindAssets("t:Scene").
```
nickjbenson: .THIS.MEANS.WAR.

jk it's fine you do you

nickjbenson:
This code is really clean! Very nice. I nearly missed the progress bar code, for a moment I thought it was all Unity API calls!

Getting philosophical for just a moment; Unity's success has to be at least in part due to the grace with which one can actually create an editor extension like this. (...As long as these wonderfully readable APIs actually function the way they're supposed to. :P )

nickjbenson:
One question before approving: The command appears to drop the user into the VRVisualizer scene after running it -- maybe the first or last scene it loads?

I don't mind the scene change because the command should feel very "noticeable", but I wonder if it's more convenient to drop the user back into the scene they were originally editing.

Amarcolina:
The reason I didn't want to deal with dropping them back into the original scene was I didn't want to deal with the edge cases of figuring out which scene to load.  They might have multiple scenes loaded theoretically, or no scenes at all if they just opened an empty unsaved scene.  We could add the feature but since this is really only something we are going to see (notice it's not going into core) I don't think it's too big of a deal.

nickjbenson:
Ah, didn't think about the fact that multiple scenes could be open.

Yep, sounds fine.

PR #877, opened 2017-09-23
Add Rectified Image Example

jselstad:
Adds a new Example to Glint which demonstrates how to receive rectified leap image data from the GPU with very low performance overhead.   This example also demonstrates how to set up a triple-buffered scheme for receiving images continuously within Glint.   Also the existing example was modified to take advantage of ".LoadRawTextureData()" which results in a ~10x speed-up.

The .dll in this PR has also been modified to support unsigned byte and single-channel float textures (which have significantly lower transfer overhead).

Also the redundant array copy in LeapImageRetriever.cs has been removed as well.


jselstad:
1) Added raw image visualization
2) Fixed garbage.

PR #880, opened 2017-09-27
Cubic Hermite Splines

Amarcolina:
Adding new CubpicHermiteSplines to core!  These splines have the useful property that they let you specify the position _and_ velocity at the two endpoints, and smoothly interpolate between them.  Very useful for animation, and giving physical-feeling feedback while still giving animation guarantees.

The splines provided are all structs, and have methods for getting position, velocity, and both at the same time.  Velocity is calculated analytically.  We can probably add a bunch of additional spline utility methods as time goes on and we used them.

Since C# is super annoying and doesn't allow addition or division operators to be used in generics, I opted for a lightweight code-gen solution for generating the float, Vector2, and Vector3 variants of the spline code.  This has made is a lot easier to iterate and evolve the interface over time.  The generation interface is super lightweight, and basically amounts to a structured way to trigger a custom method at edit time.

To test:
 - [x] Run the splines through some tests, see how they behave!
 - [x] Take a look at the code to make sure the splines look like they are suitable
 - [x] Check off on the code-gen solution


Amarcolina:
I added the 2D version simply for convenience and speed, and since the code gen made it easy :)

Yeah I agree we could find a better name than CHS, but expanding the full CubicHermiteSpline also makes me sad :( Since there are so many spline types, do you guys think it makes sense to just bite the bullet and go for the full name?

PR #887, opened 2017-10-03
Fix VS build (...again...)

nickjbenson:
We built a new LeapCSharp.DLL, but there's some issue where you have to delete your machine's System.DLL for .NET 4.0 when you build from platform because (I don't know actually).

As long as we do this before we build new LeapCSharps we should be fine! But we also need to make sure that the VS build is tested before any Core release so we don't regress on it.

To test:
- [x] Open any script after importing UnityModules and tell VS to build
- [x] Make sure scenes are in working order (e.g. we still get hand data from LeapC correctly)


nickjbenson:
@testmasterbright Would you mind having Michael test this PR for having a working VS build?

mmenicutch:
I was able to build a couple of scripts just fine with VS and all the scenes have hands and look like they're working good!

PR #890, opened 2017-10-07
Fix grabbing bug on objects with joints

jselstad:
Any object with a joint will report itself as "Position Locked"; made for some weird behavior when testing with hinged interaction behaviours/objects.

This adds an extra check to ensure that the object that joint is connecting to is kinematic.


Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -1504,18 +1504,18 @@ public enum GraspedMovementType {
 
         Joint[] joints = rigidbody.GetComponents<Joint>();
         foreach (var joint in joints) {
-          if (joint is FixedJoint) {
+          if (joint is FixedJoint && joint.connectedBody.isKinematic) {
```
Amarcolina: You need null checks for all of these since a joint doesn't _have_ to be connected to another body, it can be connected to the 'world'.

Amarcolina:
Looks good!

PR #892, opened 2017-10-12
Updated to 2017.2

Amarcolina:
- Upgrade project version to 2017.2.0f3
 - Re-saved all scenes and assets 
 - Re-baked all lighting for scenes that used it
 - Upgraded script references from VR to XR namespace
 - Upgraded Readme to specify that we only support 2017.2 (if we want to change this, we need to discuss)


nickjbenson:
IE example scenes look good, IR viewer looks good, approving.

Amarcolina:
Ok merging

PR #897, opened 2017-10-20
Pull some things from Particles

nickjbenson:
This makes their signatures match the newer style of IE callbacks, which are also argument-less except in their more specially-decorated versions.

It makes them much faster to hook into, and they never provided any useful information: Anchor references are accessible already via `anchObj.anchor`, and you _always_ already have a reference to an AnchorableBehaviour when you subscribe to the event in the first place. (This was also one of the reasons the Interaction Behaviour callbacks got their signatures simplified.)

- [x] Verify example scenes involving AnchorableBehaviours still work as expected
- [x] Double-check build :P

EDIT: Sorryyyy should have used a new PR for this:

- [x] Adds Pose struct
- [x] Adds From/Then syntax for Vector3, Quaternion, Matrix4x4, and float
- [x] Adds Pose data tracking and getters for InteractionBehaviour


Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/Anchors/AnchorableBehaviour.cs
```diff
@@ -829,15 +829,13 @@ public enum EventType {
...
-
-    private void setupCallback<T>(ref Action<T> action, EventType type)
-                                  where T : AnchorableBehaviour         {
-      action += (anchObj) => _eventTable.Invoke((int)type);
+    private void setupCallback(ref Action action, EventType type) {
+      if (_eventTable.HasUnityEvent((int)type)) {
+        action += () => _eventTable.Invoke((int)type);
+      }
+      else {
+        action += () => { };
```
Amarcolina: The callbacks already have a null delegate when they are initialized, can't this else clause be removed?

nickjbenson: Maybe? This is just the pattern we have in InteractionBehaviour..

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -734,6 +734,85 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+        while (toVisit.Count > 0) {
+          curTransform = toVisit.Pop();
+
+          // Recursively search children and children's children.
+          foreach (var child in curTransform.GetChildren()) {
+            // Ignore children with OwnerType components of their own; its own OwnerType
+            // component owns its own ComponentType components and the ComponentType
+            // components of its children.
+            if (child.GetComponent<OwnerType>() == null
+                && (includeInactiveObjects || child.gameObject.activeSelf)) {
```
Amarcolina: Not sure if this is what you intend.  Active self can be true even if the game object is deactivated due to a parent being inactive.  Maybe activeHierarchy is what you want instead?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -734,6 +734,85 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+                                                List<ComponentType> ownedComponents,
+                                                bool includeInactiveObjects = false)
+                                               where OwnerType : Component {
+      ownedComponents.Clear();
+      Stack<Transform> toVisit = Pool<Stack<Transform>>.Spawn();
+      List<ComponentType> componentsBuffer = Pool<List<ComponentType>>.Spawn();
+
+      try {
+        // Traverse the hierarchy of this object's transform to find
+        // all of its Colliders.
```
Amarcolina: Out of date comment, not just for colliders anymore

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -764,6 +843,118 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
 
     #endregion
 
+    #region Quaternion Utils
+
+    /// <summary>
+    /// Converts the quaternion into an axis and an angle and returns the vector
+    /// axis * angle.
```
Amarcolina: Probably worth mentioning in the comment that the magnitude of the vector is measured in _degrees_ and not radians.  So the user better not try using this directly with PhysX angular velocity without a conversion for example.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -764,6 +843,118 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+      q.ToAngleAxis(out angle, out axis);
+      return axis * angle;
+    }
+
+    /// <summary>
+    /// Returns a Quaternion described by the provided angle axis vector.
+    /// </summary>
+    public static Quaternion QuaternionFromAngleAxisVector(Vector3 angleAxisVector) {
+      if (angleAxisVector == Vector3.zero) return Quaternion.identity;
+      return Quaternion.AngleAxis(angleAxisVector.magnitude, angleAxisVector.normalized);
```
Amarcolina: would be nice to calculate the magnitude beforehand so that you can use it to normalize the vector for less cost.  Also might want to check if the magnitude of the axis matters at all, it might not, in which case you could skip the normalization step altogether.

Also like above would be good to specify if this method expects radians or degrees.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -764,6 +843,118 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+    }
+
+    /// <summary>
+    /// A.From(B) produces the quaternion that rotates from B to A.
```
Amarcolina: Interesting discussion point, in your explanation of using From, the order is reversed compared to the arguments, and the key operator is _to_ and not from.  Maybe we could entertain adopting that instead?  

`B.To(A)` instead of `A.From(B)`?
`B.Then(B.To(A)) == A` could make sense

thoughts?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -764,6 +843,118 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+    /// </summary>
+    public static Vector3 Then(this Vector3 thisVector, Vector3 otherVector) {
+      return thisVector + otherVector;
+    }
+
+    /// <summary>
+    /// Rightward syntax for applying a Quaternion rotation to this vector; literally
+    /// byQuaternion * thisVector.
+    /// </summary>
+    public static Vector3 Rotate(this Vector3 thisVector, Quaternion byQuaternion) {
```
Amarcolina: Maybe `RotateBy` instead?  To emphasize that the affected party is the vector instead of the argument?  Also might want to consider a separate  naming convention so not to conflict with things like Vector3.Scale, which does a conceptually similar thing, but _modifies_ the vector instead of returning a new one.

```
Vector3 v;
v.Scale(blah);  //legal, modifies v 
v.Rotate(blah); //legal, but doesn't change v!!!
```

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -1039,6 +1230,253 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
       return new Rect(r.x + padding, r.y + padding, r.width - (padding * 2), r.height - (padding * 2));
     }
 
+    /// <summary>
+    /// Returns a new Rect with the argument padding as an outward margin on each border
+    /// of this Rect; the result is a larger Rect.
+    /// </summary>
+    public static Rect PadOuter(this Rect r, float width) {
```
Amarcolina: Would change the name of the argument to `padding` instead of `width` to make sure there is no misunderstanding that it only affects the rect in the X direction.

Amarcolina: Also I see now that this does something conceptually different compared to PadTop/PadBottom....   maybe could be named something different to be super explicit?  Maybe Extrude instead?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -1039,6 +1230,253 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+      return new Rect(r.x - width, r.y - width,
+                      r.width + (width * 2f), r.height + (width * 2f));
+    }
+
+    #region Pad, No Out
+
+    public static Rect PadTop(this Rect r, float padding) {
```
Amarcolina: Not sure what exactly should happen, but I'd like all of these pad methods to have specifically defined edge cases where the padding is greater than a dimension.  Probably good to make sure the user doesn't end up with a rect with negative dimensions.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -1039,6 +1230,253 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+    /// <summary>
+    /// Returns the rotation that makes a transform at objectPosition point its forward
+    /// vector at targetPosition and keep its rightward vector parallel with the horizon
+    /// defined by the upwardDirection normal.
+    /// 
+    /// For example, this will point an interface panel at a user camera while
+    /// maintaining the alignment of text and other elements with the horizon line.
+    /// </summary>
+    public static Quaternion FaceTargetWithoutTwist(Vector3 objectPosition, Vector3 targetPosition, Vector3 upwardDirection, bool flip180 = false) {
+      Vector3 objToTarget = -1f * (Camera.main.transform.position - objectPosition);
```
Amarcolina: Is this supposed to use Camera.main?  Should it be using targetPosition instead?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -803,10 +811,41 @@ public enum GraspedMovementType {
...
 
+    #region Pose & Movement
+
+    private Pose _worldPose;
+    private Pose _worldPoseLastFrame = new Pose();
+    private bool _hasWorldPoseLastFrame = false;
+
+    public Pose worldPose {
+      get {
+        return _worldPose;
```
Amarcolina: Any reason this doesn't just construct the pose on the spot?  If the user moves a rigidbody manually it will become stale.

Assets/LeapMotion/Modules/ImageProcessing.meta
```diff
@@ -0,0 +1,10 @@
+fileFormatVersion: 2
```
Amarcolina: Meta file for nonexistent folder

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -1039,6 +1230,253 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+
+    /// <summary>
+    /// Returns the rotation that makes a transform at objectPosition point its forward
+    /// vector at targetPosition and keep its rightward vector parallel with the horizon
+    /// defined by the upwardDirection normal.
+    /// 
+    /// For example, this will point an interface panel at a user camera while
+    /// maintaining the alignment of text and other elements with the horizon line.
+    /// </summary>
+    public static Quaternion FaceTargetWithoutTwist(Vector3 objectPosition, Vector3 targetPosition, Vector3 upwardDirection, bool flip180 = false) {
```
Amarcolina: This method seems to behave exactly the same way as  Quaternion.LookRotation?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -803,10 +811,41 @@ public enum GraspedMovementType {
...
       fixedUpdateLayers();
+      fixedUpdatePose();
 
       if (_appliedForces) { FixedUpdateForces(); }
     }
 
+    #region Pose & Movement
+
+    private Pose _worldPose;
+    private Pose _worldPoseLastFrame = new Pose();
```
Amarcolina: Maybe<Pose> ?  😁

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
```diff
@@ -0,0 +1,167 @@
...
+    public bool ApproxEquals(Pose other) {
+      return position.ApproxEquals(other.position) && rotation.ApproxEquals(other.rotation);
+    }
+
+    /// <summary>
+    /// Returns a pose interpolated (Lerp for position, Slerp for rotation)
+    /// between a and b by t from 0 to 1. This method clamps t between 0 and 1; if
+    /// extrapolation is desired, see Extrapolate.
+    /// </summary>
+    public static Pose Interpolate(Pose a, Pose b, float t) {
```
Amarcolina: How about Lerp and LerpUnclamped to go along with how Vector3/Quaternion/friends do?

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
```diff
@@ -0,0 +1,167 @@
...
+    public override bool Equals(object obj) {
+      if (!(obj is Pose)) return false;
+      else return this.Equals((Pose)obj);
+    }
+    public bool Equals(Pose other) {
+      return other.position == this.position && other.rotation == this.rotation;
+    }
+
+    public override int GetHashCode() {
+      return this.position.GetHashCode() ^ this.rotation.GetHashCode() * 7;
```
Amarcolina: may I suggest the fancy Hash struct?
```
return new Hash() {
  position,
  rotation
};
```

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
```diff
@@ -0,0 +1,167 @@
...
+    public Vector3    position;
+    public Quaternion rotation;
+
+    public Pose(Vector3 position, Quaternion rotation) {
+      this.position = position;
+      this.rotation = rotation;
+    }
+
+    public static Pose identity {
+      get { return new Pose(); }
```
Amarcolina: Watch out, when a struct is instantiated, all fields are initialized to their default values.  So the rotation is initialized to the default value of Quaternion.  Quaternion is also a struct, so the same rule applies and all of it's fields (x, y, z, _and_ w) get initialized to zero, making the quaternion invalid instead of the identity quaternion.

I'd recommend something along the lines of 
`public static readonly Pose identity = new Pose(Vector3.zero, Quaternion.identity);`

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: argh. I couldn't find any documentation for those units so I honestly wasn't sure how PhysX was measured. I should probably switch this to return radians for compatibility's sake.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: OK, will use `RotatedBy`

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: yeah can do

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: woops good catch

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: oh neat. OK!

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: Because it uses Quaternion.Slerp; I'm worried that calling it Lerp will cause the assumption that it's also Quaternion.Lerp-ing

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: oh geez, good catch

Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/Anchors/AnchorableBehaviour.cs
nickjbenson: removed the else clause

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: yeah, that's fair. I could see both being utilized but I think activeInHierarchy is the better default; we can add another option later if we need it

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: removed

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: Actually, for the sake of not double-converting I'm going to leave it in degrees and just clarify the comment.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: Quaternion.AngleAxis will automatically normalize the provided axis vector anyway, so I simplified this to just not normalize the vector (so we don't double-calculate magnitude more than the AngleAxis function already would).

Clarified comment to expect angle measure in degrees.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: I generally prefer the From ordering because in "from B", I like to think of the "from" as implying the inversion of B.

To() is a totally reasonable way to think about it, though. I can add To() syntax.

Amarcolina: I'd probably lean towards only having one, just so we don't have the same concept being represented by multiple methods.  I'm a-ok with using From!

nickjbenson: nooow you tell me, after I've just gone through and added all the To's

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
Amarcolina: Ah I see hmmmmmm.  I'd argue that Interpolate doesn't imply spherical interpolation either :P

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
Amarcolina: Aw man, that's why I called it a discussion!

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: le done

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
nickjbenson: good idea

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: It is basically another way to express Quaternion.LookRotation, I just wanted a more verbose syntax for it.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
nickjbenson: done

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: modified comment to clarify that, even though it is now called Lerp, it in fact uses Quaternion.Slerp for the rotation component.

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: done

Assets/LeapMotion/Core/Scripts/DataStructures/Pose.cs
nickjbenson: done

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
Amarcolina: If that's the case, why not just use Quaternion.LookRotation instead of all this extra math?  Would also probably good to mention in the docs that it does the same thing as Quaternion.LookRotation

PR #901, opened 2017-11-03
Fix destroying buttons in the OnPress callback

nickjbenson:
This PR:

- [x] Adds tests for destroying/disabling buttons inside press and unpress callbacks
- [x] Checks the interaction lock state in those tests so a controller doesn't get its lock state stuck

Conceptually, it's a little bad that interaction controllers don't hold onto a reference to whatever is actually locking them. But this fix at least addresses the bug without changing the locking architecture.


Assets/LeapMotion/Modules/InteractionEngine/Testing/Scripts/Tests/InteractionButtonTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        GameObject.Destroy(button);
+      };
+
+      // Wait until the button is pressed.
+      int framesWaited = 0;
+      while (!pressed && framesWaited < WAIT_FOR_INTERACTION_FRAME_LIMIT) {
+        yield return null;
+        framesWaited++;
+      }
+      Assert.That(framesWaited != WAIT_FOR_INTERACTION_FRAME_LIMIT);
```
Amarcolina: Would be good to add in the optional string params to each of these asserts, that way if they fail sometime later it is easier to know what exactly is going wrong.  Currently, it will say something like 'expected true but was false', would be nice if it said something like 'recording failed to press button'.

Assets/LeapMotion/Modules/InteractionEngine/Testing/Scripts/Tests/InteractionButtonTests.cs
```diff
@@ -0,0 +1,165 @@
...
+
+      // Primary hover lock should NOT be active once the button is destroyed.
+      Assert.That(rightHand.primaryHoverLocked == false);
+
+      yield return wait(endingTestWait);
+    }
+
+    [UnityTest]
+    public IEnumerator CanDisableButtonWhenPressed() {
```
Amarcolina: Might be nice to somehow roll this test into CanDestroyButtonWhenUnpressed, since it is almost line-for-line exactly the same.  One thing I like to do is create an enum with a list of actions to perform, and parameterize the test by using the Values attribute, which will cause the test to be run multiple times with different arguments.

An example of possible syntax:
```
public IEnumerator PerformActionWhenUnpressed([Values(Action.Destroy, Action.Disable)] Action toTakeWhenUnpressed)
```

Amarcolina: Could even parameterize it further by having an enum for the callback to subscribe to as well, rolling all of these tests into a single method with some arguments.

Assets/LeapMotion/Modules/HierarchyRecording/Scripts/LeapRecording/VectorHandRecording/QuaternionCurves.cs
```diff
@@ -60,11 +60,17 @@ public struct QuaternionCurves {
...
-      Quaternion evaluated = new Quaternion(xCurve.Evaluate(time),
-                                            yCurve.Evaluate(time),
-                                            zCurve.Evaluate(time),
-                                            wCurve.Evaluate(time));
-      return Quaternion.Lerp(evaluated, evaluated, 1.0f);
+      var x = xCurve.Evaluate(time);
+      var y = yCurve.Evaluate(time);
+      var z = zCurve.Evaluate(time);
+      var w = wCurve.Evaluate(time);
+      if (x == 0f && y == 0f && z == 0f && w == 0f) {
```
Amarcolina: Would want to use something like Mathf.Approximately to compare these, because there are non-zero values that can still trip up Quaternion.Lerp because they are so small.

nickjbenson: Sure, I can add that, although the case was primarily meant to catch `default(Quaternion)` specifically.

Amarcolina: In fact instead of trying to match the exact behavior of Quaternion.Lerp, it might just be safer to calculate the magnitude of the quaternion manually, and do the normalization yourself.  That way you can be sure you never pass a quaternion to Lerp that Lerp is unable to normalize.

nickjbenson: Oh, well, upon some basic research that appears to be very straight-forward. TIL Quaternions are literally just four-dimensional vectors and normalizing them means exactly what you would expect. I'll do that instead.

PR #907, opened 2017-11-10
Add ArrayToString() extension for T[]s

nickjbenson:
When testing the Shuffle PR I was kind of surprised we didn't have a quick debug method for printing the contents of an array in one line, so here you go.

There are of course a couple of ways to expand upon the utility of this, like supporting string formatting parameters for float arrays and Vector3 arrays, and also supporting Lists and the like. This one's just quick and simple for arrays.

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
+++ b/Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
@@ -364,6 +364,28 @@ namespace Leap.Unity {
     }
     #endregion
 
+    #region Array Utils
+
+    /// <summary>
+    /// Prints the elements of an array in a bracket-enclosed, comma-delimited list,
+    /// prefixed by the elements' type.
+    /// </summary>
+    public static string ToArrayString<T>(this T[] arr) {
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -364,6 +364,28 @@ public static class Utils {
     }
     #endregion
 
+    #region Array Utils
+
+    /// <summary>
+    /// Prints the elements of an array in a bracket-enclosed, comma-delimited list,
+    /// prefixed by the elements' type.
+    /// </summary>
+    public static string ToArrayString<T>(this T[] arr) {
```
Amarcolina: How about using `IList<T>` instead of `T[]`, that way you catch arrays _and_ List<T> objects?  Or hell, how about IEnumerable<T> to catch everybody since this is going to allocate a bunch of stuff anyway?

nickjbenson: yeah that's a good idea. This allocates garbage like nobody's business anyway. I'll change it to IEnumerable

PR #908, opened 2017-11-10
Improve ImplementsInterface inspector experience

nickjbenson:
- [x] Now the object field for a MonoBehaviour marked "ImplementsInterface" will display the name of the interface that needs to be satisfied in the field type, as opposed to "MonoBehaviour."

- [x] ISupportDragAndDrop is now an combinable property drawer attribute. This allows you to do things when the user tries to drag and drop things into your drawn property's box.

- [x] You will get an "invalid" cursor when you try to drag an ineligible object into an ImplementsInterface property that doesn't implement the interface.

- [x] Make sure ScriptableObject support isn't broken.

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Attributes/CombinablePropertyAttribute.cs
+++ b/Assets/LeapMotion/Core/Scripts/Attributes/CombinablePropertyAttribute.cs
@@ -17,6 +17,8 @@ using System.Collections.Generic;
 
 namespace Leap.Unity.Attributes {
 
+  using UnityObject = UnityEngine.Object;
+
   public interface IPropertyConstrainer {
 #if UNITY_EDITOR
     void ConstrainValue(SerializedProperty property);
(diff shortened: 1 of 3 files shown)


Amarcolina:
A ScriptableObject that implements the interface causes an 'accepted' cursor to be chosen, despite the type not matching for a Component field.  I think there needs to be type check somewhere in the Find logic to make sure the object accepted matches the field type.

Amarcolina:
Also, wanted to discuss changing the ISupportDragAndDrop interface.  Currently the array of drag-n-drop objects are individually validated, and drag/drop is only supported if _all_ of the items pass the validity test, and then they are added one at a time.  I think that in order to have maximum flexibility, the interface would be better working with the array of objects directly.  That way if someone wants to accept a drag and drop even though only 2/3 dragged objects are valid, they are free to write that logic themselves.

PR #909, opened 2017-11-16
Adding application build definitions

Amarcolina:
This PR takes the concept of our existing package definition, and expands it to applications as well!  Features include:
 - Support for multiple scenes per-definition.  Define entire experiences with a single definition!
 - Support for multiple build targets per-definition.  Build an experience to multiple platforms with a single click!
 - Support for specifying build options, allowing you to turn on and off the various build features.  (Also provides an easy button to enable all 'development' options)
 - Just like the package system, generate menu items so that you can build from the task bar at the top instead of needing to root around in your project window.
 - Option to automatically tag each build with the current git hash.  Makes it easy to keep track of versions.

This also fixes an issue with the EnumFlags attribute.  Previously, the EnumFlags attribute would _only_ work if the flags followed exactly a 1, 2, 4, 8.... pattern.

To test:
 - [x] Make a build definition using the context menu 
 - [x] Add multiple scenes to the build definition, ensure they all are included in the build
 - [x] Generate a build dropdown, ensure it actually shows up in the build menu
 - [x] Try out Android and Windows build targets, make sure both work as expected
 - [x] Try out different build options, make sure things work as expected 
 - [x] Make sure the package system still works correctly with no regressions 
 - [x] Make sure the EnumFlags attribute works as expected


nickjbenson:
Create -> Build produces a new asset called "Build" but with a Package Name of "MyBuild" -- I think these two strings should match when a new asset is created to avoid confusion

nickjbenson:
Also took me a bit of poking to find out that `_generateBuildDropdown` means generate a menu item. I don't think of the main application menu as a "dropdown", even though technically I would grant that the menu items drop down.

Could we rename the property to, e.g., `_generateBuildMenuItem` ?

PR #916, opened 2017-11-21
Minor IE performance improvements

nickjbenson:
This PR:

- [x] `autoSyncTransforms` is set to false in `InteractionManager.Start()`. We need to verify that this doesn't cause any problems in the example scenes, but it does totally remove the performance drops that the auto-syncing was incurring.
- [x] Fixes garbage when constructing Maybe<T>s where T is a struct
- [x] Adds the IEquatable<Pose> interface implementation for Leap.Unity.Pose
- [x] Adds an optimization where we don't actually track the "pose last frame" on an InteractionBehaviour unless the application begins requesting that information for a given InteractionBehaviour regularly (20-frame timeout). Pose-last-frame is intended to be helpful data, not wasteful data -- a single-frame delay is a small price to pay to avoid the extra processing overhead per-object.
- [x] Adds additional ProfilerSamples to make future IE profile runs contain more detailed information

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
@@ -89,8 +89,12 @@ namespace Leap.Unity {
     /// will have a value.  If the value is null, this maybe will have no value.
     /// </summary>
     public Maybe(T t) {
-      hasValue = t != null;
-      _t = t;
+      using (new ProfilerSample("set hasValue")) {
+        hasValue = t != null;
+      }
+      using (new ProfilerSample("_t = t")) {
+        _t = t;
+      }
     }
 
     /// <summary>
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
@@ -16,7 +16,7 @@ namespace Leap.Unity {
     public static readonly NoneType None = new NoneType();
 
     public static Maybe<T> Some<T>(T value) {
-      return Maybe<T>.Some(value);
+      return new Maybe<T>(value);
     }
 
     public static void MatchAll<A, B>(Maybe<A> maybeA, Maybe<B> maybeB, Action<A, B> action) {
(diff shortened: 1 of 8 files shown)


Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
```diff
@@ -85,20 +85,27 @@ public struct Maybe<T> : IEquatable<Maybe<T>>, IComparable, IComparable<Maybe<T>
...
 
     /// <summary>
-    /// Constructs a Maybe given a specific value.  This value needs to always be
-    /// non-null.
+    /// Constructs a Maybe given a specific value. This value needs to always be
+    /// non-null if the type is a reference type.
     /// </summary>
     public static Maybe<T> Some(T t) {
-      if (t == null) {
+      if (Type<T>.isValueType && t == null) {
```
Amarcolina: I think this needs to be `!Type<T>.isValueType`.  Currently the check is always false, since being a value type and being null are mutually exclusive.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -829,6 +832,9 @@ public enum GraspedMovementType {
 
     public Pose worldDeltaPose {
       get {
+        if (!_recordLastPose) _recordLastPose = true;
```
Amarcolina: Hmm, I don't think I agree with this optimization being so automatic.  This means that the first time the user requests the delta pose, it will be incorrect, since it was not being tracked.  And if the user has the misfortune of requesting the delta pose once every 21 frames, they will _never_ get anything but the identity pose.

If this is actually something that we want to optimize, I would take a more transparent approach and have worldDeltaPose simply throw an error if you access it without first enabling a flag manually.  That being said, is this something that is a significant performance hit?  It seems like the only work that is being bypassed is the creation and assignment of a pose struct, which should be very very fast (probably faster even than the profiler samples that wrap those operations).

nickjbenson: Could have been that the profiler samples were causing most of the overhead, but this feature isn't used _anywhere_ for any purpose right now -- and may wind up changing again entirely as `Pose` itself matures. Adding an unused measurement per-object is an anti-optimization even if it's relatively cheap, so I'm going to go with the script-only-flag strategy.

nickjbenson: (actually, just going to delete it)

PR #927, opened 2017-12-08
Support changing GR Spaces at runtime

nickjbenson:
This PR:

- [x] Adds support for changing the LeapSpace used by a GR at runtime. Just remove the old space component and add a new one. The GR will use the top-most Space component on its GameObject.

To test:

- [x] Make sure example scenes work
- [x] Do some weird stuff to make sure nothing unexpected was broken (e.g. copy/paste GRs, undo ops)


Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
```diff
@@ -217,6 +222,48 @@ public partial class LeapGraphicRenderer : MonoBehaviour, ISerializationCallback
...
+        GetComponents<LeapSpace>(spaces);
+        _space = spaces.Query().FirstOrDefault(s => s.enabled);
+      }
+      finally {
+        spaces.Clear();
+        Pool<List<LeapSpace>>.Recycle(spaces);
+      }
+
+      if (Application.isPlaying
+          && (origSpace != _space || (_space == null && !_lastSpaceWasNull))) {
```
Amarcolina: Hmmm why iss the _lastSpaceWasNull check needed at all?  Won't the originSpace check cover all possibilities?

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphic.cs
```diff
@@ -290,6 +290,13 @@ public abstract partial class LeapGraphic : MonoBehaviour, ISpaceComponent, ISer
       }
     }
 
+    /// <summary>
+    /// Called by graphic groups when a renderer's attached space changes.
+    /// </summary>
+    public void SetAnchor(LeapSpaceAnchor anchor) {
```
Amarcolina: I'd recommend changing the name of this method to something like `OnUpdateAnchor`, to make it more clear that it's not something that should be called by the user.

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
```diff
@@ -179,6 +179,11 @@ public partial class LeapGraphicRenderer : MonoBehaviour, ISerializationCallback
       }
     }
 
+    private void Update() {
+      // Validate the attached space to support it changing at runtime.
+      validateSpaceComponent();
```
Amarcolina: I'd recommend doing the validation inside of LateUpdate right before the GraphicRenderer does all the rest of it's work.  That way the user can't destroy or switch spaces between Update and potentially mess things up.

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
nickjbenson: No, because if the space is deleted, our reference to it will magically become null out from under our feet -- including the "cached" reference. :(

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphic.cs
nickjbenson: Yeah, can do. Happy to modify the pathway here further if you'd like, this was quick-and-dirty as a solution.

Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs
nickjbenson: Oh! Cool. Will make that change as well

Amarcolina:
I'm getting a null reference

```
NullReferenceException: Object reference not set to an instance of an object
Leap.Unity.GraphicalRenderer.LeapBakedRenderer.OnUpdateRenderer () (at Assets/LeapMotion/Modules/GraphicRenderer/Scripts/RenderingMethods/LeapBakedRenderer.cs:113)
Leap.Unity.GraphicalRenderer.LeapGraphicGroup.UpdateRenderer () (at Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicGroup.cs:335)
Leap.Unity.GraphicalRenderer.LeapGraphicRenderer.doLateUpdateRuntime () (at Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs:219)
Leap.Unity.GraphicalRenderer.LeapGraphicRenderer.LateUpdate () (at Assets/LeapMotion/Modules/GraphicRenderer/Scripts/LeapGraphicRenderer.cs:194)
```

Repro steps (using the baked renderer scene)
1) start game
2) add spherical space
3) remove spherical space
4) undo

PR #928, opened 2017-12-15
[IN PROGRESS]: Upgrading the LeapCSharp api

Amarcolina:
This is a work-in-progress PR that aims to upgrade the image api to the new version that uses far less memory and copies than the previous.  This is a change that *requires* the new service to use, so should not be merged/released until a new compatible service is available.

Changes:
 - LeapImageRetriever has been updated to use the new image API
 - LeapImageRetriever now only deals with a single image instead of separating raw and 'bright'
 - Deleted LeapDeviceInfo entirely!  Now you need to connect to the Controller to get info.
 - Deleted that weird class EyeType.  We use correct and non-hacky apis 'round here!
 - LeapEyeDislocator now grabs correct baseline from the current device, or uses a custom baseline the user can specify in the inspector
 - LeapEyeDislocator now uses the tilt and offsets in the XRServiceProvider so that the hands always line up with the image
 - LeapEyeDislocator now uses the correct StereoViewMatrix apis instead of doing weird hacky stuff in OnPreRender.  Now supports single pass stereo!
 - LeapCG.cginc has been modified to support single pass stereo.
 - Added the ability to profile the worker thread in the LeapCSharp dll
 - Added an OnDeviceSafe callback to the LeapServiceProvider for ease of getting access to leap devices
 - LeapServiceProvider no longer can create multiple controllers (createController is a no-op if a controller already exists)
 - Leap shaders have been moved into a Shaders folder instead of Resources so that they don't always get included in every single build
 - Added small attribute that allows you to control the name of a property in the inspector
 - Added TryPeek and a new TryDequeue overload to the ProduceConsume buffer

Things left to do:
 - [ ] Build a 32 bit version of the LeapC.dll
 - [ ] Build the android version of the LeapC.so
 - [ ] Build a version of LeapCSharp that allows debugging
 - [x] Implement a pooling strategy inside of LeapCSharp so we are not constantly allocating tons of memory (PR 8236 on platform)
 - [x] Stop hardcoding the baseline (currently hardcoded to 0.04 so things don't line up properly on some devices)
 - [x] Consider making this the PR where we finally get device info correct? (Pls?)
 - [x] Add diagnostic tools to the provider so we can test the efficiency of the new api
 - [x] Allow the device tilt params on the temporal warping script to affect the images as well as the hands


yeswalrus:
Just checked out to do some debugging with the latest LeapC/CSharp & I'm getting a bunch of errors with missing references to BeginProfilingForThreadArgs. any ideas?

Amarcolina:
Hmm, what version of Unity are you using, and what OS are you using?  Anything but Windows is going to give you trouble for the moment.

yeswalrus:
Fixed - needed to merge my platform branch with develop

nickjbenson:
@Amarcolina I would recommend that we merge the LeapCSharp source PR into this PR and then merge it into develop. That way we can dogfood it prior to the next release.

The PR already has LeapC for Windows x64; LeapC.so for Android, and LeapCs for other platforms will come down the pipe closer to release, and will be validated by QA at the latest when we cut a release branch.

PR #929, opened 2017-12-19
Upgrade 2017.3

Amarcolina:
This PR upgrades UnityModules for the newly released 2017.3 version of Unity.  The changes include:
 - Adding custom assemblies for Core as well as all Modules.  This will decrease compile time as well as make it impossible to violate our cross-module dependency rule!
 - Removing references to the now unsupported Dx9 API
 - Re-saving all scenes and assets (only minor changes here)

To test:
 - [x] Make sure that the project opens without warning or error
 - [x] Make sure all major scenes still function correctly
 - [x] Make sure scenes can still be built


nickjbenson:
Just pushed the fix for IE example 7. One-line; Physics.SyncTransforms() right after the Ship movement update.

PR #932, opened 2018-01-04
Adding a dedicated 'swap grasp' function 

Amarcolina:
The 'swap grasp' allows you to swap a currently grasped object for another one.  The newly grasped object should behave like you had been grasping it all along.  Similar conceptually to calling ReleaseGrasp and then TryGrasp, but does some extra stuff under the hood to make the transition smoother and more reliable (doesn't enable soft contact and manually copies classifier state to the new object).

To test:
 - [x] Make sure that basic scenes still function as expected (shouldn't be a problem, all added code is standalone)
 - [x] Test to see if SwapGrasp behaves as you would expect (you can check out `test-ie-swap-grasp` to give my test scene a try!)
 - [x] Verify that added code looks correct and follows IE conventions


nickjbenson:
Grasping the center objects with two hands appears to break the example scene. Does SwapGrasp support multigrasp situations? (It probably should, otherwise it should be explicitly forbidden in the documentation for it and throw an error if this case is attempted)

PR #941, opened 2018-01-23
Merge master into develop

nickjbenson:
The biggest change here is the addition of `XRSupportUtil` to provide a common interface to various XR/VR methods for both Unity 5.6 and Unity 2017+ (via ifdefs).


nickjbenson:
(John approved via Slack)

PR #948, opened 2018-01-26
Switch HandModelBase to use edit time frames

Amarcolina:
Now that a provider can provide actual edit time frames, this PR upgrades the HandModelBase to use those frames instead of constructing their own!  This simplifies the code as well as supporting edit time feedback when using the hierarchy recorder!

If there is no provider in the scene the hands will still be constructed on their own.

To test:
 - [x] Make sure that example scenes still function as expected in the editor (provider should control the pose of the hands)
 - [x] Make sure that a hand on it's own, when there is no provider in the scene, still renders correctly as expected
 - [x] Make sure this all works for both rigged hands and capsule hands


nickjbenson:
Nice, this is working for me. Yay consistent pipelines!

PR #949, opened 2018-01-27
Added null check to HandPool

Amarcolina:
Added a null check to the OnValidate callback of HandPool.cs.  This null check guards against errors that can occur if HandPool is added with an AddComponent call.  It also initializes the value to a non-null value to make the chance of a null reference even less.

To test:
 - [x] Make sure example scenes still work properly
 - [x] Make sure a new rig can be constructed from scratch
 - [x] Make sure that when you use a call to AddComponent to add a HandPool that no errors occur


nickjbenson:
Pinging this; if we're not going to make a hotfix we should close this (and make sure the fix is in `develop` but I think it is) -- otherwise we should raise to QA to try and get the patch tested and released.

Amarcolina:
I can't actually find the original email thread for this >.> 

It's not in master or develop as far as I know, this PR is the only Pr that actually has the changes.

Amarcolina:
But yeah, i'd like to just bite the bullet and merge this somewhere, or acknowledge that we are not going to have any bandwidth until the next version of core, and just release this into develop.

nickjbenson:
I would vote let's just get it into `develop`. We'll get a new Core release out soon(tm)

Amarcolina:
We can always to a cherry-pick hotfix if we _really really_ need to get it into master early

PR #950, opened 2018-02-01
Fixed failing template tests

Amarcolina:
The template for the BitConverter tests was showing up in the test window!  They were of course failing because they are templates and not actual tests, so I tweaked the template/generation to have the tests commented out, and get activated during generation.

Also added a small tweak to make sure that the AssetDatabase is correctly refreshed after generation happens.

To test:
 - [x] Make sure BitConverter tests pass
 - [x] Make sure that the _template_ tests no longer show up in the test window
 - [x] Make sure generation does not change the files

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Utils/BitConverterNonAlloc.cs
+++ b/Assets/LeapMotion/Core/Scripts/Utils/BitConverterNonAlloc.cs
@@ -9,7 +9,7 @@
 
 using System;
 
-namespace Leap.Unity.Tests {
+namespace Leap.Unity {
 
   public class BitConverterNonAlloc {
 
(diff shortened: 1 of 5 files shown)


Assets/LeapMotion/Core/Scripts/Utils/BitConverterNonAlloc.cs
```diff
@@ -9,7 +9,7 @@
 
 using System;
 
-namespace Leap.Unity {
+namespace Leap.Unity.Tests {
```
jselstad: Do we want the BitConverter itself in the Tests namespace?

Amarcolina: No we do not!  Fixed!

PR #951, opened 2018-02-05
Merge a number of utilities from the Paint project

nickjbenson:
This PR:

- [x] Puts another tier of organization on Utils.cs because it's growing quite large
- [x] Adds a few new things to Utils.cs -- if anything is missing a doc comment that is also non-obvious, I can address
- [x] Adds a QuickButton attribute, which is a CombinablePropertyAttribute that adds an inspector button, mainly intended for quick functionality testing or very common tasks related to a field (a custom editor is more appropriate for big-deal scripts)
- [x] Puts the RingBuffer/DeltaBuffer/DelayBuffer classes in the Leap.Unity namespace where they should be
- [x] Adds LeapColor.cs, which is just a bunch of static getters for named Colors
- [x] Adds the RunTimeOnly attribute, since we have an EditTimeOnly attribute (and is useful in some special-snowflake inspector scenarios)
- [x] Adds the IIndexable interface (which is Query-able), so anything that makes sense to be indexed can automatically get querying, and in the future help define UI-related components or other organizational components
- [x] Adds some older Slice and ReadonlySlice scripts, which don't allocate and allow you to pass small slices of larger collections around.

To test:
- [x] Mess with all the things!!!
- [x] Make sure builds still work


Assets/LeapMotion/Core/Scripts/DataStructures/Slice.cs
```diff
@@ -0,0 +1,153 @@
...
+
+    public static Slice<T> FromIndex<T>(this List<T> list, int fromIdx) {
+      return Slice(list, fromIdx);
+    }
+
+  }
+
+  public struct Slice<T> : IIndexable<T> {
+
+    private List<T> _list;
```
Amarcolina: List and arrays both inherit from the same interface IList<T> actually!  You can use that and not need to have a switch to determine the backing structure.

Assets/LeapMotion/Core/Scripts/DataStructures/IIndexable.cs
```diff
@@ -0,0 +1,62 @@
...
+      return new IIndexableEnumerator<T>(indexable);
+    }
+
+    public static QueryWrapper<T, IIndexableEnumerator<T>> Query<T>(this IIndexable<T> indexable) {
+      return new QueryWrapper<T, IIndexableEnumerator<T>>(GetEnumerator(indexable));
+    }
+
+  }
+
+  public struct IIndexableEnumerator<T> : IQueryOp<T> {
```
Amarcolina: If you modify the signiture for this a bit, you can have it be the enumerator for all things IIndexable, instead of needing to write a new enumerator for things like Slice and ReadonlySlice.

Something along the lines of
```
public struct IIndexableEnumerator<T, K> : IQueryOp<K> where T : IIndexable<K> {
  T indexable;
  ...
```

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+
+  public class QuickButtonAttribute : CombinablePropertyAttribute, IBeforeFieldAdditiveDrawer {
+
+    public const float PADDING_RIGHT = 12f;
+
+    public string label = "Quick Button";
+    public string methodOnPress = null;
+    public string tooltip = "";
+
+    public Type type;
```
Amarcolina: These have no need to be variables I think?  They are only used within the Draw method?

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+                     + "non-optional parameters.");
+        return;
+      }
+
+      if (GUI.Button(rect.PadInner(0, 0, 0, PADDING_RIGHT), new GUIContent(label, tooltip))) {
+        foreach (var target in targets) {
+          Undo.RegisterFullObjectHierarchyUndo(target, "Perform QuickButton Action");
+        }
+        foreach (var target in targets) {
+          method.Invoke(target, new object[] { });
```
Amarcolina: Would recommend a TryCatch around this and Debug.LogException any errors caught.  That way if one target throws an error, it doesn't prevent any of the other methods from being called.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+    public Type type;
+    public System.Reflection.MethodInfo method;
+
+    public QuickButtonAttribute(string buttonLabel, string methodOnPress, string tooltip = "") {
+      this.label = buttonLabel;
+      this.methodOnPress = methodOnPress;
+      this.tooltip = tooltip;
+    }
+
+    public float GetWidth() {
```
Amarcolina: I don't think this method is used?

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+          method.Invoke(target, new object[] { });
+        }
+      }
+
+    }
+#endif
+
+  }
+
+  public static class RectExtensions {
```
Amarcolina: Would these be better suited to the Utils file?

Assets/LeapMotion/Core/Scripts/Utils/HandUtils.cs
```diff
@@ -482,6 +482,30 @@ public static class HandUtils {
...
+    public static Hand Get(this LeapProvider provider, Chirality whichHand) {
+      List<Hand> hands;
+      if (Time.inFixedTimeStep) {
+        hands = provider.CurrentFixedFrame.Hands;
+      }
+      else {
+        hands = provider.CurrentFrame.Hands;
+      }
+
+      return hands.Query().FirstOrDefault(h => h.IsLeft == (whichHand == Chirality.Left));
```
Amarcolina: Could refactor this to use the extension method above, which targets a Frame object, to reduce code duplication

Assets/LeapMotion/Core/Scripts/Utils/RunTimeOnly.cs
```diff
@@ -0,0 +1,24 @@
...
+ * between Leap Motion and you, your company or other organization.           *
+ ******************************************************************************/
+
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+namespace Leap.Unity.Attributes {
+
+  public class RunTimeOnly : CombinablePropertyAttribute, IPropertyDisabler {
```
Amarcolina: It's standard convention to suffix attributes with the word 'Attribute', so to make it `RunTimeOnlyAttribute`.  It still lets you type `[RunTimeOnly]` when you use it though.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public float GetWidth() {
+      return GUI.skin.label.CalcSize(new GUIContent(label)).x + 12f + PADDING_RIGHT;
+    }
+
+#if UNITY_EDITOR
+    public void Draw(Rect rect, SerializedProperty property) {
+      
+      type = targets.Query().FirstOrDefault().GetType();
+      method = type.GetMethod(methodOnPress, System.Reflection.BindingFlags.Instance
```
Amarcolina: GetMethod can throw if there is an Ambiguous Match (more than one method with the same name), might be good to get all the methods and filter so you can throw a more descriptive error message.

https://msdn.microsoft.com/en-us/library/05eey4y9(v=vs.110).aspx

Assets/LeapMotion/Core/Scripts/DataStructures/DeltaBuffer.cs
```diff
@@ -8,108 +8,173 @@
...
+    /// change over time. A buffer size of 5 is a good start for 60-90 Hz updates.
+    /// </summary>
+    public override Vector3 Delta() {
+      if (Count == 0) { return Vector3.zero; }
+
+      Vector3 deltaPerTimeSum = Vector3.zero;
+      for (int i = 0; i < Count - 1; i++) {
+        deltaPerTimeSum += (Get(i + 1) - Get(i)) / (GetTime(i + 1) - GetTime(i));
+      }
+      return deltaPerTimeSum / Count;
```
Amarcolina: Is this intended that the logic changed from dividing by `length - 1`, to just dividing by `Count`?  As far as I can see Length was just renamed to Count, so why did this logic change?

nickjbenson: Ah, woops, good catch, I didn't look at this one closely enough. I need to be returning zero if the buffer is empty _or_ if the buffer contains only one sample, which will then allow me to properly do "Count - 1" to account for the fact that each delta change is a comparison of two consecutive samples. Will fix.

Assets/LeapMotion/Core/Scripts/DataStructures/ReadonlySlice.cs
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+
+    public static ReadonlySlice<T> FromIndex<T>(this ReadonlyList<T> list, int fromIdx) {
+      return ReadonlySlice(list, fromIdx);
+    }
+
+  }
+
+  public struct ReadonlySlice<T> {
```
Amarcolina: Any reason this doesn't implement the IIndexable interface?

nickjbenson: Yes, it was written before IIndexable existed and I forgot to add it :P

Assets/LeapMotion/Core/Scripts/DataStructures/Slice.cs
```diff
@@ -0,0 +1,153 @@
...
+      _beginIdx = 0;
+      _endIdx = arr.Length;
+      _direction = _beginIdx <= _endIdx ? 1 : -1;
+
+      _isArr = true;
+    }
+
+    public T this[int index] {
+      get {
+        if (index > Count - 1) { throw new System.IndexOutOfRangeException(); }
```
Amarcolina: If you are gonna have OutOfRange exceptions it probably would be best to cover both sides.

Assets/LeapMotion/Core/Scripts/Utils/LeapColor.cs
```diff
@@ -0,0 +1,174 @@
...
+
+    public static Color lavender {
+      get { return new Color(0xB5 / 255f, 0x7E / 255f, 0xDC / 255f); }
+    }
+
+    #endregion
+
+    #region Shorthand
+
+    private static Color Lerp(Color a, Color b, float amount) {
```
Amarcolina: Nitpick, but this should be a lowercase method name.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -898,88 +824,96 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+    /// the max movement amount from the original position.
     /// </summary>
-    public static Quaternion FaceTargetWithoutTwist(Vector3 fromPosition,
-                                                    Vector3 targetPosition,
-                                                    Vector3 upwardDirection,
-                                                    bool flip180 = false) {
-      Vector3 objToTarget = targetPosition - fromPosition;
-      return Quaternion.LookRotation((flip180 ? -1 : 1) * objToTarget,
-                                     upwardDirection);
+    public static Vector3 MovedTowards(this Vector3 thisPosition,
```
Amarcolina: So this looks like Vector3.MoveTowards?  If that is true, it probably would be best to use Vector3.MoveTowards under the hood as well.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -898,88 +824,96 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+      if (delta.sqrMagnitude > maxMovementAmount * maxMovementAmount) {
+        delta = Vector3.ClampMagnitude(delta, maxMovementAmount);
+      }
+      return thisPosition + delta;
     }
 
-    #endregion
-
-    #region Quaternion Utils
+    public static Vector3 ToVector3(this Vector4 v4) {
```
Amarcolina: There is already an implicit conversion from Vector4 to Vector3, so I don't think this method is necessary.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -998,168 +932,108 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
+      return Quaternion.LookRotation((flip180 ? -1 : 1) * objToTarget,
+                                     upwardDirection);
     }
 
-    /// <summary>
-    /// Additive Then syntax for floats. Literally, thisFloat + otherFloat.
-    /// </summary>
-    public static float Then(this float thisFloat, float otherFloat) {
-      return thisFloat + otherFloat;
+    public static Vector3 GetRight(this Quaternion q) {
```
Amarcolina: I'm not convinced that we need dedicated extension methods on a Quaternion to do a rotation like this.  I think it would need a more details name to be obvious what it actually does, by which time you might as well be writing the original code in the first place.

Assets/LeapMotion/Core/Scripts/DataStructures/IIndexable.cs
nickjbenson: Turned out to be even simpler than that! Pushed a change that just uses IIndexable's Query/Enumerator support to support Slice and ReadonlySlice

nickjbenson: The problem I hit with <T, K> is that that approach destroys type argument inference, which is really important for having the API be nice -- I tried it and lost the ability to just call `GetEnumerator()` and `Query()`. Turns out I was able to just use a single type argument for the element though

Amarcolina: New interface is gonna create garbage every time you call GetEnumerator though since it will box the slice (since slices are structs) when it references it by the IIndexable interface.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+                                             | System.Reflection.BindingFlags.Public
+                                             | System.Reflection.BindingFlags.NonPublic);
+
+      if (method == null) {
+        Debug.LogError("QuickButton tried to prepare " + methodOnPress + " for calling, "
+                     + "but the type " + type.Name + " has no such method.");
+        return;
+      }
+
+      if (method.GetParameters().Query().Any(p => !p.IsOptional)) {
```
Amarcolina: Also might want to check if the method contains generic parameters.  Currently trying to invoke GenericMethod<T>() throws an error.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
Amarcolina: Actually looks like they might already be in Utils!  Should use those and remove these.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
nickjbenson: le done

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
nickjbenson: good idea

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
nickjbenson: IBeforeFieldAdditiveDrawer uses it to determine the button width -- added a comment.

Assets/LeapMotion/Core/Scripts/Attributes/QuickButtonAttribute.cs
nickjbenson: Yes indeedy! Thanks for catching

Assets/LeapMotion/Core/Scripts/Utils/HandUtils.cs
nickjbenson: refactore'd

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: yup dunno why I didn't do that

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: In a situation where I'm going back and forth between Vector3s and Vector4s (say, for a normal-calculation algorithm on a mesh), I like the conversions to be explicit. Do you mind if we keep this? I'll add a comment that says that an implicit version exists so there's no confusion

nickjbenson: (This is particularly relevant if I'm doing a bunch of sequential calls like `MovedTowards(bluh).ToVector4().MyCustomVector4Extension().ToVector3().RotatedBy()...`)

Assets/LeapMotion/Core/Scripts/DataStructures/IIndexable.cs
```diff
@@ -0,0 +1,85 @@
...
+    public static QueryWrapper<T, IIndexableEnumerator<T>>
+                    Query<T>(this IIndexable<T> indexable) {
+      return new QueryWrapper<T, IIndexableEnumerator<T>>(GetEnumerator(indexable));
+    }
+
+  }
+
+  public struct IIndexableEnumerator<Element> : IQueryOp<Element> {
+
+    IIndexable<Element> indexable;
```
Amarcolina: Referring to an indexable by interface is fine, _unless_ the indexable is a struct, which will cause it to be boxed and allocate garbage.  This will happen with the slices (and possibly the array indexable as well?)

nickjbenson: Ohhh man... is this why you were thinking about having a second generic argument that was something like `K : IIndexable<T>`? I might need to give that another go in the name of remaining allocation-free

Amarcolina: Yeah that would be the idea, but I also tried to get it to work and couldn't get it to infer the types properly :/ Not 100% sure what to do about that.

PR #959, opened 2018-02-14
Added FindAllAssetsOfType to fix weird bug

Amarcolina:
Apparently `Resources.FindObjectsOfTypeAll` when used to find _assets_ will only find them if they are 'loaded' into memory.  So previously things like 'Run All Generators' or 'Build All Apps' could fail to target specific instances if they had not been loaded into memory yet.  This adds a util method to the existing EditorResources class that is specifically for finding _assets_ of a given type, and should be much more robust.

To test:
 - [x] Verify that upon opening Unity, you can immediately run all generators, or build all apps, and that it doesn't miss anything.


nickjbenson:
Unitys hate this one weird bug...

will check it out after our folder namoganza PR!

nickjbenson:
I was able to run all generators and I _think_ it worked? But it's hard to tell because there is no feedback, and no code changes from the generation results to check via git diff.

Maybe we should add a console message like, "Successfully generated N files"?

Meanwhile, currently "Build/All Apps" appears to do nothing in my editor.

Amarcolina:
Good idea, and also yeah build/all apps does nothing because you probably have no app defs in your project :D It sounds like it would be good to gray out that field if there are no apps to build

PR #961, opened 2018-02-17
Tune Grab and Soft-Contact; add support back to 5.6

jselstad:
- Enhancement: Tune the Grab Classifier to work better with small objects
- Enhancement: Add code support going back to 5.6 (except in HierarchyRecorder)
- Bug: Fix grab classifier allowing immediate regrabs when the object was forcefully let go
- Bug: Fix Soft Box -> Object collisions and spurious/anomalous non-deterministic behavior of soft contact.

- [x]  Test picking up small objects. 
- [x]  Test opening the project in Unity 5.6 (or any of the intermediate versions)
       (Note: We don't support opening UM-the-project in 5.6, but we support exported packages from 2017.3 being imported into 5.6, so this is what I'll test --Nick)
- [x]  Test two-handed grab repositioning; make sure it's not happening
- [x]  Observe difference in behavior in spheres resting on palm before and after this change


Assets/LeapMotion/Core/Scripts/LeapXRServiceProvider.cs
```diff
@@ -8,7 +8,11 @@
  ******************************************************************************/
 
 using UnityEngine;
+#if UNITY_2017_2_OR_NEWER
```
nickjbenson: We should avoid having any #ifdefs to support XR/VR across Unity versions outside of the dedicated file for it: XRSupportUtil. So no script other than it should need `using UnityEngine.XR` or use an XR/VR namespace.

I was holding off on adding support for prior Unity versions to `develop` so that we could make all of the changes at once before a release cycle goes through, but if we're going to do it in this PR, we should do it in a way that doesn't pollute our code with #ifdefs all over. :)

Assets/LeapMotion/Core/Scripts/Utils/Infix.cs
```diff
@@ -73,13 +73,18 @@ public static class Infix {
     /// Infix sugar for Vector3.SignedAngle(a, b).
     /// </summary>
     public static float SignedAngle(this Vector3 a, Vector3 b, Vector3 axis) {
+#if UNITY_2017_1_OR_NEWER
       return Vector3.SignedAngle(a, b, axis);
+#else
+      Debug.LogError("SignedAngle not supported in Unity versions before 2017.1");
```
nickjbenson: We should provide a custom implementation of SignedAngle instead of just logging an error, since our code may come to rely on it as well.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -896,10 +896,16 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
     #region Matrix4x4 Utils
 
     public static Matrix4x4 CompMul(Matrix4x4 m, float f) {
+#if UNITY_2017_1_OR_NEWER
       return new Matrix4x4(m.GetColumn(0) * f,
                            m.GetColumn(1) * f,
                            m.GetColumn(2) * f,
                            m.GetColumn(3) * f);
+#else
+      Debug.LogError("CompMul not supported in Unity versions before 2017.1");
```
nickjbenson: As for SignedAngle, we should implement the function.

Assets/LeapMotion/Core/Scripts/XR/XRHeightOffset.cs
```diff
@@ -101,11 +103,21 @@ public class XRHeightOffset : MonoBehaviour {
           autoRecenterOnUserPresence = false;
         }
       }
+#else
+      var trackingSpaceType = UnityEngine.VR.VRDevice.GetTrackingSpaceType();
```
nickjbenson: As for LeapXRServiceProvider, we should avoid having ifdefs and use XRSupportUtil so our cross-version support is cleaner.

Assets/LeapMotion/Experimental Modules/HierarchyRecording/Scripts/HierarchyRecorder.cs
```diff
@@ -90,7 +94,11 @@ public enum RecordOn {
     }
 
     protected void LateUpdate() {
-      if (XRDevice.isPresent && XRDevice.userPresence == UserPresenceState.Present && !_isRecording && (recordWhen & RecordOn.HMDPresence) != 0) {
+      if (!_isRecording && (recordWhen & RecordOn.HMDPresence) != 0 
+#if UNITY_2017_2_OR_NEWER
```
nickjbenson: as above, use XRSupportUtil

Assets/LeapMotion/Internal Modules/RealtimeGraph/Scripts/RealtimeGraph.cs
```diff
@@ -300,10 +300,14 @@ public enum GraphMode {
         AddSample("Render Delta", GraphUnits.Miliseconds, _renderTicks);
 
         float gpuTime;
-#if UNITY_5
-        gpuTime = UnityEngine.VR.VRStats.gpuTimeLastFrame;
-#else
+#if UNITY_5_6_OR_NEWER
+#if UNITY_2017_2_OR_NEWER
         UnityEngine.XR.XRStats.TryGetGPUTimeLastFrame(out gpuTime);
+#else
```
nickjbenson: as above, use XRSupportUtil

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionController.cs
```diff
@@ -1255,6 +1255,9 @@ public abstract class InteractionController : MonoBehaviour,
 
         if (_softContactCollisions.Count > 0) {
           _disableSoftContactEnqueued = false;
+          if (_delayedDisableSoftContactCoroutine != null) {
+            manager.StopCoroutine(_delayedDisableSoftContactCoroutine);
```
nickjbenson: Suuuper interesting. Is _this_ the line that seemingly fixes the soft contact weirdness? What's your understanding of how it actually fixes the issue?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionManager.cs
```diff
@@ -345,6 +345,14 @@ public class InteractionManager : MonoBehaviour, IInternalInteractionManager,
           // (This will clear softContacts and originalVelocities as well.)
           using (new ProfilerSample("Apply Soft Contacts")) {
             if (_softContacts.Count > 0) {
+              RuntimeGizmoDrawer drawer;
+              if (_drawControllerRuntimeGizmos && RuntimeGizmoManager.TryGetGizmoDrawer(out drawer)) {
```
nickjbenson: The InteractionManager is already an IRuntimeGizmoComponent, so it has an OnDrawRuntimeGizmos method; we should do all runtime gizmo drawing there so it's all in one place. I don't have anything against TryGetGizmoDrawer, but I don't want to mix the two gizmo drawing patterns.

Amarcolina:
> Bug: Fix Soft Box -> Object collisions and spurious/anomalous non-deterministic behavior of soft contact.
How exactly would you recommend going about testing this?  I would definitely be interested in trying a before/after test, I just am not 100% sure what to look for.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionController.cs
jselstad: Well, the delayed soft contact disabling routine would get spuriously called between 1 and 3 times before this; after this it seems to only ever execute once.

jselstad: As far as the root cause of the issue... I don't fully know.   To @Amarcolina 's question, I had been testing the soft contact by disabling the disable call to it, but it seems like the bugs I was getting only appeared when I disabled Soft Contact's ability to disable itself (and go away when it's allowed to enable and disable itself...)

jselstad: (One hypothesis is that the velocity queries get corrupted when the soft contact colliders turn to trigger colliders)

Assets/LeapMotion/Core/Scripts/Utils/XRSupportUtil.cs
```diff
@@ -24,50 +24,69 @@ namespace Leap.Unity {
...
+        outputPresenceWarning = true;
+      }
       #else
-      return XRDevice.isPresent;
+      if (!outputPresenceWarning){
+        Debug.LogWarning("XR UserPresenceState is only supported in in 2017.2 and newer.");
+        outputPresenceWarning = true;
+      }
       #endif
+      return false;
```
Amarcolina: Might be worth discussing whether or not this method should assume that 'unsupported' == 'false'.  This might cause certain behaviors to _never_ trigger because they are waiting for user presence.  Maybe defaulting to _true_ might be a good default, possibly with the option to pass in an optional param that specifies the value to return if the feature is unsupported?

Assets/LeapMotion/Core/Scripts/XR/Editor/XRHeightOffsetEditor.cs
```diff
@@ -43,8 +48,12 @@ public class XRHeightOffsetEditor : CustomEditorBase<XRHeightOffset> {
     //}
 
     private bool isRoomScaleTrackingDetected() {
```
Amarcolina: Any reason this still needs to exist?  Should this not be replaced with XRSupportUtils.IsRoomScale ?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionController.cs
Amarcolina: What does this piece of code _do_ exactly?  It seems to be called from within `fixedUpdateSoftContact`, does that mean it's being triggered every frame?

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -896,10 +896,19 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
...
     public static Matrix4x4 CompMul(Matrix4x4 m, float f) {
+#if UNITY_2017_1_OR_NEWER
       return new Matrix4x4(m.GetColumn(0) * f,
                            m.GetColumn(1) * f,
                            m.GetColumn(2) * f,
                            m.GetColumn(3) * f);
+#else
+      Matrix4x4 toReturn = m;
+      for (int i = 0; i < 4; i++) {
+        toReturn.setColumn(toReturn.GetColumn(i) * f);
```
Amarcolina: Should read `toReturn.SetColumn(i, toReturn.GetColumn(i) * f);`

nickjbenson:
Could you validate if the grasp repositioning fix is relevant to the behaviour I'm seeing here?

Multi-grasp is disabled, but when I grasp with a new hand, the old hand's fingertips appear to affect the object still when the object comes into contact with it.

https://drive.google.com/file/d/0B_w9_hQJupIYTGNuSlgwQW92Ukk/view?usp=sharing

nickjbenson:
I think this looks almost ready, I'd just like validation on the grasp behavior I was seeing. Willing to  trade for a ticket and another future PR-fix IFF it's an unrelated issue.

PR #966, opened 2018-02-27
Fix dictionary and hashset serializing

### LINKED ISSUE #893: AnchorGroup/SerializableHashset throws IndexOutOfRangeException
First error:
IndexOutOfRangeException: Array index is out of range.
System.Collections.Generic.HashSet`1[T].Add (.T item)
Leap.Unity.SerializableHashSet`1[T].OnAfterDeserialize () (at Assets/LeapMotion/Core/Scripts/DataStructures/SerializableHashSet.cs:74)
UnityEditor.HostView:OnGUI()

and after that continuously:
IndexOutOfRangeException: Array index is out of range.
System.Collections.Generic.HashSet`1[T].GetLinkHashCode (Int32 index)
System.Collections.Generic.HashSet`1+Enumerator[Leap.Unity.Interaction.Anchor].MoveNext ()
Leap.Unity.SerializableHashSet`1[T].OnBeforeSerialize () (at Assets/LeapMotion/Core/Scripts/DataStructures/SerializableHashSet.cs:103)
UnityEditor.HostView:OnInspectorUpdate()

To reproduce I'd suggest trying the following (I'm not exactly sure adding a lot of Anchors is the only reason this happens)

1.) Open the Anchors scene of the InteractionEngine examples
2.) Duplicate an Anchor (e.g. the Red Vive Anchor) about 10-20 times
3.) Select an AnchorGroup (e.g. Red Anchor Group) and add all the duplicated Anchors

This should already log the error above. When building an application (and sometimes also when you enter/exit playmode) the changes made to the AnchorGroup are lost.

Amarcolina:
Unfortunately with the advent of .Net4.5, Unity can no longer properly serialize all of the private fields present in HashSet<> and Dictionary<,>, and so we need to resort to completely manual serialization.  SerializableHashSet and SerializableDictionary no longer inherit from HashSet and Dictionary directly, and instead embed a copy of the data structure inside.  That way serialization can be completely controlled.  

The classes have added some methods that mirror the methods present in the regular data structures so they can be used in basically the same way.  If you actually need a HashSet or a Dictionary, you can still cast and get the actual result.  But all the basic needs are covered, you can ask for count, add, remove, and foreach.

Now that the classes are not inheriting directly, we can have them inherit from a non-generic base, allowing the property drawers to be 'always-on' instead of requiring an attribute (which has been marked obsolete).

Fixes #893

To test:
 - [x] Verify no compiler errors in editor or in build
 - [x] Verify that things still work as expected in editor and in build

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/SerializableDictionaryEditor.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/SerializableDictionaryEditor.cs
@@ -14,7 +14,7 @@ using System.Collections.Generic;
 
 namespace Leap.Unity {
 
-  [CustomPropertyDrawer(typeof(SDictionaryAttribute))]
+  [CustomPropertyDrawer(typeof(SerializableDictionaryBase), useForChildren: true)]
   public class SerializableDictionaryEditor : PropertyDrawer {
 
     private ReorderableList _list;
--- a/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/SerializableHashSetEditor.cs
+++ b/Assets/LeapMotion/Core/Scripts/DataStructures/Editor/SerializableHashSetEditor.cs
@@ -15,7 +15,7 @@ using UnityEditorInternal;
 
 namespace Leap.Unity {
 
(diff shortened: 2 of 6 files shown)


nickjbenson:
Alright, looks good now!

PR #967, opened 2018-03-01
Merge Paint Things 2 (electric boogaloo)

nickjbenson:
This PR brings over some other Core changes that I've made incrementally as Paint work has accumulated, some newer, some older.

- [x] AttachmentPointBehaviour no longer whines when it sees an AttachmentPointFlags.None, it's a static utility method that's useful for using AttachmentPointFlags in other components and None can be totally valid.
- [x] Minor change to CombinablePropertyAttribute that breaks out some functionality as static to facilitate some custom attribute work in other custom editors. Also made ImplementsInterface friendlier when `new()`'d as a thing to use in a different custom editor that seeks to work similarly to it but that can't implement it directly due to generic typing. (Paint uses this for Streams, which will come down the experimental modules pipeline sometime soon probably.)
- [x] OnEditorChange detects superclass methods now
- [x] Maybe has ValueOr methods added for both a custom default and a Maybe-custom-default (the former is obvious, the latter is useful for e.g. overlaying optional overrides)
- [x] ReadonlyList was missing IndexOf and also now implements IEnumerable
- [x] RuntimeColliderGizmos can now optionally draw triggers, and don't draw if the component is disabled
- [x] Added view-aligned wire spheres to RuntimeGizmoManager, these look REALLY NICE when you're in a headset (obviously doesn't work in the scene view though). Still worth it!
- [x] Also added drawing Rects for RuntimeGizmos, also modified the look of DrawPosition slightly
- [x] Added some utilities for doing Quaternion compression and decompression for future networking applications
- [ ] Added Grasping callbacks: The old kind that including the object itself as an argument, which was actually useful for a Paint experiment and is useful in a manager-object situation, and added Controller grasping callbacks (why didn't we have these before??)
- [x] InteractionManager lost some primo whitespace somehow, this fixes that, please double-check these changes were actually whitespace-only
- [x] Added wasSlid and setters for horizontal and vertical SliderPercents to InteractionSlider
- [x] added Toggle() and Untoggle() methods to InteractionToggle

With apologies for being massive. Some of these changes had been lingering for a while. Better late than never!


Assets/LeapMotion/Core/Scripts/DataStructures/ReadonlyList.cs
```diff
@@ -40,8 +47,20 @@ public struct ReadonlyList<T> {
       return _list.GetEnumerator();
     }
 
+    IEnumerator<T> IEnumerable<T>.GetEnumerator() {
```
Amarcolina: If you use the List<T>.Enumerator you will be able to foreach a ReadonlyList without allocating garbage.

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -531,8 +532,65 @@ public class RuntimeGizmoDrawer {
...
+      PopMatrix();
+    }
+
+    public void DrawCameraAwareWireCircle(float radius,
+                                          Vector3 normal, Vector3 radialStartDir,
+                                          int numSegments = 32) {
+
+      var origColor = this.color;
+
+      var camPos = matrix.inverse.GetPose().Then(Camera.main.transform.ToPose()).position;
```
Amarcolina: I'd want this method to perform some partially correct action if there is no main camera, that way things don't throw a null reference for an unknown reason.  (Also might want to look at if Camera.current would work better instead?)

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -656,38 +729,115 @@ public class RuntimeGizmoDrawer {
       PopMatrix();
     }
 
+    public void DrawCollider(Collider collider, bool useWireframe = true) {
```
Amarcolina: DrawColliders should probably be refactored to use this method

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -656,38 +729,115 @@ public class RuntimeGizmoDrawer {
...
+        targetScale = overrideScale.Value;
+      }
+      else {
+        targetScale = 0.06f; // 6 cm at 1m away.
 
-      var mainCam = Camera.main;
-      var posWorldSpace = matrix * pos;
-      if (mainCam != null) {
-        float camDistance = Vector3.Distance(posWorldSpace, mainCam.transform.position);
+        var mainCam = Camera.main;
```
Amarcolina: Another main camera that should be validated (or maybe Camera.current)

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
```diff
@@ -547,6 +605,21 @@ public class RuntimeGizmoDrawer {
       PopMatrix();
     }
 
+    /// <summary>
+    /// Draws a wire gizmo circle at the given position, with the given radius;
+    /// the normal is assumed to be towards Camera.main.
+    /// </summary>
+    public void DrawWireCircle(Vector3 center, float radius) {
```
Amarcolina: The current DrawWireCircle that accepts a normal value will include a cross in the center (since it is a squashed wire sphere actually).  Since the method names are exactly the same, they should probably have the same appearance.  I'd recommend changing the existing one to use DrawWireArc as well.

(and this method should probably just calculate the circle normal and then use the existing DrawWireCircle)

Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/InteractionToggle.cs
```diff
@@ -131,5 +131,20 @@ public class InteractionToggle : InteractionButton {
     private void OnPressed() {
       isToggled = !isToggled;
     }
+
+    /// <summary>
+    /// Sets this InteractionToggle to the "toggled" state. Calling this function won't
+    /// oscillate the state of the toggle; to 'untoggle' the control, call Untoggle().
+    /// </summary>
+    public void Toggle() {
```
Amarcolina: Any reason for these to exist?  Doesn't `isToggled` already provide all of this functionality?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
```diff
@@ -357,6 +357,11 @@ public class InteractionBehaviour : MonoBehaviour, IInteractionBehaviour {
     /// If this event is fired on a given frame, it will occur after OnGraspEnd and before OnGraspStay.
     /// </remarks>
     public Action OnGraspBegin;
+    /// <summary>
+    /// Called when the object becomes grasped, if it was not already held by any interaction controllers on the
+    /// previous frame. Also passes itself as an event argument.
+    /// </summary>
+    public Action<InteractionBehaviour> OnObjectGraspBegin;
```
Amarcolina: Hmmm I'd be opposed to adding these extra callbacks unless there was really something you couldn't do without them.  There are already 2 different types of grasp callbacks, adding a third seems a little crowded.  

If you are subscribing to the event, you already know what object you are subscribing to, and could use that argument in your callback, right?

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionController.cs
```diff
@@ -196,6 +196,21 @@ public abstract class InteractionController : MonoBehaviour,
     /// </summary>
     public Action<InteractionBehaviour> OnStayPrimaryHoveringObject = (intObj) => { };
 
+    /// <summary>
+    /// Called when the InteractionController begins grasping an object.
+    /// </summary>
+    public Action OnGraspBegin = () => { };
```
Amarcolina: It doesn't look like these are called properly during a SwapGrasp

Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
```diff
@@ -23,8 +23,14 @@ public class CombinablePropertyDrawer : PropertyDrawer {
...
     private List<CombinablePropertyAttribute> attributes = new List<CombinablePropertyAttribute>();
     private void getAttributes(SerializedProperty property) {
-      if (!_cachedAttributes.TryGetValue(fieldInfo, out attributes)) {
-        attributes = new List<CombinablePropertyAttribute>();
+      GetAttributes(property, fieldInfo, ref attributes);
+    }
+
+    public static void GetAttributes(SerializedProperty property,
+                                     FieldInfo fieldInfo,
+                                     ref List<CombinablePropertyAttribute> outAttributes) {
```
Amarcolina: this can be an out property instead of a ref.  The field is always assigned to so it would make more sense.

Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
```diff
@@ -36,24 +42,35 @@ public class CombinablePropertyDrawer : PropertyDrawer {
...
-    public override void OnGUI(Rect position, SerializedProperty property, GUIContent label) {
+    public override void OnGUI(Rect position, SerializedProperty property,
+                               GUIContent label) {
       getAttributes(property);
 
+      CombinablePropertyDrawer.OnGUI(this.attributes, this.fieldInfo,
+                                     position, property, label);
+    }
+
+    public static void OnGUI(List<CombinablePropertyAttribute> attributes,
```
Amarcolina: Out of curiosity, what kind of work are you doing that requires calling into the OnGUI code of CombinablePropertyDrawer directly?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -55,16 +56,22 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
     /// the interface that this attribute constrains objects to, and returns the object
     /// that implements that interface, or null if none was found.
     /// </summary>
-    public UnityObject FindImplementer(UnityObject obj) {
+    public UnityObject FindImplementer(SerializedProperty property, UnityObject obj) {
```
Amarcolina: Why was the SerializedProperty argument added?  It isn't used anywhere?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -55,16 +56,22 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
...
 
-      if (!fieldInfo.FieldType.IsAssignableFrom(obj.GetType())
-          && !(typeof(Component).IsAssignableFrom(fieldInfo.FieldType)
-               && obj.GetType() == typeof(GameObject))) {
+      // Don't use fieldInfo here because it is very convenient for it to be left null
+      // on the CombinablePropertyAttribute when dealing with generic-type situations
+      // where the ImplementsInterface class has to be constructed manually in a custom
+      // editor. (Specific case: StreamConnectorEditor.cs)
+
+      if (!this.type.IsAssignableFrom(obj.GetType())
```
Amarcolina: This changes functionality.  Now when I drop a ScriptableObject that does implement the interface, onto a field that requires the interface but is of type Component, the drag-drop reports that the drop should succeed, but it doesn't.  In fact, probing the code shows that the scriptable object is indeed assigned to the serializedProperty, but luckily Unity throws no errors and takes no action.  But that is not something we can guarantee happening for all time (and in fact assigning an object of an invalid type would make sense for an error in the future!)

Assets/LeapMotion/Core/Scripts/Utils/RuntimeColliderGizmos.cs
```diff
@@ -16,10 +16,16 @@ public class RuntimeColliderGizmos : MonoBehaviour, IRuntimeGizmoComponent {
     public Color color = Color.white;
     public bool useWireframe = true;
     public bool traverseHierarchy = true;
+    public bool drawTriggers = false;
+
+    void Start() { }
```
Amarcolina: Why was the start callback added?

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: I didn't do that because it appeared to be strictly less efficient to do it that way, but happy to make the change anyway in the interest of avoiding code duplication

Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/InteractionToggle.cs
nickjbenson: Can't access properties in UnityEvents nor in tools like PlayMaker.

nickjbenson: As a practice, we should try to provide all API functionality as plain public methods for this reason. I don't see a way around it :\

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
nickjbenson: I added these because there was really something I couldn't do without them: I had a manager class that sought to handle grasp events for the objects it managed. This is very important for designs involving multi-object grasping interactions, or interactions where there are multiple handles for a single object, where the logic to handle grasps should still be centralized. Without the grasped object passed as an argument, it's impossible to know which managed object was grasped, unless I did something weird like scanned through all objects to look for a just-changed "isGrasped" property.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/InteractionToggle.cs
Amarcolina: UnityEvents can definitely access properties, and as far as I'm aware, so can PlayMaker?  This page I found shows that PlayMaker can use a SetProperty method in order to modify the property of an object.

image

Assets/LeapMotion/Core/Scripts/Utils/RuntimeColliderGizmos.cs
nickjbenson: to give the component a checkbox so it can be enabled and disabled.

Assets/LeapMotion/Core/Scripts/DataStructures/ReadonlyList.cs
nickjbenson: will fix

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: will fix

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
Amarcolina: Hmm if there is indeed a performance difference, I'd want to see if we could remove the duplication by another means (maybe by having a dedicated private method for each collider, so at least _that_ code isn't being duplicated?)

But I'd guess that the performance difference isn't going to be significant, especially compared to the overhead of actually drawing the gizmos in the first place.

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: Will look into; I thought Camera.current only has a setting when you're in a camera callback like OnPreCull but maybe gizmos are drawn during just such a callback?

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: Will look at adding code re-use, but I think DrawWireCircle should never draw a cross in the center because that's not what a wire circle is (this actually made DrawWireCircle unusable for me)

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
Amarcolina: Couldn't you do something like this?

```
foreach(var obj in allTheObjs) {
  obj.OnGraspBegin += () =>  onObjGrasped(obj);
}

private void onObjGrasped(InteractionBehaviour obj) { //stuff }
```

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
Amarcolina: Makes sense, in that case, maybe the original DrawWireCircle needs to be updated to use the new arc code?  That way everything is consistent.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionController.cs
nickjbenson: will fix

Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
nickjbenson: Will look into. I don't remember why I chose ref instead of out for this.

Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
nickjbenson: will investigate :D there waaaas almost certainly a good reason

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
nickjbenson: Hm, this might be a dangling argument then. Will fix if it's truly unused

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
nickjbenson: I knew this would break something when I changed it, I just couldn't remember what exactly. :P will fix!

nickjbenson: and will add a comment because that's what comments are for

Assets/LeapMotion/Core/Scripts/DataStructures/Maybe.cs
```diff
@@ -158,6 +157,35 @@ public struct Maybe<T> : IEquatable<Maybe<T>>, IComparable, IComparable<Maybe<T>
...
+    /// <summary>
+    /// Returns this Maybe if it has a value, otherwise returns the argument Maybe value.
+    /// Useful for overlaying multiple Maybe values.
+    /// For example, if I want to overlay a "maybe override font" variable with
+    /// another "maybe override font" variable, I can call:
+    /// this.font = other.font.ValueOr(this.font);
+    /// </summary>
+    public Maybe<T> ValueOr(Maybe<T> maybeCustomDefault) {
+      if (hasValue) {
+        return Some(_t);
```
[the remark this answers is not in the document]
nickjbenson: will fix

Assets/LeapMotion/Core/Scripts/Utils/RuntimeColliderGizmos.cs
Amarcolina: Ahh makes sense, might be good to add a comment in that case 👍

Assets/LeapMotion/Core/Scripts/Utils/RuntimeGizmoManager.cs
nickjbenson: Can do :+1:

Assets/LeapMotion/Modules/InteractionEngine/Scripts/UI/InteractionToggle.cs
nickjbenson: OK well you straight up got me there with UnityEvents, I must have mixed it in with my PlayMaker experience with Martin.

I definitely remember Martin showing me an inability to manipulate properties when we were doing some IE work, but maybe there's just a feature we missed!

........But I still like having this in the API. Yes, it's duplicate functionality, but it's also an extremely direct and obvious usage for a "Toggle" object. I'd much rather have isToggled and Toggle()/Untoggle() both so that a user can just type "uhhh 'toggle.Toggle()', oh good, that's a function" and have it work.

Amarcolina: That makes sense, alright, I agree with it staying in!

Assets/LeapMotion/Core/Scripts/DataStructures/ReadonlyList.cs
nickjbenson: hm OK. I wonder why I marked ReadonlyList<T> as IEnumerable since we know that allocates. /shrug

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionBehaviour.cs
nickjbenson: That.. is a good point.
It wouldn't work in a UnityEvent situation where you couldn't script up a lambda for it, and this does allocate-- but the allocation doesn't seem like a big deal.

I'll remove the callbacks.

Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
nickjbenson: Here's where I used it.
 
   private void drawStreamProperty(SerializedProperty property) {
      var rect = EditorGUILayout.GetControlRect(true);
      var label = new GUIContent("Stream ("
                                 + (target as StreamConnector).streamDataType.Name + ")");

      var streamType = (target as StreamConnector).streamType;

      var spoofedImplementsInterfaceAttribute
        = new ImplementsInterfaceAttribute(streamType);

      var attributes = Pool<List<CombinablePropertyAttribute>>.Spawn();
      attributes.Clear();
      try {
        attributes.Add(spoofedImplementsInterfaceAttribute);
        CombinablePropertyDrawer.OnGUI(attributes, null, rect, property, label);
      }
      finally {
        attributes.Clear();
        Pool<List<CombinablePropertyAttribute>>.Recycle(attributes);
      }
    }

nickjbenson: Whole file here:
https://github.com/leapmotion/LeapPaint/blob/master/Assets/AppModules/AssetTools/Streams/Editor/StreamConnectorEditor.cs

nickjbenson:
@Amarcolina Okay! I've done a full pass addressing the comments you made on the PR and fixed the merge and some indention issues with the Interaction Manager, so this is ready for another look when you have the time.

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -55,35 +57,78 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
...
+        else {
+          // First group is the whole match, second is the inner group, the class name.
+          var fieldTypeName = fieldTypeMatch.Groups[1].Value;
+
+          var fieldIsScriptableObject = fieldTypeName.Equals("ScriptableObject");
+          if (fieldIsScriptableObject && (objIsComponent || objIsGameObject)) {
+            isTypeDefinitelyIncompatible = true;
+          }
+
+          var fieldTakesComponent = fieldTypeName.Equals("MonoBehaviour")
```
Amarcolina: There are more ways for a field to accept a Component.  Really, the field could be anything that inherits from Component.  Behaviour is one that exists inside the UnityEngine namespace even, which is just a component that can enabled or disabled.

There is a lot of logic added to allow fieldInfo to be passed in as null, might it be easier to get your hands on the fieldInfo in question?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -55,35 +57,78 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
...
+        if (objIsComponent) {
+          isTypeDefinitelyIncompatible = false;
+        }
+
+        // However, you can't assign a ScriptableObject to a field expecting a Component,
+        // or vice-versa, no matter what.
+
+        // Matches the "Foo" in "PPtr<$Foo>" (as an inner group)
+        var fieldTypeMatch = Regex.Match(property.type, "PPtr<\\$(.+)>");
+        if (fieldTypeMatch.Groups.Count == 0) {
```
Amarcolina: The if clause here assigns 'true' to the variable `isTypeDefinitelyIncompatible`.  In the else clause, the variable is assigned to if the field is either takes a scriptable object, or of it takes a component.  But those are the only two possibilities, and so `isTypeDefinitelyIncompatible` will always be assigned to, making the 3 assignments above (line 69, 76, and 84) not necessary.

Assets/LeapMotion/Core/Scripts/Utils/RuntimeColliderGizmos.cs
Amarcolina: Bump?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
nickjbenson: >There is a lot of logic added to allow fieldInfo to be passed in as null, might it be easier to get your hands on the fieldInfo in question?

Yes it would! I was unable to actually get access to it. I made an attempt the first time to try and avoid having to get rid of passing it in from the custom property, then I tried a second time when I was addressing your earlier comment on this method in the PR. I just couldn't figure out how to get it.. Maybe I was just missing something simple?

Amarcolina: If this is inside a PropertyDrawer, you should already have access to the fieldInfo directly.  If this is inside a custom editor, you would have to get the fieldInfo manually with reflection like `target.GetType().GetField(property.name)`.

Assets/LeapMotion/Core/Scripts/Utils/RuntimeColliderGizmos.cs
nickjbenson: woops sorry added

PR #974, opened 2018-03-13
Fix playmode tests

Amarcolina:
Playmode tests got broken due to the code for them living inside of an ifdef that only gets enabled when we want to build tests.  This caused our refactoring to fail to find all of the relevant code, breaking the tests when the code was enabled again.

Furthermore, there are some tests that are not passing that I will need the help of @nickjbenson to fix.  It might be related to some of the interaction engine test prefabs, which have some unbound scripts that may or may not be important?

To test:
 - [ ] Playmode tests must compile and pass

### DIFF
--- a/Assets/LeapMotion/Modules/GraphicRenderer/Testing/Scripts/Tests/TestActualShaderOutput.cs
+++ b/Assets/LeapMotion/Modules/GraphicRenderer/Testing/Scripts/Tests/TestActualShaderOutput.cs
@@ -25,6 +25,7 @@ namespace Leap.Unity.GraphicalRenderer.Tests {
     /// the vertices of a graphic in all situations.
     /// </summary>
     [UnityTest]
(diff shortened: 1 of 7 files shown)


nickjbenson:
Alright, IE tests work for me now, I cleaned up a missing script reference in the test rig prefabs but the key fix appeared to be ensuring that the test recording didn't auto-play as soon as the test began.

nickjbenson:
However, GR play-mode tests are failing for me, so that task remains for this PR. @Amarcolina

Amarcolina:
What tests are failing for you?  All tests pass for me.  What version of Unity did you use for testing?

nickjbenson:
I used 2017.3.0f3. I'll run them again in the morning and report back.

Amarcolina:
Arg, I see, they are failing on and off for me, and different _editor window configurations_ are making the issue better or worse.  Arggg

PR #980, opened 2018-03-22
Develop

caibf:
Why the Interaction Engine document is not the lastest with the current asset package?

1. Drag the LeapCameraRig prefab into your scene: LeapMotion/Core/Prefabs. -- not found
2.Examples 1 through 5 can be played -- already 8 examples
..., etc


nickjbenson:
Hi @caibf,

Our documentation matches the currently available package on developer.leapmotion.com/unity .
Working in `develop` is a different story. We're working on improving our documentation system to handle our versioning more gracefully, but at the moment our latest `develop` work does not reflect what we publically support. The changes in `develop` are undergoing QA and we don't merge it until it's proven ready and stable.

PR #982, opened 2018-03-27
Overhaul query testing and fixed some bugs!

Amarcolina:
The query testing has been overhauled to test a much wider variety of values on all of the different operators available.  This includes multiple different lengths, empty queries, as well as queries where the backing array is significantly larger than the size of the array.

Through this extra testing, I also managed to uncover some subtle bugs in some of the new query operations!   Two occurred when the backing array was significantly larger than the size of the array, and another occurred with small or empty queries.  Now that we have much better coverage, errors like these should be harder to introduce!

To test:
 - [x] Verify code looks good
 - [x] Verify tests pass

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Query/Editor/QueryTests.cs
+++ b/Assets/LeapMotion/Core/Scripts/Query/Editor/QueryTests.cs
@@ -15,24 +15,23 @@ using NUnit.Framework;
 
 namespace Leap.Unity.Tests {
   using Query;
+  using System;
 
   public class QueryTests {
-    public int[] LIST_0 = { 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 9, 1, 900, int.MinValue, int.MaxValue };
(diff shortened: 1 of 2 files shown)


Assets/LeapMotion/Core/Scripts/Query/Editor/QueryTests.cs
```diff
@@ -63,42 +62,48 @@ public class QueryTests {
...
+    public void DistinctTest([ValueSource("list0")] QueryArg arg) {
+      Assert.That(arg.ToQuery().Distinct().OrderBy(t => t).ToList(), Is.EquivalentTo(
+                  arg.ToList().Distinct().OrderBy(t => t).ToList()));
     }
 
     [Test]
-    public void ElemenAtTest() {
-      Assert.AreEqual(LIST_0.ElementAt(3),
-                      LIST_0.Query().ElementAt(3));
+    public void ElemenAtTest([ValueSource("list0")] QueryArg arg, [Values(0, 3, 100)] int index) {
```
jselstad: Typo

PR #983, opened 2018-03-27
Replace LeapCSharp DLL with the actual LeapCSharp source code, bring tests back

nickjbenson:
- [x] LeapCSharp is dead
- [x] Long live LeapCSharp
- [x] Verify all the examples are working and that builds don't accumulate memory indefinitely
- [x] Sanity-check the LeapCSharp source and make sure nothing crazy is in there
- [x] Make sure the LeapCSharp tests pass. Some are Ignore'd, the reasons should be sensible

I may go through and confirm that we don't have any dangling unused code like old ObjectPool and PooledObject, but this is ready for review.


Assets/LeapMotion/Core/Plugins/LeapCSharp/.gitignore
```diff
@@ -0,0 +1,2 @@
+AssemblyInfo.cs
```
Amarcolina: Does this gitignore need to be here?  Those files are not needed anymore right?

nickjbenson:
meeergiiiing

PR #989, opened 2018-04-03
Introduce the "Post-Process Provider" Pattern

jselstad:
The old post process system was clever, but too limited in terms of its ability to change the number of hands or to selectively apply hand post-processes to only the graphical model (or only to the logical model).

To solve this, we now have a `PostProcessProvider` class, which acts as a base to any sort of hand post processes you might want to implement.   It takes in the output of another LeapProvider (either a Service Provider or _another PostProcessProvider_ and asks the developer to implement:
``` public override void ProcessFrame(ref Frame inputFrame) { }```

which will modify a copy of the incoming frame.   Once the frame is processed, any consuming class can access the modified data via the emitted DispatchUpdateEvents or through CurrentFrame/CurrentFixedFrame.   This means HandModelManagers, InteractionManagers, and more can simply hook into the Post-Process Provider instead of the original service provider.  

This PR includes an example scene demonstrating the simultaneous use of an "InertiaHands" post-process and a "ProjectionHands" post-process (similar to Extendo Hands™) on a trio of HandModelManagers.  These-post processes can optionally be used in parallel, or applied in sequence (for maximum processing).

- [x] Pattern is actually more powerful/desirable than the previous one.
- [x] Everything has the appropriate names/comments and is located in the appropriate directories.
- [x] Ensure nothing breaks (from the removal of the old post-process system).

### CODE REVIEW

**Comment by nickjbenson** [comment]
To-do's for me on this:

- [ ] Apply UM style
- [ ] Editor scripts (hide unnecessary edit-time pose field, have Input be first field for PostProcessProvider)
- [ ] other stylistic things probably

**Comment by nickjbenson** [comment]
Hmmm. I'm attempting to modify the Inertia post-process to be framerate-independent by looking at `TimeVisible` on the hand, but it looks like that field is giving terrible values. Going to merge `develop` into this PR branch and check against the latest service, otherwise we've got an issue to resolve 👀

File: Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+
+    [Tooltip("The LeapProvider whose output hand data will be copied, modified, and "
+           + "output by this post-processing provider.")]
+    [SerializeField]
+    [OnEditorChange("inputLeapProvider")]
+    protected LeapProvider _inputLeapProvider;
+    public LeapProvider inputLeapProvider {
+      get { return _inputLeapProvider; }
+      set {
+        validateInput();
```
Amarcolina: Doesn't make sense to call validateInput here, you want it at the _end_ of the method once _inputLeapProvider has been assigned.

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+          processUpdateFrame(_inputLeapProvider.CurrentFixedFrame);
+        }
+        #endif
+        return _cachedFixedFrame;
+      }
+    }
+
+    protected virtual void OnEnable() {
+      if (_inputLeapProvider == null && Hands.Provider != this) {
+        _inputLeapProvider = Hands.Provider;
```
Amarcolina: Maybe do a null check here with a nice error message?

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+          _inputLeapProvider.OnUpdateFrame += processUpdateFrame;
+        }
+      }
+    }
+
+    public enum DataUpdateMode { UpdateOnly, FixedUpdateOnly, UpdateAndFixedUpdate }
+    [Tooltip("Whether this post-processing provider should process data received from "
+           + "Update frames, FixedUpdate frames, or both. Processing both kinds of frames "
+           + "is only recommended if your post-process is stateless.")]
+    public DataUpdateMode dataUpdateMode = DataUpdateMode.UpdateOnly;
```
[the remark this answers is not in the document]
nickjbenson: >Within ProcessFrame there isn't really a way to know what kind of frame you are processing.

I discussed this with John while I was fixing the bug, there is a way to handle state correctly when consuming both events, you just use Time.inFixedTimeStep. I decided not to duplicate the functionality of Time.inFixedTimeStep and pollute the function signature with a boolean that only-matters-sometimes anyway.
https://docs.unity3d.com/ScriptReference/Time-inFixedTimeStep.html

Regarding the public field:
The example is constructed to ensure that delta times passed to the simulation step are always positive, so that the simulation doesn't break terribly. It won't be _correct_ necessarily, but it should still look feasible. (If you don't ensure positive timesteps, you get terrible breakage, which is why I put the check in there.) If changing the update mode to "both" _does_ make the simulation look much more incorrect, then I screwed something up and should address it as a bug.

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
nickjbenson: will take a look

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
nickjbenson: will take a look

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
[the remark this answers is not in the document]
Amarcolina: Hmmmm that makes sense, and good to know Time.inFixedTimeStep exists, I did not!  I'd still lean towards having the inertial provider be a correct implementation of a stated post process provider.  If it is supposed to serve as an example, we wouldn't want to encourage practices that could have subtle (or not subtle) errors right?

Assets/LeapMotion/Core/Scripts/Utils/PostProcessProvider.cs
nickjbenson: I'm not sure an error is warranted here. Sometimes a reference is just null because you're setting up your scene or you have dangling objects for works-in-progress. I modified this check to avoid calling Hands.Provider twice, but otherwise I would propose a silent fail since Hands.Provider isn't fool-proof as it is.

PR #990, opened 2018-04-05
VectorHand encoding -- and some encoding progress

nickjbenson:
Early stages for a Networking module that are being pulled in to support some dependencies in the Hierarchy Recorder.

I am not a huge fan of the random Encoding interfaces, they have not been fully fleshed out.

- [x] VectorHand should have functioning I/O
- [x] We should probably figure out whether we need the interfaces at all, I can just remove them because they probably are overkill right now? The idea is they can provide the common link for different types of hand encodings in a future NetworkedProvider.
- [x] The files are placed in Core/Scripts/Encodings because they probably are going to be dependencies for various modules including Networking and the Hierarchy Recorder.


nickjbenson:
Heyy what do you know, ReadBytes doesn't actually work. Fixin' soon. Don't.. merge.

image

PR #995, opened 2018-04-07
Add ScaleLastFingerBones option to RiggedHand, enabled by default

nickjbenson:
This is a problem I've experienced a lot because I have rather small hands. Our tracking pipeline actually does a decent job of shrinking the tracked finger bones to fit, and Deform Positions captures _almost_ all of the corresponding discrepancies in the model, except the last finger bones! This is because only the bases of bones are adjusted, which works to adjust bone lengths until you hit the last bone.

To fix this, I've added an option that is enabled by default, that works only when Deform Positions is also enabled but that _scales_ the last finger bone to match the tracked last-finger-bone lengths compared to that of the standard edit-time hand that the Rigged Hand was modeled on.

This does introduce a potential problem: We must take care that we maintain a "standard" hand around which hands are modeled, so that the scale factor is actually correct. For now, this is our single edit-time hand that is constructed via the TestHandFactory.
This feature is constructed such that the scale reference is simply calculated lazily from whatever hand is returned by default from the TestHandFactory.

To test:
- [x] Validate that, whenever the checkbox is checked during play-mode, the rigged hand's last finger bone actually scales to match the tracked hand tip position. This is best checked by rendering both Capsule Hands and the Rigged Hands at the same time.
- [x] Validate the code makes sense and can be maintained.

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
+++ b/Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
@@ -80,12 +80,25 @@ namespace Leap.Unity {
               var boneLen = boneVec.magnitude;
               var standardLen = s_standardFingertipLengths[(int)this.fingerType];
               var newScale = bones[i].transform.localScale;
-              newScale.x = boneLen / standardLen;
+              var lengthComponentIdx = getLargestComponentIndex(modelFingerPointing);
+              newScale[lengthComponentIdx] = boneLen / standardLen;
               bones[i].transform.localScale = newScale;
             }
           }
         }
       }
+    }
+
+    private int getLargestComponentIndex(Vector3 pointingVector) {
+      var largestValue = 0f;
+      var largestIdx = 0;
+      for (int i = 0; i < 3; i++) {
+        var testValue = pointingVector[i];
+        if (testValue > largestValue) {
+          largestIdx = i;
+        }
+      }
+      return largestIdx;
     }
 
     public void SetupRiggedFinger (bool useMetaCarpals) {
--- a/Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
+++ b/Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
@@ -12,19 +12,25 @@ using System.Collections;
 using Leap;
 
 namespace Leap.Unity {
-  /**
-* Manages the orientation of the bones in a model rigged for skeletal animation.
-* 
-* The class expects that the graphics model bones corresponding to bones in the Leap Motion 
-* hand model are in the same order in the bones array.
-*/
-  public class RiggedFinger : FingerModel {
 
-    /** Allows the mesh to be stretched to align with finger joint positions
-     * Only set to true when mesh is not visible
-     */
+  /// <summary>
+  /// Manages the position and orientation of the bones in a model rigged for skeletal
+  /// animation.
+  ///  
+  /// The class expects that the graphics model's bones that correspond to the bones in
+  /// the Leap Motion hand model are in the same order in the bones array.
+  /// </summary>
+  public class RiggedFinger : FingerModel {
+    
+    /// <summary>
+    /// Allows the mesh to be stretched to align with finger joint positions.
+    /// Only set to true when mesh is not visible.
+    /// </summary>
     [HideInInspector]
     public bool deformPosition = false;
+
+    [HideInInspector]
+    public bool scaleLastFingerBone = false;
 
     public Vector3 modelFingerPointing = Vector3.forward;
     public Vector3 modelPalmFacing = -Vector3.up;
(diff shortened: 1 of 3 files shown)


Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,87 @@
...
     }
+    
+    /// <summary> Backing store for s_standardFingertipLengths, don't touch. </summary>
+    private static float[] s_backingStandardFingertipLengths = null;
+    /// <summary>
+    /// Lazily-calculated fingertip lengths for the standard edit-time hand.
+    /// </summary>
+    private static float[] s_standardFingertipLengths {
+      get {
+        if (s_backingStandardFingertipLengths == null) {
```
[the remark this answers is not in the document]
nickjbenson: What value does that add? What if the consuming application never uses a RiggedFinger? In general, I prefer lazy initialization because you only ever pay for what you use. But I think it's an interesting discussion too.

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,87 @@
...
+              // This will only scale correctly if the model was constructed to match
+              // the standard "test" edit-time hand model from the TestHandFactory.
+              var boneTipPos = GetJointPosition(i + 1);
+              var boneVec = boneTipPos - boneRootPos;
+              var boneLen = boneVec.magnitude;
+              var standardLen = s_standardFingertipLengths[(int)this.fingerType];
+              var newScale = bones[i].transform.localScale;
+              newScale.x = boneLen / standardLen;
+              bones[i].transform.localScale = newScale;
+              if (this.fingerType == Finger.FingerType.TYPE_INDEX
```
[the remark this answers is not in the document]
nickjbenson: Woops! Will fix :D

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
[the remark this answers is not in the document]
Amarcolina: Value would be less variables, no need for a property, simpler in general.  Also the way C# works is that a class is not initialized unless it is used!  So in a way, you can use a static constructor as a lazy initializer for all static data of the class!

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,87 @@
...
+            if (i == 3 && scaleLastFingerBone) {
+              // Set fingertip base bone scale to match the bone length to the fingertip.
+              // This will only scale correctly if the model was constructed to match
+              // the standard "test" edit-time hand model from the TestHandFactory.
+              var boneTipPos = GetJointPosition(i + 1);
+              var boneVec = boneTipPos - boneRootPos;
+              var boneLen = boneVec.magnitude;
+              var standardLen = s_standardFingertipLengths[(int)this.fingerType];
+              var newScale = bones[i].transform.localScale;
+              newScale.x = boneLen / standardLen;
```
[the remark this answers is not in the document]
protodeep: Correct.  I believe that 'Model Finger Pointing' field is exposed specifically to support joint with different axes pointing down the bone.

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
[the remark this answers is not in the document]
nickjbenson: I still prefer lazy initialization because as a general strategy it's more robust to, for example, order-of-execution issues. We've also been burned in the past by Unity and C# "guarantees" :P 
But I moved this logic to a static initializer.

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,95 @@
...
+                           unitType: TestHandFactory.UnitType.UnityUnits);
+      for (int i = 0; i < 5; i++) {
+        var fingertipBone = testHand.Fingers[i].bones[3];
+        s_backingStandardFingertipLengths[i] = fingertipBone.Length;
+      }
+    }
+    /// <summary>
+    /// Fingertip lengths for the standard edit-time hand.
+    /// </summary>
+    private static float[] s_standardFingertipLengths {
```
Amarcolina: Why the getter, can't you just use the array directly?

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,95 @@
...
         }
       }
     }
+
+    private int getLargestComponentIndex(Vector3 pointingVector) {
+      var largestValue = 0f;
+      var largestIdx = 0;
+      for (int i = 0; i < 3; i++) {
+        var testValue = pointingVector[i];
+        if (testValue > largestValue) {
```
Amarcolina: you need an Abs here, in case the pointing axis points in a negative direction

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
nickjbenson: woop sorry yeah can change

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
nickjbenson: woops good catch, meant to do that but missed it in implementation

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -12,40 +12,90 @@
...
       }
     }
+
+    private int getLargestComponentIndex(Vector3 pointingVector) {
+      var largestValue = 0f;
+      var largestIdx = 0;
+      for (int i = 0; i < 3; i++) {
+        var testValue = pointingVector[i];
+        if (Mathf.Abs(testValue) > largestValue) {
+          largestIdx = i;
```
Amarcolina: Er I just realized this isn't a proper 'get largest component index' since `largestValue` is never assigned.  This is actually get the last component that is non-zero.  Which is _usually_ correct but if there are subtle deviations in the pointing vector can make the answer wrong.  

(Like a vector of 0.999, 0, 0.001)

nickjbenson: Woops! Nice catch. pushing fix

nickjbenson: pushed

PR #1000, opened 2018-04-11
Remove old apis from LeapCSharp

yeswalrus:
The upcoming release deprecates several fields in LeapC, and these edits are required to maintain consistency.


Amarcolina:
@yeswalrus in develop I can receive hands from the service, but after this change I cannot.  I believe this might have to do with the fact that removing the fields from LEAP_TRACKING_EVENT has desynced the event structs layout from the layout of LeapC.h.  Might this PR also need an update to the LeapC dll?

Amarcolina:
Only thing that works for _me_ is to move the .so file into the x86 folder.  @jselstad @testmasterbright is this solution acceptable to you / works for you as well?  Don't know if it will still work for different brands of phone.....

PR #1003, opened 2018-04-17
Packages: replace "Core" with "CoreWithAndroid"

nickjbenson:
This PR removes the older Core package that exluded our in-progress Android support and replaces it with the CoreWithAndroid package so that we can release Android support in Core.

Current issues:
- [x] The packages module is preserving both the "Core" and "Core With Android" options in the Build menu for my editor (restarting the editor does not address the problem). Some cache might need to be refreshed?
- [x] ~Core does not build (when imported from a package) in Unity `2017.1.1` due to a `LeapProfiling` dependency on something that was added after that version. We can address this now or later, but it's a blocker to testing Core as a UnityPackage on earlier versions of Unity.~ We'll address backwards compatibility later, so this is fine; for now, test on 2017.3.0f3.

To test:
- [ ] Export the Core package, and import it into a fresh project. Ensure it compiles.
- [ ] Open an example scene, target Android, and build. The build should complete successfully.
- [ ] Hands should track in a built example scene on a phone with the correct service.

This branch should reflect an older Android libLeapC and so it should work with our older service.


Amarcolina:
> The packages module is preserving both the "Core" and "Core With Android" options in the Build menu for my editor
Seems that the codegen does not run in all the cases it should.  In this case, deleting a definition does not trigger the codegen.  I went ahead and manually triggered a regen, so that should be fixed.  Also fixed an outdated folder so that the gen could actually succeed.  I'll make a different PR that fixes the codegen issue, since it's much less important than this.

PR #1004, opened 2018-04-17
Add more auto-recenter options to XRHeightOffset

nickjbenson:
This PR modifies XRHeightOffset:

- [x] Renames autoRecenterOnUserPresence to recenterOnUserPresence
- [x] adds recenterOnStart, true by default
- [x] adds recenterOnKey, false by default
- [x] adds a conditionally-drawn recenterKey with the default value KeyCode.R

To test:
- [x] Ensure recenters occur on application start and when the R key is pressed for a good sampling of the example scenes.


nickjbenson:
OK, from our discussion it looks like we'd prefer to have all defaults such that _no_ key gains functionality that wasn't opted-into by the developer. So I'm going to change the script defaults and prefabs to have the recenter on-key option disabled, then make sure the example scenes all override that option to be true so we will have recentering on-key.

- [ ] Disable recenter-on-key defaults in scripts and prefabs
- [ ] Make sure the prefabs look correct and are up-to-date
- [ ] Make sure recenter-on-key is a scene override on the rig prefabs for all example scenes

PR #1010, opened 2018-04-20
[In progress] Release 4.4.0

Amarcolina:
This is the PR for release 4.4.0!  ~We are still waiting for some stuff, so this is in progress and doesn't need testing yet.~ Nick B Edit: essential features are in place. New merges may come in if there's time for issues to be addressed before the rubber stamp. See below for release notes.

EDIT: Nick B: Placing Changelog here.

Current release notes as of 4/23:

# UnityModules Core 4.4.0 Changelog

- Supported Unity versions: 5.6.2, 2017.1-4, 2018.1.

## VR Rig Simplification
- The number of scripts required to construct a Leap Motion-enabled VR rig has been greatly reduced, and the required rig hierarchy has been heavily simplified. Hand data that is adjusted correctly for XR headset and device latency can now be gotten by adding a single component to the Main Camera, and the hand model pipeline has been simplified to a single manager component driving managed model components. If you are upgrading from an older project that incorporated Leap rigs from Core 4.3.4 or earlier, you can auto-upgrade these rigs by opening the scene that contains them and checking the new Leap Motion SDK window: Window->Leap Motion.

## Leap SDK
- Added the Leap Motion SDK Window, accessible via Window->Leap Motion. Use this window to scan and upgrade old (Core 4.3.4 and earlier) Leap rigs, check settings for the Interaction Engine, and adjust module preferences for the Graphic Renderer.
- The LeapCSharp source is now provided directly in UnityModules and no longer hidden behind a managed DLL. It has received some upgrades in the process, and although it's still a little rough around the edges, it is a useful reference if you would like to construct LeapC bindings for a new language.
- Added VectorHand, one of many potential lightweight encodings of a Leap Hand. This lossy encoding is suitable for lightweight recording and playback or network transmission.
- Fixed a bug where hand.TimeVisible was not being set with the correct units.
- Added the PostProcessingProvider base class, a pattern for applying your own post-processes to Leap frame data.
- Added an example scene demonstrating a stateless and a stateful PostProcessProvider: ProjectionPostProcessProvider and InertiaPostProcessProvider.

## Utilities
- Query system: Overhauled to remove reliance on type-generic structs to support IL2CPP.
- BitConverter no longer requires unsafe code EXCEPT when targeting IL2CPP. If you wish to compile using IL2CPP, you must allow unsafe code.
- Added an ArrayPool data structure, used by the overhauled Query system.
- Fixed an issue when ImplementsInterface attribute was used for ScriptableObjects.
- Added some Runtime Gizmo drawing functions, such as DrawEllipsoid.
- added SigmoidUpDown as a DefaultCurve for quick AnimationCurve initialization.
- Removed a warning when Hands.Provider returned null, since many valid scenes may simply not be set up for Leap devices.

# Interaction Engine 1.2.0 Changelog

## General
- Added the IgnoreCollidersForInteraction component, which causes a Collider to be ignored by the Interaction Engine. This is intended for trigger colliders that should be considered only for raycasting and not, e.g., for grasping or collision.
- Removed the (unused and deprecated) RigidbodyWarper script.
- VR controllers will now be detected at runtime in the IE example scenes that support them. (Previously, they had to be active and visible to Unity as soon as the application started running.) This functionality causes a minimal amount of garbage allocation every few seconds in those scenes, but did not have a noticeable impact on garbage collection times. You can disable this functionality by disabling "pollConnection" on the InteractionXRController script.

## Grasping
- Improved the ability to grasp small objects.
- Improved the consistency of grasping an object that was held by the other hand.
- Added OnGraspBegin, OnGraspStay, and OnGraspEnd callbacks to InteractionController.

## UI
- Modified the Basic UI example scene to demonstrate how a InteractionButton UI panel can be moved without causing the attached physical buttons to wobble.

## Curved Space Support
- Fixed a bug where Interaction Objects inherited LeapSpace components that were only used for an object's LeapGraphics. InteractionBehaviour now only considers any LeapSpaces -- such as LeapCylindricalSpace or LeapSphericalSpaces -- that are on or the parent of a Collider.

# Graphic Render 0.1.3 Changelog
- **If you are upgrading a project using the Graphic Renderer from 2017.3 to 2017.4, you may experience a crash when opening certain scenes. This is due to sprite library data incompatibility and can be resolved by deleting your Temp and Library folders and re-opening Unity.**
- Fixed an issue where graphics in Dynamic or Text groups would cause MissingReferenceExceptions when deleted. Note that you cannot delete (or add) graphics in Baked groups, although you can disable them.
- Fixed an issue where tints and vertex colors were not correctly converted to linear space when the application was targeting linear space rendering.
- Incremental bugfixes for stability.

# Hands Module 2.1.4 Changelog
- Fixed an issue where RiggedHand would occasionally raise a LookRotation error.
- Added an option for Rigged Hands to scale their last fingerbones, which allows rigged hands to achieve correct fingertip positions when tracked hands do not precisely match the scale of the model the hand was rigged on.


nickjbenson:
As this gets ready I will compile formal release notes and add them here.

hideturtle:
These errors appear when creating packages in 2017.3 and importing them into 5.6.2f1:

image

nickjbenson:
Responding the Steve above, 5.6 support should be added for exported projects now. (Looks like we've tested and know it works now.)

nickjbenson:
We're missing a null check or two in the rig upgrader so I'm placing a reminder here to add those.

nickjbenson:
We are experiencing two critical bugs presently:

- [ ] Starting an example in the Editor with an LMC plugged in, then unplugging the LMC, results in Unity freezing. It unfreezes if you plug the peripheral back in. Likely we have a loop somewhere that is blocking the main thread while the controller is unplugged.

- [ ] We are experiencing inconsistencies in temporal warping latency values and diagnosing across Rigels and LMCs.

hideturtle:
When failing to set the proper values for Time and Physics, the scene does not generate Console warnings when run.

nickjbenson:
Targeted a PR to open the LM window when checks fail (they can be ignored or they can be addressed via auto-fix).

PR #1011, opened 2018-04-20
Old Leap Rig Upgrader

nickjbenson:
This PR:
- [x] Adds a Leap Motion settings window, accessible via the Window menu.
- [x] Mirrors Leap Preferences settings in that window in a tab.
- [x] Adds a tab for Project Checks and an API for adding project checks.
- [x] Adds a Leap Rig Upgrader project check box and a tab for its results to the LM window.
- [x] Allows automatically upgrading any detected old Leap rig with the click of a button.

Still in-progress:
- [x] Image rigs do not upgrade and fully work out of the box, their upgrade path needs some tweaks.
- [x] Settings from the LeapServiceProvider should be copied over into the LeapXRServiceProvider.
- [x] "Missing component removal" notifications and functionality should be totally removed from the code -- this is no longer necessary since we delete those objects during an upgrade. This is good because this functionality was crashing the editor!
- [x] The workflow is missing buttons in key places: You should be able to re-scan directly in the upgrader tab (easy change, need to refactor some GUI code). Pressing the scan button from the Project Settings tab should open the Upgrader tab if any old rigs were detected in the current scene.
- [x] Upgrading a rig automatically invalidates any older detected rig data. That part is fine, but we should auto-rescan the open scene after upgrading a rig.
- [x] Upgrade details display is messy, just needs some tweaking and spacing to look professional.
- [x] Remove ~planning documents and~ test scenes

To test:
- [x] Try upgrading various forms of old Leap rigs. After hitting the button, you should be able to press play and have everything look fine.
- [x] Tweak your old leap rigs. Change transform names, Add components to it. Add extra transform children. Validate that tweaked old-style Leap rigs that still work are upgradeable and work, and that extra transform children are not lost. (This will require working with an old UnityModules version.)
- [x] Validate that the workflow makes sense and that upgrading is discoverable.
- [x] Upon first loading UnityModules into your Unity editor, the LM settings window should open automatically. Careful, this sets an EditorPref that sticks around with your Unity installation :)


nickjbenson:
Alex approved; merging.

PR #1020, opened 2018-04-25
Detatch on destroy if we are at runtime

Amarcolina:
Simple fix that allows graphics to be destroyed at runtime (if they can be detached at runtime!)


nickjbenson:
Hm, if I go into the `Curved Space Examples` scene, click a graphic and delete it from the editor at runtime, I get MissingReferenceExceptions currently.

Issue #1021, opened 2018-04-26
Interaction Engine: Controllers are not found, when app starts too fast


Tom-LeFx:
Issue in Interaction Engine Modules for Unity, even in example scenes.

Steps to reproduce:

- Deactivate Splash Screen (from Unity or any other one)
- Make a build, containing only this one level, which gets loaded on startup
- find yourself without working controllers

---
Alternative reproduction (like we found out about it):

- Import SteamVR (for examplen when you want to enable haptic controller feedback)
- let it adjust all suggested player settings (e.g. disable splash, etc)
- build any level, that directly starts as VR
- no hands

This should really be looked at, or at least be documented. 

Kind of the same direction:

If controllers are not turned on and detected at app-start, they will not be found later on. SteamVR/VRTK handles this without any issues...Leap Interaction Engine requires them to be present and active on startup

nickjbenson:
Thanks for the ticket! I'm investigating.

We made this decision originally because the method to detect controllers allocates garbage. Looking at Unity's SDK, they still recommend using `Input.GetJoystickNames()`, which allocates garbage to returns a `string[]`:
https://docs.unity3d.com/ScriptReference/Input.GetJoystickNames.html

I find this to be pretty terrible! We've taken a lot of care to prevent the Interaction Engine from ever allocating garbage, because GC collects kill frames and make people sick. Unfortunately we may have to accept some periodic allocation from our example scene's controller detection script in order to address this issue.

nickjbenson:
OK, we're getting https://github.com/leapmotion/UnityModules/pull/1022 reviewed for our latest release, which should address this issue.

nickjbenson:
It is still suspicious to me that you say you're able to produce a scenario where you aren't even getting hand data. Do your non-SteamVR repro steps also prevent hand data from appearing in the application?
@Tom-LeFx

nickjbenson:
Circling back again. We found another issue with controller initialization and the fix is being vetted before we apply it to all of our example scenes.
https://github.com/leapmotion/UnityModules/pull/1024

PR #1022, opened 2018-04-26
Feature: Periodic Controller Polling

nickjbenson:
This PR addresses #1021 by making some important changes to InteractionXRController.

- [ ] Fundamentally, the active/inactive state of InteractionXRControllers should determine whether or not it _tries_ to have tracking, not whether it _has_ tracking. Disabling the InteractionXRController's gameObject or Component should prevent updates but not reflect the fact that the joystick wasn't detected.
- [ ] `pollConnection` is now an option, plus `pollConnectionInterval`. This option, enabled by default, allows the controller to call Input.GetJoystickNames(), which unfortunately allocates garbage. To minimize allocation, `pollConnection` only checks joystick names if the script is active and enabled but hasn't seen the joystick yet in the device list. Once it has seen the joystick, you must manually call `RefreshControllerConnection` to find out if the device list has removed the joystick.
- [ ] `isTracking` now requires both the joystick tokens to have matched a device in the device list AND that XR tracking data is available from the XRNode tracking provider. The joystick token test is skipped if the tokens string is empty or null.
- [ ] Adds a convenience List to the InteractionXRController inspector allowing the user to specify GameObjects to keep active IFF the controller `isTracking`.
- [ ] Some null checks were added to the Interaction Engine to avoid accessing `contactBones` if the array is null, solving reference errors happening when OnDisable occurred before contact bones were initialized.
- [ ] Converted even **inactive** colliders beneath InteractionXRControllers into ContactBones to prevent an issue where contact bones were not initializing when controllers were not immediately detected as active joysticks. Trigger colliders beneath InteractionXRControllers are also no longer converted to ContactBones so they can't apply forces to objects.

To test:
- [ ] Make sure `pollConnection = false` prevents the controller from allocating garbage.
- [ ] Make sure that the example scenes function with InteractionXRControllers, and that even if you create a build with the splash screen disabled or your controllers aren't connected on application startup that you can still get controller data to the application once they wake up.
- [ ] Make sure the feature changes make sense
- [ ] Make sure the code changes make sense

### DIFF
--- a/Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
+++ b/Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
@@ -563,6 +563,8 @@ namespace Leap.Unity.Interaction {
         includeInactiveObjects: true);
 
       foreach (var collider in _colliderBuffer) {
+        if (collider.isTrigger) continue; // Contact Bones are for "contacting" colliders.
+
         ContactBone contactBone = collider.gameObject.AddComponent<ContactBone>();
         Rigidbody body = collider.gameObject.GetComponent<Rigidbody>();
         if (body == null) {
--- a/Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
+++ b/Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
@@ -238,12 +238,7 @@ namespace Leap.Unity.Interaction {
                                  .Any(joystick => controllerSupportTokens.Query()
                                                  .All(token => joystick.Contains(token)));
 
-        if (matchesController) {
-          _isJoystickDetected = true;
-        }
-        else {
-          _isJoystickDetected = false;
-        }
+        _isJoystickDetected = matchesController;
       }
     }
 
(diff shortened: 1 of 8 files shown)


Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
```diff
@@ -99,9 +124,132 @@ public class InteractionXRController : InteractionController {
...
+
+    private float _pollTimer = 0f;
+
+    private void fixedUpdatePollConnection() {
+      if (_pollConnection && !_isJoystickDetected) {
+        _pollTimer += Time.fixedDeltaTime;
+      }
+
+      if (_pollTimer >= pollConnectionInterval) {
+        _pollTimer = 0f;
```
jselstad: @Amarcolina taught me this one: rather than setting the timer to zero, you want to subtract `pollConnectionInterval` from the current timer value.

This keeps you from losing the bits of time that `Time.fixedDeltaTime` causes `_pollTimer` to _exceed_ `pollConnectionInterval`.

Assets/LeapMotion/Modules/InteractionEngine/Scripts/InteractionXRController.cs
```diff
@@ -99,9 +124,132 @@ public class InteractionXRController : InteractionController {
...
+        string[] joysticksConnected = Input.GetJoystickNames().Query()
+          .Select(s => s.ToLower()).ToArray();
+        string[] controllerSupportTokens = deviceJoystickTokens.ToLower()
+          .Split(" ".ToCharArray());
+        bool matchesController = joysticksConnected.Query()
+                                 .Any(joystick => controllerSupportTokens.Query()
+                                                 .All(token => joystick.Contains(token)));
+
+        if (matchesController) {
+          _isJoystickDetected = true;
```
jselstad: Should perhaps set `_isJoystickDetected` directly to `matchesController`

nickjbenson:
le resolved in le person

PR #1024, opened 2018-04-27
Bring it back agaaain. Fix controller switching in IE example scenes, for Vive especially

nickjbenson:
OK, this is ready for checking.

- [ ] Vive controllers have L and R marked so you don't get them confused, this is important for some example scenes
- [ ] Controller support double-checked through.
- [ ] InteractionControllerSwitcher gone. Replaced with a simpler example script on a per-controller basis. I don't love the new script that much better but it is simpler. Scene references are bleh.

Boy it is annoying making changes to all of our example scenes. Nested prefabs one day...! Automated testing another one day after that...!


Amarcolina:
Excellent!

PR #1032, opened 2018-05-12
Add attempts to reconnect to the service when not connected

jselstad:
This PR attempts to solve the sporadic issue that sometimes occurs where users will client applications will spuriously not receive hands; we suspect it's caused by the service not acknowledging the connection attempt by the app the first time, so we attempt to simulate an "app restart" by destroying and recreating the controller object.

You can easily test whether doing this could break working apps by adding ` && !Input.GetKey(KeyCode.Space)` to the service connection check on line 506 and holding down Space for several seconds (to artificially induce a "Destroy/Create" cycle).


Amarcolina:
do we want to reset `_numberOfReconnectionAttempts` if the service connects again?
i also might argue that we should never stop retrying
if the device is unplugged for X seconds, we could 'use up' all our retries before it even matters

PR #1040, opened 2018-05-18
Fixed and improved image retriever logic

Amarcolina:
- Added logic to reset the image queue if the service jumps back in time, otherwise the queue will stall until the sequence id catches back up to where it was.  This typically only happens if the service restarts during the application lifecycle
 - Also reset the image queue if it gets filled up and new images can't be added, we also log a warning in this case to warn the user weird stuff is happening
 - Upped the image queue length from 32 images to 128.  It's just a safety precaution, I don't expect we will get anywhere close to 128 during normal operation.  Which means that if the buffer _does_ fill up, we can be that much more sure something strange is happening, and we can log a warning letting the user know their images might be strange.
 

To test: 
 - [x] Make sure image scenes still work as expected
 - [x] Make sure that you can restart the service while the application is running, and images don't freeze for long periods of time

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/LeapImageRetriever.cs
+++ b/Assets/LeapMotion/Core/Scripts/LeapImageRetriever.cs
@@ -41,8 +41,11 @@ namespace Leap.Unity {
     private EyeTextureData _eyeTextureData = new EyeTextureData();
 
     //Image that we have requested from the service.  Are requested in Update and retrieved in OnPreRender
-    protected ProduceConsumeBuffer<Image> _imageQueue = new ProduceConsumeBuffer<Image>(32);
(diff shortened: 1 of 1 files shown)


Assets/LeapMotion/Core/Scripts/LeapImageRetriever.cs
```diff
@@ -337,7 +348,21 @@ public class EyeTextureData {
 
     private void onImageReady(object sender, ImageEventArgs args) {
       Image image = args.image;
-      _imageQueue.TryEnqueue(image);
+
+      if (!_imageQueue.TryEnqueue(image)) {
+        Debug.LogWarning("Image buffer filled up.  This is unexpected means images are being provided faster than " +
```
nickjbenson: `This is unexpected means` <-- typo. Most string output we have elsewhere has a single space after periods, can we enforce that here as well?

PR #1043, opened 2018-06-01
Fix null when spoofing Combinable Properties (AppModules)

nickjbenson:
Why would you want to spoof Combinable Property attributes, you ask?

AppModules does this in the `Streams` library because it needs to use Reflection to figure out what interface actually needs to be satisfied at runtime in the Editor. Attributes don't support generic types because they have to be resolved at compile-time, so to utilize the ImplementsInterface combinable property attribute, I can construct it in the custom editor and call CombinablePropertyDrawer's OnGUI() manually to do the appropriate checks DRYly.

Long story short this works great, we just need to not RE-initialize the attribute's fieldInfo with a bad fieldInfo when OnGUI is used in this way (with `null` for fieldInfo, to account for the fact that the attributes are already initialized with the appropriate fieldInfo).

### DIFF
--- a/Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
+++ b/Assets/LeapMotion/Core/Scripts/Attributes/Editor/CombinablePropertyDrawer.cs
@@ -79,7 +79,9 @@ namespace Leap.Unity.Attributes {
 
       IFullPropertyDrawer fullPropertyDrawer = null;
       foreach (var a in attributes) {
-        a.Init(fieldInfo, property.serializedObject.targetObjects);
+        if (fieldInfo != null) {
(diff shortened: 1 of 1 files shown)


Amarcolina:
I'd like to suggest a solution that doesn't involve constructing an ImplementsInterface attribute manually
how about an overload of ImplementsInterface is created where you pass in the name of a static method that returns the type that is to be implemented?

```[ImplementsInterface("getInterfaceType")]
private MonoBehaviour _stream;

private static Type getInterfaceType() {
  return typeof(StreamType);
}```

That way the functionality is built into the ImplementsInterface class, and doesn't require any custom editor functionality to work, so it could be more easily re-used if this case comes up again
and it also prevents you from needing to construct mock attributes and deal with cases where the field info is null.  I'd like to not support a second workflow where attributes are being manually created and gui methods manually being called.

Issue #1044, opened 2018-06-11
StreamingFolder does not Serialize/Deserialize in .unitypackage


jselstad:
When I try serializing a StreamingFolder as a public member of a monobehaviour in a unitypackage, it deserializes as an empty string :/  (meaning that it will try using the top level folder path of StreamingAssets).

My workaround is to include the streaming assets sub-folder into the "file name" (since that seems to be serializing properly).

PR #1048, opened 2018-06-12
Fixed streaming assets and folders when in packages

Amarcolina:
When a unity package loads a streaming asset or folder, the relative path is correct, but the asset reference will be null since packages don't bring folder references across.  This causes the folder to overwrite itself with an incorrect null value.  This PR makes a change so that when serializing, if it sees that the reference is null, but the path is not, it will try to see if there is a valid folder and uses that one instead of nulling out the path.


nickjbenson:
(theoretically) fixes #1044 once merged

PR #1049, opened 2018-06-13
Import UMward things from AppExperiments

nickjbenson:
Yeeep it's another big one! I have placed checkboxes on items so we can check them off when they are good-to-go review-wise or otherwise completed.

- [ ] Adds Utilities:
  - [ ] Transform.GetAllChildren (breadth-first and depth-first variants, garbageless)
  - [x] More things to Infix utils
  - [x] Nullable (e.g. `bool?`, `Vector3?`) ValueOr a la Maybe<T>.ValueOr
  - [x] WithX/WithY/WithZ & WithW (conversion to Vector4) extensions on Vector3 (think: basic Swizzle without going FULL `using Swizzle;`, I think these are useful and don't pollute intellisense too much)
  - [x] Rigidbody.SetPose and Rigidbody.MovePose
  - [ ] RuntimeGizmo extensions, DrawPose, DrawRay, DrawDashedLine, DrawBar. (Come to think about it, these should probably be in RuntimeGizmoManager now isstead!)
  - [ ] Gradient.ToTexture (needs a doc comment)
  - [x] Pose.Integrated(this Pose, Movement, deltaTime)
  - [x] DELETES From/Then extension methods. These were a mistake, old me was wrong. New me is right.
- [ ] Adds Movement.cs, a struct containing the derivative of a Pose, with angular velocity as an angle-axis vector
- [ ] Adds PeriodicBehaviour.cs, a base class for MonoBehaviours that don't actually need Update _every_ frame
- [ ] Adds a RuntimeTransformGizmo component, a la RuntimeColliderGizmos but for a simple gizmo at the root of a transform.
- [ ] Adds FollowHandPoint.cs! I find this script extremely useful. Think Attachments without all the fluff. It also provides a Pose Stream, but using that is totally optional.
- [ ] Adds Streams, in a "batteries-included" fashion: Connectors, Pose filters, support for Rays and Vector3s as well.
- [ ] Adds Splines, including handy-dandy support for Hermite Pose Splines and even a Catmull-Rom constructor for Hermite Pose Splines
- [ ] Adds an official way to put hand data into shader globals via `HandData.cginc` and the HandShaderGlobalSupport.cs component.
- [ ] Moves HandUtils to the Hands folder instead of the Utils folder. ... I guess either makes sense.
- [ ] Adds Switches! Minus SequenceSwitches...
- [ ] Since we have IIndexable in Core, adds SequenceSwitches, which now use IIndexable with dynamic ImplementsInterface constraining! Give the SequenceSwitch a reference to an object that implements `IIndexable<GameObject>`, and it will automate switching the objects in the sequence one after another, as long as all of those GameObjects themselves have an IPropertySwitch on them. These things are super slick for fancy interface animations.
- [ ] Adds HandRayStream, which provides a stream of Rays that can be configured to cast out from the head, an elbow approximant, or a shoulder approximant, as we have used in various experiments. This will give us some example code to point to if people ask about raycasting with hands.

Oh, how could I forget?
- [ ] Adds the **Gestures** module as a new Experimental module! Featuring the PinchGesture from Paint, minus the DevGui.

*Tasks Remaining (help appreciated?)*
- [ ] Move RuntimeGizmo extensions into the RuntimeGizmoManager proper, if it makes sense
- [ ] StreamConnectors should _not_ use the bad bad OnGUI/ImplementsInterface cheat, they should take advantage of the fact that ImplementsInterface can be provided with a getter method now.
- [ ] Is it possible to make a generic serializable StreamConnector that doesn't have to use the UnityEvent-declaration-style? @Amarcolina There has got to be some better, awful, reflection-y way to do this, rather than having to declare `public class Vector3StreamConnector : StreamConnector<Vector3> { }`
- [ ] The ImplementsInterface attribute drawer needs to stringify generic types better -- currently they are drawn like `IIndexable'1` when they should be e.g. `IIndexable<GameObject>`
- (Did I miss something? Feel free to add more to-dos)

*Example-related Tasks (help appreciated!)*
- [ ] Streams, Splines, Switches, FollowHandPoint, HandRayStream -- these things could use an example scene or two to demonstrate why they are interesting at all. I think this could really spice up our Core examples actually!
- [ ] Gestures should also have an example scene or two, at least one demonstrating the PinchGesture and what the optional safety checks do.
- [ ] The TwoPoseTRS example needs validation to make sure my quaternion math is right, I might have an order-of-rotation bug in there
- [x] Test/Example for HandData.cginc and HandShaderGlobalSupport.cs functionality.


Assets/LeapMotion/Core/Resources/HandData.cginc
```diff
@@ -0,0 +1,311 @@
...
+// =========
+
+#define IDENTITY_MATRIX4x4 { 1.0, 0.0, 0.0, 0.0, \
+                             0.0, 1.0, 0.0, 0.0, \
+                             0.0, 0.0, 1.0, 0.0, \
+                             0.0, 0.0, 0.0, 1.0 }
+
+float4x4 Leap_HandData_Preprocess_Matrix = IDENTITY_MATRIX4x4;
+
+void Leap_ClearPreprocessMatrix() {
```
Amarcolina: Should this method just get removed?  It's a no-op

Assets/LeapMotion/Core/Resources/HandData.cginc
```diff
@@ -0,0 +1,311 @@
...
+  float3 pa = p - a, ba = b - a;
+  float h = clamp( dot(pa,ba)/dot(ba,ba), 0.0, 1.0 );
+  return Leap_SqrLength( pa - ba*h );
+}
+
+float Leap_SqrDistLine_WithThickness(float3 p, float3 a, float3 b, float approxR) {
+  return max(Leap_SqrDistLine(p, a, b) - approxR * approxR, 0);
+}
+
+float dot2(float3 v) { return dot(v,v); }
```
Amarcolina: dot2 is the same as Leap_SqrLength, we should probably keep one and ditch the other.

Assets/LeapMotion/Core/Resources/HandData.cginc
```diff
@@ -0,0 +1,311 @@
...
+	dist = min(dist, Leap_SqrDist(pos * scale, mul(Leap_HandData_Preprocess_Matrix, _Leap_RH_Fingertips[2]) * scale));
+	dist = min(dist, Leap_SqrDist(pos * scale, mul(Leap_HandData_Preprocess_Matrix, _Leap_RH_Fingertips[3]) * scale));
+	dist = min(dist, Leap_SqrDist(pos * scale, mul(Leap_HandData_Preprocess_Matrix, _Leap_RH_Fingertips[4]) * scale));
+	return dist;
+}
+
+float Leap_SqrDistToHand(float3 pos) {
+  // Fingertips
+	//float dist =     Leap_SqrDist(pos, mul(Leap_HandData_Preprocess_Matrix, _Leap_LH_Fingertips[0]));
+  //for (int i = 1; i < 5; i++) {
```
Amarcolina: Lots of commented out code here, still needed?

Assets/LeapMotion/Core/Resources/HandData.cginc
```diff
@@ -0,0 +1,311 @@
...
+	return dist;
+}
+
+// ------
+// Planes
+// ------
+
+//float3 Leap_FingertipsDepthInPlane_AssumeHandsInLocalPlaneSpace() {
+//  float3 closest = _Leap_LH_Fingertips[0];
+//  for (int i = 1; i < 5; i++) {
```
Amarcolina: To remove?

Assets/LeapMotion/Core/Resources/HandData.cginc
```diff
@@ -0,0 +1,311 @@
...
+//    if (_leap_LH_Fingertips[i].z < )
+//  }
+//}
+
+// Copied-- plane displacement shader
+// TODO: Rename, potentially refactor
+
+// Uses dot product to get the component distance along the normal.
+// Also sets the parallel and perpendicular vectors.
+float distanceToPlane(float3 normal, float4 point1, float4 point2, out float3 perp, out float3 para) {
```
Amarcolina: Does this belong up with the other distance functions?  It has a different naming convention than the rest.

Assets/LeapMotion/Core/Scripts/Attributes/CombinablePropertyAttribute.cs
```diff
@@ -129,4 +130,43 @@ public abstract class CombinablePropertyAttribute : PropertyAttribute {
     public virtual void OnPropertyChanged(SerializedProperty property) { }
 #endif
   }
+
+  public static class AttributeExtensions {
+
+    public static MethodInfo FindMethodByName(this Type type,
```
Amarcolina: The class Type already has a method called GetMethod that takes a string argument.  What makes this method different?  I think that the name of the method should reflect what makes it special.

Amarcolina: Also, why is this extension method placed inside of the CombinablePropertyAttribute class?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -45,28 +73,48 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
...
+    /// </summary>
+    public Type GetInterfaceType() {
+      if (this._type != null) {
+        return _type;
+      }
+      else { // use _typeGetterMethodName.
+        var target = targets.Query().FirstOrDefault();
+        var objectType = target.GetType();
+
+        var method = objectType.FindMethodByName(_typeGetterMethodName);
```
Amarcolina: Might be nice to have some more exact method signature requirements enforced here.  Like the method must be static, must take zero parameters, and must not be generic, and must return a value of type Type.  Right now if the method returns something other than type, there would be an unclear null reference exception somewhere.

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
```diff
@@ -45,28 +73,48 @@ public class ImplementsInterfaceAttribute : CombinablePropertyAttribute,
                                                          property.objectReferenceValue);
 
         if (implementingObject == null) {
-          Debug.LogError(property.objectReferenceValue.GetType().Name + " does not implement " + type.Name);
+          Debug.LogError(property.objectReferenceValue.GetType().Name + " does not implement " + _type.Name);
```
Amarcolina: Is the _type variable safe to use here?  Can't that be null in certain situations?

Assets/LeapMotion/Core/Scripts/DataStructures/Movement.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public Pose ToPose() {
+      var angVelMag = angularVelocity.magnitude;
+      return new Pose(velocity,
+        Quaternion.AngleAxis(angVelMag, angularVelocity / angVelMag));
+    }
+
+    public static Movement operator +(Movement movement0, Movement movement1) {
+      return new Movement(movement0.velocity + movement1.velocity,
+                          movement0.angularVelocity + movement1.angularVelocity);
```
Amarcolina: Adding angular velocities this way is not actually a correct way to add angular velocities.  It is definitely convenient, but it might be worth noting somewhere.

Assets/LeapMotion/Core/Scripts/FollowHandPoint.cs
```diff
@@ -0,0 +1,248 @@
...
+        moveToAttachmentPointNow();
+      }
+    }
+
+    #endregion
+
+    #region On Frame Event
+
+    private void onUpdateFrame(Frame frame) {
+      if (frame == null) Debug.Log("Frame null");
```
Amarcolina: Maybe a more descriptive error message?  Using an error path instead of a log path would also be great.

Assets/LeapMotion/Core/Scripts/Hands/HandShaderGlobalSupport.cs
```diff
@@ -0,0 +1,323 @@
...
+          var hand = rightHand;
+          var palmTris = _rightPalmTris;
+          int i = 0;
+          
+          {
+            Finger thumb = hand.GetThumb(), index = hand.GetIndex(),
+              middle = hand.GetMiddle(), ring = hand.GetRing(), pinky = hand.GetPinky();
+
+            // Thumb Tri
+            palmTris[i++] = thumb.bones[1].NextJoint.ToVector3().WithW(1);
```
Amarcolina: could there be a way to remove this duplication between the left and right hands?

Assets/LeapMotion/Core/Scripts/Splines/CatmullRom.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// </summary>
+    private static void CalculateCatmullRomParameterization(Vector3 p0, Vector3 p1,
+                                                            Vector3 p2, Vector3 p3,
+                                                            out Vector3 v1,
+                                                            out Vector3 v2) {
+      v1 = Vector3.zero;
+      v2 = Vector3.zero;
+
+      // (Uniform Catmull-Rom)
+      //v1 = (p2 - p0) / (t2 - t0);
```
Amarcolina: Is this needed anymore?

Assets/LeapMotion/Core/Scripts/Splines/CatmullRom.cs
```diff
@@ -0,0 +1,174 @@
...
+                                                            out Vector3 aV1,
+                                                            out Vector3 aV2) {
+      aV1 = Vector3.zero;
+      aV2 = Vector3.zero;
+
+      // Handy little trick for using Euclidean-3 spline math on Quaternions, see
+      // slide 41 of
+      // https://www.cs.indiana.edu/ftp/hanson/Siggraph01QuatCourse/quatvis2.pdf
+      // "The trick: Where you would see 'x0 + t(x1 - x0)' in a Euclidean spline,
+      // replace (x1 - x0) by log((q0)^-1 * q1) "
```
Amarcolina: Is this trick not relevant anymore?

Assets/LeapMotion/Core/Scripts/Splines/CatmullRom.cs
```diff
@@ -0,0 +1,174 @@
...
+      return new HermitePoseSpline(p1, p2, new Movement(v1, aV1), new Movement(v2, aV2));
+    }
+
+    #region Archived code -- legacy code relies on this. It is incorrect though!
+
+    // TODO: The below isn't ACTUALLY a centripetal Catmull-Rom spline, it's just a
+    // spline parameterization -- centripetal-ness requires choosing "time values" that
+    // correspond to the distances between the four positions. The implementation
+    // above is better.
+    // deleeeeete it!
```
Amarcolina: Is this todo still needed?  Should something be deleted?

Assets/LeapMotion/Core/Resources/HandData.cginc
nickjbenson: yup yup

Assets/LeapMotion/Core/Resources/HandData.cginc
nickjbenson: ye

Assets/LeapMotion/Core/Resources/HandData.cginc
nickjbenson: I don't remember :D will see if it's useful for anything, might have been an early version that only checked fingertip positions

Assets/LeapMotion/Core/Resources/HandData.cginc
nickjbenson: Will rename and move

Assets/LeapMotion/Core/Scripts/Attributes/CombinablePropertyAttribute.cs
nickjbenson: because it was copied from Quickbutton -- I still need to make Quickbutton use this method instead, so I'll do that. But this sort of reflection has been useful for attributes specifically, is why. I can move it to some.. reflection utility alternatively?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
nickjbenson: Don't see a reason for requiring the method to be static. I should enforce returning Type, but it should be fine if the method is generic as well?
Also wouldn't the only requirement be that it doesn't have non-optional arguments?

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
nickjbenson: oops. I'm going to refactor this script a bit, the way I set up the type options is very fragile.

Assets/LeapMotion/Core/Scripts/DataStructures/Movement.cs
nickjbenson: Will comment

Assets/LeapMotion/Core/Scripts/FollowHandPoint.cs
nickjbenson: :+1:

Assets/LeapMotion/Core/Scripts/Attributes/ImplementsInterface.cs
Amarcolina: I'm not sure how you would go about invoking a generic method, how would you determine the generic parameter types to use?  And yeah I forgot we have a reference to target, so totally fine to have instance method instead of static.

Assets/LeapMotion/Core/Scripts/Splines/SplineUtil.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Leap.Unity.Splines {
+
+  using IPositionSpline = ISpline<Vector3, Vector3>;
+  using IPoseSpline = ISpline<Pose, Movement>;
+  public static class SplineUtil {
+
+    #region Conversion Utils
+
+    public static PoseToPositionSplineWrapper AsPositionSpline(this IPoseSpline poseSpline) {
```
Amarcolina: Is there any pose  spline that doesn't already implement ISpline<Vector3, Vector3> ?  This doesn't seem to be used anywhere.

Assets/LeapMotion/Core/Scripts/Streams/Abstract/StreamSkippingFilter.cs
```diff
@@ -0,0 +1,119 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Leap.Unity {
+
+  public abstract class StreamSkippingFilter<T> : MonoBehaviour,
```
Amarcolina: What is the point of making this an abstract class?  It seems to have concepts that are only applicable for pose-like streams (and indeed, it seems the only implementation is of type pose)

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -682,6 +696,41 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
       public void Dispose() { }
     }
 
+    /// <summary>
+    /// Scans all the children in order of the argument Transform, appending each transform
+    /// it finds to toFill. Children are added depth-first by default.
+    ///
+    /// Pass breadthFirst: true to fill the list breadth-first instead.
+    /// </summary>
+    public static void GetAllChildren(this Transform t, List<Transform> toFill,
```
Amarcolina: How about GetDescendants?

Assets/LeapMotion/Core/Scripts/Streams/IStreamReceiver.cs
```diff
@@ -0,0 +1,15 @@
+﻿
+
+namespace Leap.Unity {
```
Amarcolina: If there is a Leap.Unity.Streams namespace, I think all these scripts in the Streams folder should probably be in that namespace!

Assets/LeapMotion/Core/Scripts/Streams/PoseStreams/DeadzonePoseFilter.cs
```diff
@@ -0,0 +1,35 @@
...
+    public event Action OnOpen = () => { };
+    public event Action<Pose> OnSend = (pose) => { };
+    public event Action OnClose = () => { };
+
+    public void Close() { OnClose(); }
+
+    public void Open() { OnOpen(); }
+
+    public void Receive(Pose data) {
+      if (data.position.sqrMagnitude > deadzoneRadiusSqr) {
```
Amarcolina: Would it be nice when the position enters the deadzone, you output a pose at the center of the deadzone, that way the pose isn't perpetually stuck at the edge of the deadzone?

Assets/LeapMotion/Core/Scripts/Streams/PoseStreams/LineRendererPoseReceiver.cs
```diff
@@ -0,0 +1,44 @@
...
+  /// <summary>
+  /// Fills received poses into a buffer, then fills a LineRenderer object with
+  /// the received Pose data on Close().
+  /// </summary>
+  public class LineRendererPoseReceiver : MonoBehaviour,
+                                          IStreamReceiver<Pose> {
+    
+    public LineRenderer lineRenderer;
+
+    public const int POINTS_BUFFER_SIZE = 1024;
```
Amarcolina: Why a hardcoded max limit of 1024?  Would a List<> be more suitable?

Assets/LeapMotion/Core/Scripts/Streams/PoseStreams/PoseStreamMovingAverageFilter.cs
```diff
@@ -0,0 +1,92 @@
...
+      bool bufferWasNotFull = false;
+      if (!buffer.IsFull) {
+        bufferWasNotFull = true;
+      }
+
+      buffer.Add(data);
+
+      if (buffer.IsFull) {
+        if (bufferWasNotFull) {
+          for (int i = 1; i < buffer.Count; i += 2) {
```
Amarcolina: Could you explain the logic here?  It seems that the stream doesn't send anything until the buffer fills up, but once it does, it quickly outputs one value for every value in the buffer?  Why not be sending averages as the buffer fills up?

Assets/LeapMotion/Core/Scripts/Streams/StreamConnector.cs
```diff
@@ -0,0 +1,197 @@
...
+    public abstract System.Type streamReceiverType { get; }
+  }
+
+  [ExecuteInEditMode]
+  public abstract class StreamConnector<StreamType> : StreamConnector,
+                                                      IRuntimeGizmoComponent,
+                                                      ISerializationCallbackReceiver {
+
+    [SerializeField]
+    // Sadly, this doesn't compile (name isn't constant), so we'll have to
```
Amarcolina: With the new implements interface, is this still needed?

Assets/LeapMotion/Core/Scripts/Switches/Internal/Editor/PropertySwitchEditor.cs
```diff
@@ -0,0 +1,83 @@
...
+    private bool shouldDrawInterpolationCurve() {
+      return targets.Query().Any(t => PropertySwitch.IsInterpolatable(t.propertyType));
+    }
+
+    // TODO: Delete
+    //#region Event Table
+
+    //private EnumEventTableEditor _tableEditor;
+
+    //private void drawEventTable(SerializedProperty property) {
```
Amarcolina: Dead code here?

Assets/LeapMotion/Core/Scripts/Switches/Internal/IPropertySwitch.cs
```diff
@@ -0,0 +1,85 @@
...
+    bool GetIsOffOrTurningOff();
+
+    /// <summary>
+    /// This method must result in the same effects as On(), but additionally reflect the
+    /// FINISHED state of the object after an On() call, immediately.
+    /// 
+    /// This method must be valid to call during edit-time, and its effects must also
+    /// respect Undo behaviour through the use of Undo.X() calls.
+    /// </summary>
+    void OnNow();
```
Amarcolina: The fact that this is prefixed by On had me confused for a second, thinking it was a callback of the OnBlah style.  Something like, get an OnNow every time a new 'now' time is calculated or something.  Maybe it could be renamed to something like 'TurnOnNow'?

Amarcolina: Also, I'd want to look to see if we can avoid the whole
'DoTheThing'
'DoTheThingNow'
pattern that unity uses for some of their methods.  I'd rather that both methods are clear about the time they take to perform their actions.  Maybe something like:
'BeginActivation'
'ActivateNow'
?

Assets/LeapMotion/Core/Scripts/Switches/Internal/Updater.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using UnityEngine;
+
+namespace Leap.Unity {
+
+  public class Updater : MonoBehaviour {
```
Amarcolina: This looks like this might be a concept orthogonal to switches, maybe it could be moved into the general utilities instead?

Amarcolina:
Playing around with SwitchTree, seeing some strange behavior that I'm not sure is intentional or not.  
 - The hierarchy in the inspector has the reverse order compared to the hierarchy itself.
 - I'm not sure what the blue line means compared to a game object having a blue outline.  They seem to be independent concepts, but are also inconstant with the state of the game object?
 - The 'isOn' bool checkbox on the switch tree itself cannot be unchecked 
 - The SwitchTree doesn't always seem to refresh the active state of all children objects, sometimes I can have many active objects, but if i switch to them and switch back, they become disabled

Amarcolina:
treeglitch
Got into a weird state with the tree switch, maybe because some of the child switches were duplicated?  Unsure....

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
```diff
@@ -1364,6 +1541,24 @@ public struct ChildrenEnumerator : IEnumerator<Transform> {
 
     #endregion
 
+    #region Gradient Utils
+
+    public static Texture2D ToTexture(this Gradient gradient, int resolution = 256, TextureFormat format = TextureFormat.ARGB32) {
+      Texture2D tex = new Texture2D(resolution, 1, format, mipmap: false, linear: true);
+      tex.filterMode = FilterMode.Bilinear;
+      tex.wrapMode = TextureWrapMode.Clamp;
+      tex.hideFlags = HideFlags.HideAndDontSave;
```
Amarcolina: I think this method should not assume that the user wants a temporary texture.  Nor should it assume that they want an unreadable texture, since that operation cannot be undone.  Also, we have dealt with issues in the past with gradient textures being linear when they shouldn't have been, so would definitely think that should be another optional argument.

Assets/LeapMotion/Experimental/Gestures/Scripts/GestureUnityEvent.cs
```diff
@@ -0,0 +1,26 @@
...
+    public MonoBehaviour _gesture;
+    public IGesture gesture {
+      get { return _gesture as IGesture; }
+      set { _gesture = value as MonoBehaviour; }
+    }
+
+    public UnityEvent OnGestureFinished;
+
+    void Update() {
+      if (gesture.wasFinished) {
```
Amarcolina: Why only an event for a gesture finishing?

Assets/LeapMotion/Experimental/Gestures/Scripts/Gesture.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// input whether or not the gesture "could be activated" given its current state.
+    /// </summary>
+    public virtual bool isEligible { get { return true; } }
+
+    protected enum DeactivationReason {
+      FinishedGesture,
+      CancelledGesture,
+    }
+
+    #region TODO: Seriously consider removing Actions from the system.
```
Amarcolina: Might be a good idea?  Seems like these events are not used by all gestures.

Assets/LeapMotion/Core/Scripts/Utils/PeriodicBehaviour.cs
```diff
@@ -0,0 +1,82 @@
+﻿using UnityEngine;
+
+namespace Leap.Unity {
+
+  public abstract class PeriodicBehaviour : MonoBehaviour {
+
+    #region Periodicity
+
+    private Maybe<int> _lastKnownPeriod = Maybe.None;
+    private int _updatePeriod = 2;
```
Amarcolina: No inspector interface for the update period?

Assets/LeapMotion/Core/Examples/4. Hand Data Shaders/Standard AlphaFade HandProximity.shader
```diff
@@ -0,0 +1,70 @@
...
+    _ProximityMapping ("Map: DistMin, DistMax, GradMin, GradMax", Vector)
+		  = (0, 0.01, 1, 0)
+
+    // Standard parameters
+		_MainTex ("Albedo (RGB)", 2D) = "white" {}
+		_Color ("Color", Color) = (1, 1, 1, 1)
+		_Glossiness ("Smoothness", Range(0,1)) = 0.5
+		_Metallic ("Metallic", Range(0,1)) = 0.0
+		_EmissionColor ("Emission Color", Color) = (0, 0, 0, 0)
+	}
```
Amarcolina: Tabs/Spaces are all messed up here

Assets/LeapMotion/Core/Examples/4. Hand Data Shaders/Standard AlphaFade HandProximity.shader
```diff
@@ -0,0 +1,70 @@
+﻿Shader "Leap Motion/Examples/Standard AlphaFade HandProximity" {
```
Amarcolina: The convention we have been using so far in all our shaders is to exclude spaces.

nickjbenson: :sob:

Assets/LeapMotion/Core/Examples/Common Example Assets/Standard Fixed-White.shader
```diff
@@ -0,0 +1,46 @@
+﻿Shader "Leap Motion/Example Shaders/Standard Fixed-White" {
```
Amarcolina: This shader uses a different prefix ("Example Shaders" instead of "Examples") and doesn't seem to do anything the standard shader couldn't do?

Assets/LeapMotion/Core/Scripts/Hands/HandDataShaderSupport.cs
```diff
@@ -0,0 +1,323 @@
...
+
+        // Palm Tris
+        {
+          var hand = leftHand;
+          var palmTris = _leftPalmTris;
+          int i = 0;
+
+          // Left Hand Palm Tris
+          {
+            Finger thumb = hand.GetThumb(), index = hand.GetIndex(),
```
Amarcolina: I think I might have made a comment here before, maybe a diff erased it?  This looks like a lot of duplicate code for left and right, would love it if they could be combined.

Assets/LeapMotion/Core/Scripts/Streams/StreamConnector.cs
```diff
@@ -0,0 +1,197 @@
...
+      OnDisable();
+      OnEnable();
+    }
+
+    private void Awake() {
+      initCallbacks();
+    }
+
+    private void OnEnable() {
+      initCallbacks();
```
Amarcolina: Making a note here, if the wire is disabled or enabled at the right time, it can cause the receiver to receive callbacks in an illegal order (two opens in a row without a close for example)

Assets/LeapMotion/Core/Scripts/Utils/DebugPing.cs
```diff
@@ -0,0 +1,548 @@
...
+    //     if (labeledPositionFunc != null) {
+    //       return labeledPositionFunc();
+    //     }
+    //     return labeledPosition;
+    //   }
+
+    //   public override void Draw(RuntimeGizmoDrawer drawer) {
+    //     var facingPosition = getFacingPosition();
+    //     var labeledPosition = getLabeledPosition();
+
```
Amarcolina: Old code can be removed?

Assets/LeapMotion/Core/Scripts/Splines/SplineUtil.cs
nickjbenson: >pose spline that doesn't already implement ISpline<Vector3, Vector3>
Pose Splines are ISpline<Pose, Movement>, this converts them to being used for their position only, ignoring rotation info. This is one of those things I used for something else and then wound up not using that something else. Could remove, but I do think it's useful

Assets/LeapMotion/Core/Scripts/Streams/Abstract/StreamSkippingFilter.cs
nickjbenson: The other implementer is still in AppModules -- or rather, was used in Paint, and was used for StrokePoints, which have poses, but also have other data.

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: That's not very intellisense-discoverable, don't you think? Typing "GetChil-" or "children" should show this in the list of intellisense options

Assets/LeapMotion/Core/Scripts/Streams/Abstract/StreamSkippingFilter.cs
Amarcolina: Ah I see, fair enough

Assets/LeapMotion/Core/Scripts/Splines/SplineUtil.cs
Amarcolina: Roger, makes sense

Assets/LeapMotion/Core/Scripts/Utils/Utils.cs
nickjbenson: GetChildrenRecursive could be more clear

Assets/LeapMotion/Core/Scripts/Streams/IStreamReceiver.cs
nickjbenson: agree!

Assets/LeapMotion/Core/Scripts/Streams/PoseStreams/DeadzonePoseFilter.cs
nickjbenson: hmmm! maybe it should. I should try the two options in an example scene to validate.

Assets/LeapMotion/Core/Scripts/Streams/PoseStreams/LineRendererPoseReceiver.cs
nickjbenson: No, because Lists aren't memory-contiguous, which is a goal here for speed. (Not sure if this script actually sees that idea through, but that would be my intent.)

Amarcolina: Arg, seems like this is the best option due to Unity shenanigans, carry on!

nickjbenson:
WELP, this PR is almost old enough to say its first words, and as much as there is some useful code in here, https://github.com/LeapMotion/AppExperiments contains everything in here and more, so we can always still head there to grab code for e.g. Splines and other useful data types. Some of this stuff will wind up being merged into new UnityModules things as it has spread into other projects.

Closing.

PR #1052, opened 2018-07-11
Add ifdefs to support 2018.2 compile in GR

nickjbenson:
This PR:

- [ ] Fixes GR compilation for Unity 2018.2
- [ ] Should still allow GR compilation for previous versions of Unity

To test:
- [x] Validate Graphic Renderer test scenes run in-editor in 2018.2 and 2018.1
- [x] Validate Graphic Renderer modules don't break builds in 2018.2 and 2018.1


testmasterbright:
@hideturtle is going to look at this today.

PR #1053, opened 2018-07-11
Merge develop with master so master can be the new develop

nickjbenson:
Per our plan to keep Applications Team work closer to devs, master is going to become the main working branch for UnityModules.

Generally, our release workflow won't change much, but the main branch of the repository will no longer be the "frozen" branch that code-matches released modules. The GitHub Releases system still allows us to tie releases with commit hashes, so we don't lose any functionality.

We gain the ability to direct bleeding-edge developers to the repository without the additional step of switching to an irregular branch.

@testmasterbright, I'm adding your review for visibility, but no need to review code or test. `master` will basically act like our `develop` used to. Release reviews will still happen in release PRs on feature-freeze branches ultimately targeting `master`.

@jselstad We shouldn't need to review this code because it was already reviewed to get PR'd into develop. Feel free to peruse what the changes to master are going to look like, though.


testmasterbright:
So when do we decide to release then if things are released all the time? What happens the next time we have to implement breaking changes? Will we have one giant PR of features that could cause internal confusion and take way longer to vet instead of having them split into multiple PRs that can be merged without affecting users? Seems like shifting this way will leave us open for fires to jump on because a bug was introduced and developers are complaining. This could potentially mean taking someone off of a higher priority task to play firefighter. No one likes that context switch. We need a "stable" branch to fall back on to point users to. Shifting to this model also seems like it would require more QA involvement on a daily/weekly basis to cut down on "fires" as a result of PRs that break things and were not fully checked by the code reviewer. I don't understand why its hard to point developers to the Develop branch and why this is seen as an unneeded extra step. Most successful companies have a Develop/Master/Production branch. What are the other Pros for this shift? All I see are Cons.

testmasterbright:
We discussed the pros/cons offline and we agreed to give this a shot to simplify things. We do accept that this could potentially require more frequent QA if commits are incoming more frequently in the future as well as the potential fires that could be created. If/when issues arise, we will fix them accordingly based on severity, as always. This does mirror what we do with Platform internally and that has worked well.

PR #1063, opened 2018-10-16
Fix OnGraspStay() running all the time

off99555:
OnGraspStay() always runs no matter you grasp or not. This will fix it.


nickjbenson:
Nice catch!

PR #1089, opened 2019-07-22
[WIP] General Updates

jselstad:
This is a general maintenance update.  This is phase 1 of 2, where the second phase will complete in moving `LeapMotion` to the Plugins folder.

- Add the `Geometry` module, along with a generalized `Drawing` class.
- Fix warnings up through 2019.2
- Fix the Scriptable Render Pipeline Event Subscription (LWRP and HDRP!)


Assets/LeapMotion/Core/Scripts/Animation/CubicHermiteSpline/HermiteSpline3.cs
```diff
@@ -95,16 +95,21 @@ public struct HermiteSpline3 {
...
-      float i2 = i * i;
-      float i3 = i2 * i;
-
-      Vector3 h00 = (2 * i3 - 3 * i2 + 1) * pos0;
-      Vector3 h10 = (i3 - 2 * i2 + i) * (t1 - t0) * vel0;
-      Vector3 h01 = (-2 * i3 + 3 * i2) * pos1;
-      Vector3 h11 = (i3 - i2) * (t1 - t0) * vel1;
-
-      return h00 + h10 + h01 + h11;
+      if (t > t1) {
```
nickjbenson: If you read the summary for this class, it says "t is clamped in the t0-t1 range". Not sure if you added this or if I did.. but we should make sure the summary tag isn't lying, either remove this part, or change the summary to say that t > t1 will extrapolate the evaluated point out linearly using the velocity at t1.

Assets/LeapMotion/Core/Scripts/Geometry/Shapes/RoundRect.cs
```diff
@@ -0,0 +1,9 @@
+﻿
+
+namespace Leap.Unity.Geometry {
+  
+  public struct RoundRect {
```
nickjbenson: I think we should delete this one because it never got implemented

Assets/LeapMotion/Core/Scripts/Hands/RiggedFinger.cs
```diff
@@ -73,20 +72,7 @@ public class RiggedFinger : FingerModel {
               // the standard "test" edit-time hand model from the TestHandFactory.
               var boneTipPos = GetJointPosition(i + 1);
               var boneVec = boneTipPos - boneRootPos;
-
-              // If the rigged hand is scaled (due to a scaled rig), we'll need to divide
```
nickjbenson: mm, this code seems like it is a UnityModules-but-not-Sim fix that we should keep and not remove?

jselstad:
Alright @nickjbenson I believe this PR is in a good place; if you could give it the final review; I'd greatly appreciate it.

gamodulon:
Hello, any updates on this?

PR #1091, opened 2019-11-29
null check for LeapController only happened in Editor, not Player.

AntonyNasce-UH:
This was causing errors in Built Unity Applications, that needed multi-leap.


jselstad:
Seems harmless 👍

PR #1093, opened 2020-01-13
Bugfixes and provider access

jcorvinus:
All of my projects use these, I figure you might find them useful. Two commits are bug fixes, one is a minor enhancement to the HandUtils class.


jselstad:
Retargeted PR for `develop` (where we typically bunch up changes prior to QA review); will review shortly...

PR #1099, opened 2020-04-07
Add MultiGraspHoldingMode setting in the Interaction Manager.

nickjbenson:
Adds in #1064 in a way that preserves original behavior by default (to avoid unexpectedly breaking existing scenes) while adding the (possibly better) behavior as a dropdown enum option in the Interaction Manager.


jselstad:
This behavior is kind of yikes seeming when it first happens, but I could see a universe in which a non-physical application designer found it preferable.

The true solution to this ambiguity is to not let the _hands themselves_ separate when multi-grasping an object, ala HL: Alyx.

PR #1111, opened 2020-05-20
Added LeapMotion.LeapCSharp assembly definition

### LINKED ISSUE #1110: Suggestion: LeapCS should have an assembly definition
Having an assembly definition for the files under the [LeapCSharp](https://github.com/leapmotion/UnityModules/tree/develop/Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp) folder would help us better integrate the modules for our use cases.

It shouldn't be too difficult to add, but I think it would be a breaking change for anyone directly referencing these classes.

StephenHodgson:
Fixes #1110

### DIFF
--- a/Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/LeapMotion.LeapCSharp.asmdef
+++ b/Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/LeapMotion.LeapCSharp.asmdef
@@ -1,0 +1,12 @@
+{
+    "name": "LeapMotion.LeapCSharp",
+    "references": [],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": true,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": []
+}
--- a/Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/LeapMotion.LeapCSharp.asmdef.meta
+++ b/Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/LeapMotion.LeapCSharp.asmdef.meta
@@ -1,0 +1,7 @@
+fileFormatVersion: 2
+guid: 2ee0542102ac1084e8d9a910885c1e81
+AssemblyDefinitionImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


zalo:
It seems relatively harmless; what is the necessity of adding an .asmdef for LeapCSharp?  It doesn't appear to specify any platforms... is it just to enable unsafe code?

StephenHodgson:
That is correct. for the unsafe code. But also to properly reference the assembly definition in Unity.

nickjbenson:
For clarity's sake, I am hesitating just merging this because I have no idea what this does to downstream code. I don't have a quick resource to validate whether this will cause other people's existing Leap demos to break, so I'm trying to evaluate that first with our AppExperiments repo, which requires an upgrade first. I'll report back when I have more information.

I'd appreciate any insight you might have on what this may or may not do to code that's already referencing Leap Motion behaviours in a Unity project, @StephenHodgson

StephenHodgson:
Makes sense, no worries here. Not in a rush.

I'd even suggest moving the whole project structure to align with upm packaging and people could use the repository directly in their projects.

nickjbenson:
I very much want to do that. When I last looked, upm was still kind of a Unity-first-party-package only deal with fledgling support for other projects. Is this more of a normal thing now? Are there good reference repos we can look at?

StephenHodgson:
Yes my [Open Source project](https://github.com/XRTK/XRTK-Core) uses it extensively with a custom upm server, although really just keeping it in github is enough for projects based on 2019.3+

We even have it setup with Azure Devops CI/CD to send off preview packages when our master branch has an active PR from development.

StephenHodgson:
>I'd appreciate any insight you might have on what this may or may not do to code that's already referencing Leap Motion behaviours in a Unity project

Afaik, it would only require libraries directly calling the native wrapper to add the assembly reference (If they're already using an assembly definition). If no assembly definition exists, then it should auto-reference via the default CSharp-Assembly generated by Unity.

StephenHodgson:
Although tbh it's probably best to wrap the sdk components people are using with a separate assembly definition as well, but that would have been more of a breaking change and outside the scope of the requirements I have for the work I wanted to do, and at that point might have well restructured the project to use the upm package folder layout as well.

nickjbenson:
OK, AppExperiments appears to have no trouble here. Merging. @StephenHodgson before I close the PR thread: does XRTK need a release or other additional action for this?

StephenHodgson:
Nope we're good here. We just need to update our submodule commit pointer and we're good to go. Much appreciated.

PR #1114, opened 2020-06-10
[FEAT] Implement Server Namespaces

zalo:
This PR implements an internal feature which allows different versions of the service to be run and streamed into Unity simultaneously.

Once that's done, you can run the MultipleServiceExample in the root Unity directory.  It has two LeapXRServiceProviders preconfigured to stream data from the `v4` and `v5` server namespaces simultaneously.

Once we're ready to merge, I'll remove the example scene.


zalo:
Superseded by #1138

PR #1116, opened 2020-07-24
2020.1 Support

zalo:
These code changes bring the UnityModules _package_ to full 2020.1 compatibility.

However there's a catch: the 2020.1 Oculus XR Package is not backwards compatible with the UnityModules _project_.

To ensure _project_ compatibility with prior versions of Unity, users of 2020.1 will need to install the OculusXR plugin manually via the package manager.


nickjbenson:
- Verified that importing Core in 2019.2 produces no errors.
- Verified that importing Core in 2020.1 produces an error that is resolved by importing the Oculus XR plugin. I'll write this in our release notes for the package.

Everything looks good, I am going to merge and move through the process and release the main modules in this state as 4.5.1.

PR #1119, opened 2020-08-12
Feature rigel interaction volume uca 200

MaxPalmer-UH:


Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -77,63 +97,641 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+
+    private void ParseRigelInteractionMeshData() {
+
+      if (rigelInteractionZoneMesh == null) {
+        try
+        {
+          ObjFileParser rigelMeshDataParser = new ObjFileParser();
+
+          // This data is not oriented or scaled correctly yet.
+          rigelInteractionZoneMesh = rigelMeshDataParser.FromObj(Path.Combine(Application.dataPath, "UnityModules", "Assets", "Plugins", "LeapMotion", "Core", "Models", "Rigel-interaction-cone-placeholder.obj"),
```
zalo, line 127: Loading an .obj via full path parsing is obscenely brittle; so much so that it simply doesn't work on my machine.
The result of that `Path.Combine()` on my end is:
`C:/Repositories/UnityModules/Assets\UnityModules\Assets\Plugins\LeapMotion\Core\Models\Rigel-interaction-cone-placeholder.obj`
While mixed forward/backward slash paths can be fixed with `myPath = Path.GetFullPath(myPath);`, the correct way to do all this is to use the fact that Unity can parse/import .obj models directly.

Replace it and all the OBJ Parsing code with something like:
[Mesh rigelInteractionZoneMesh = (Mesh)AssetDatabase.LoadAssetAtPath(Path.Combine("Assets", "Plugins", "LeapMotion", "Core", "Models", "Rigel-interaction-cone-placeholder.obj"), typeof(Mesh));](https://docs.unity3d.com/ScriptReference/AssetDatabase.LoadAssetAtPath.html)

And then use the fact that Unity pre-parses it to just iterate over all the triangle connections to get the "edges".

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -77,63 +97,641 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+
+            if (angle >= Math.Abs(90)) // Was <= if YZ flipping is not on 
+            {
+              Handles.color = new Color(shadingFactor, shadingFactor, shadingFactor, shadingFactor * PEAK_INTERACTION_VOLUME_OPACITY);
+              Handles.DrawAAPolyLine(edgeVertex0, edgeVertex1);
+            } else // Edge is effectively a backface edge
+            {
+              Handles.color = BackfaceEdgeColour;
+              Handles.DrawAAPolyLine(edgeVertex0, edgeVertex1);
+            }
```
zalo, line 224: This is a fun way to achieve the "cell-style" appearance, but there's an unfortunate side-effect to drawing lines one at a time this way:
image
It's horrifically slow!   The editor slows down to a crawl when the wireframe is displayed.

When `DrawAAPolyLine()` is called with only two arguments, it enqueues a single drawcall to the command buffer to execute drawing.  This is fine if it's only a single digit number of line segments, but we've got 4671 edges (and thus 4671 draw calls) being enqueued!

The easy fix is to queue up your edges into a giant Vector3[] and [call `DrawAAPolyLine()` with that array as an input argument.](https://docs.unity3d.com/ScriptReference/Handles.DrawAAPolyLine.html)  This will draw all of the lines as a single draw call.

The side effect of this is that you won't be able to change the color of each line segment at a granular level.  

The proper fix is to draw the cage with [`Graphics.DrawMeshNow()`](https://docs.unity3d.com/ScriptReference/Graphics.DrawMeshNow.html) and to use a Shader+Material to create the Transparency look that you're going for.

So you'd first create the material, and then you'd replace this whole edge iteration loop with something like:
```
// On Initialization
      rigelInteractionMaterial = (Material)AssetDatabase.LoadAssetAtPath(Path.Combine("Assets", "Plugins", "LeapMotion", "Core", "Materials", "Default.mat"), typeof(Material));
...
// Then later inside of DrawRigelInteractionZoneMesh()
      rigelInteractionMaterial.SetPass(0);
      Graphics.DrawMeshNow(rigelInteractionMesh, Matrix4x4.TRS(Vector3.zero, Quaternion.Euler(-90,0,0), Vector3.one*0.001f));
```

Assets/Plugins/LeapMotion/Core/Scripts/LeapServiceProvider.cs
```diff
@@ -534,8 +545,9 @@ public enum PhysicsExtrapolationMode {
     protected void onHandControllerConnect(object sender, LeapEventArgs args) {
       initializeFlags();
 
+      // TO REVIEW
       if (_leapController != null) {
-        _leapController.Device -= onHandControllerConnect;
+        // _leapController.Device -= onHandControllerConnect;
```
zalo, line 550: Looking upstream from this event, it doesn't look like this causes any lasting issues when using a "Single-Device" Service, so I believe it's fine to leave in (pending a second opinion from @nickjbenson ).

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -61,79 +80,205 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+        default:
+          break;
+      }
+
+    }
+
+    private void ParseRigelInteractionMeshData() {
+
+      if (_rigelInteractionZoneMesh == null) {
+        _rigelInteractionZoneMesh = (Mesh)AssetDatabase.LoadAssetAtPath(Path.Combine("Assets", "UnityModules", "Assets", "Plugins", "LeapMotion", "Core", "Models", "Rigel-interaction-cone-placeholder.obj"), typeof(Mesh));
```
zalo, line 122: There is an extra `"Assets", "UnityModules",` prepended to this Path (and thus the Editor is not able to find it)...

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -61,79 +80,205 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+        // There is a bug/feature where the asset database appears to cache changes to assets, so loading an
+        // asset again (when _rigelInteractionZoneMesh is null) causes it to retain the flipped normals 
+        // it and will then flip them again. There appears to be no way to clear this cache, so here we
+        // check the first normal in the mesh to see if we need to flip it. 
+       // if (_rigelInteractionZoneMesh.normals[0].ApproxEquals(new Vector3(-1.0f, 0.1f, 0)))
+          ReverseNormals();
+      }
+
+      if (_rigelInteractionMaterial == null) {
+        _rigelInteractionMaterial = (Material)AssetDatabase.LoadAssetAtPath(Path.Combine("Assets", "UnityModules", "Assets", "Plugins", "LeapMotion", "Core", "Materials", "RigelInteractionVolume.mat"), typeof(Material));
```
zalo, line 133: Same for this Path.... I'm curious, how is your project/folder hierarchy set up that this presumably works on your machine?

Assets/Plugins/LeapMotion/Core/Shaders/InteractionVolumeWireframe.shader
```diff
@@ -0,0 +1,100 @@
+﻿Shader "LeapMotion/Geometry/Wireframe"
```
zalo, line 1: The `.meta` files for all of the added assets were ignored in the commits.  This is bad; they should be committed, as they are how Unity stores asset GUIDs and defines inter-file connections.   It is particularly egregious for this file because the Interaction Zone Material contains a broken reference to this shader, which prevents the interaction volume from drawing.   Please commit the generated `.meta` files with the assets!

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -61,79 +80,205 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+
+      if (_rigelInteractionZoneMesh == null) {
+        _rigelInteractionZoneMesh = (Mesh)AssetDatabase.LoadAssetAtPath(Path.Combine("Assets", "UnityModules", "Assets", "Plugins", "LeapMotion", "Core", "Models", "Rigel-interaction-cone-placeholder.obj"), typeof(Mesh));
+
+        // There is a bug/feature where the asset database appears to cache changes to assets, so loading an
+        // asset again (when _rigelInteractionZoneMesh is null) causes it to retain the flipped normals 
+        // it and will then flip them again. There appears to be no way to clear this cache, so here we
+        // check the first normal in the mesh to see if we need to flip it. 
+       // if (_rigelInteractionZoneMesh.normals[0].ApproxEquals(new Vector3(-1.0f, 0.1f, 0)))
+          ReverseNormals();
```
zalo, line 129: This step should not be necessary!  The normals on the mesh are fine when they're imported (as Unity automatically calculates them upon initial import).  See the comment in the shader for more details....

Assets/Plugins/LeapMotion/Core/Shaders/InteractionVolumeWireframe.shader
```diff
@@ -0,0 +1,100 @@
...
+            };
+
+            v2g vert(appdata_base v) {
+                v2g o;
+                o.worldPos = mul(unity_ObjectToWorld, v.vertex);
+
+                float3 worldNorm = UnityObjectToWorldNormal(v.normal);
+                float3 cameraForwardVector = mul((float3x3)unity_CameraToWorld, float3(0, 0, 1));
+
+                o.cameraToSurfaceNormalAngleDegrees = degrees(acos(dot(normalize(worldNorm), normalize(cameraForwardVector))));
```
zalo, line 41: Here you can see that you're calculating the angle between the surface normal and the cameraForwardVector, which is almost right, except it means you're going to get an angle of 180 degrees when you're looking right at the normal.

The simple fix is to add a `-` in front of `worldNorm`.   This ensures that the shader works properly without any mesh-level normal flipping required.

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -61,79 +80,205 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+    private LeapServiceProvider.InteractionVolumeVisualization? GetSelectedInteractionVolume() {
+
+      return LeapServiceProvider?.SelectedInteractionVolumeVisualization;
+    }
+
+    private void DrawRigelInteractionZoneMesh() {
+
+      if (_rigelInteractionMaterial != null && _rigelInteractionZoneMesh != null) {
+        _rigelInteractionMaterial.SetPass(0);
+        Graphics.DrawMeshNow(_rigelInteractionZoneMesh, Matrix4x4.TRS(_rigelInteractionZoneMeshOffset, Quaternion.Euler(-90, 0, 0), Vector3.one * 0.001f));
```
zalo, line 206: The transform which the mesh is drawn at is always at the origin.  So if the ServiceProvider moves, the mesh will stay behind.

You can fix this by replacing `Matrix4x4.TRS(_rigelInteractionZoneMeshOffset, Quaternion.Euler(-90, 0, 0), Vector3.one * 0.001f)` with `target.transform.localToWorldMatrix * Matrix4x4.TRS(_rigelInteractionZoneMeshOffset, Quaternion.identity, Vector3.one * 0.001f)`.  This will move the interaction volume to always be in front of the service provider.

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -6,26 +6,45 @@
...
-    private const float BOX_RADIUS = 0.45f;
-    private const float BOX_WIDTH = 0.965f;
-    private const float BOX_DEPTH = 0.6671f;
+    private const float LMC_BOX_RADIUS = 0.45f;
+    private const float LMC_BOX_WIDTH = 0.965f;
+    private const float LMC_BOX_DEPTH = 0.6671f;
+
+    private Mesh _stereoIR170InteractionZoneMesh;
+    private Material _stereoIR170InteractionMaterial;
+    private readonly Vector3 _stereoIR170InteractionZoneMeshOffset = new Vector3(0.0523f, 0, 0.005f);
```
zalo, line 36:
```suggestion
```
This shouldn't be necessary (see next suggestion).

Assets/Plugins/LeapMotion/Core/Editor/LeapServiceProviderEditor.cs
```diff
@@ -61,79 +80,187 @@ public class LeapServiceProviderEditor : CustomEditorBase<LeapServiceProvider> {
...
+      return LeapServiceProvider?.SelectedInteractionVolumeVisualization;
+    }
+
+    private void DrawStereoIR170InteractionZoneMesh() {
+
+      if (_stereoIR170InteractionMaterial != null && _stereoIR170InteractionZoneMesh != null) {
+        _stereoIR170InteractionMaterial.SetPass(0);
+
+        Graphics.DrawMeshNow(_stereoIR170InteractionZoneMesh,
+           target.transform.localToWorldMatrix * Matrix4x4.TRS(_stereoIR170InteractionZoneMeshOffset, Quaternion.Euler(-90, 0, 0), Vector3.one * 0.001f));
```
zalo, line 194: The Service Providers contain built-in offsetting systems (which modify the "device origin", affecting the interaction zone).   

Luckily, it's pretty easy to use these:
```suggestion
           target.transform.localToWorldMatrix * Matrix4x4.TRS(controllerOffset, deviceRotation * Quaternion.Euler(-90, 0, 0), Vector3.one * 0.001f));
```

zalo (approved):
Looks entirely good to me!

PR #1121, opened 2020-09-08
Connection stores an incorrect IntPtr to the Device

lazloringUH:
In Connection.cs, the IntPtr that get's stored in the Device class is incorrect, resulting in future GetDeviceInfo calls from within that class to have incorrect references.

Quick example:
Call LeapC.GetDeviceInfo(Handle, ref deviceInfo); from within Device.cs, it will always return a status code of 0.


zalo:
Good Catch!  Thank you for spotting this one!

PR #1123, opened 2020-09-22
Error Checking/IntPtr fixing

lazloringUH:
Added the following error checking:
IsStreaming
IsSmudged
IsLowResource
IsLightingBad

Added a Secondary IntPtr to account for different refs between event's and internal struct functions.


zalo:
These changes work on my end and the code is clean.  Good PR 👍

PR #1128, opened 2020-10-09
More than two devices supported

novacto3:
The commit fixes KabschSolver arguments because the Solver did not work properly when more than two devices were connected.


novacto3:
I also updated the calibration info in the scene to take more than two devices into account.

PR #1129, opened 2020-10-13
Feature deviceinfo reference fix

lazloringUH:
Added in support for DeviceStatusChange events, moved previous device status messages to support new event, removed bad entrypoint in LeapC and replaced it with working pause call.


zalo:
This appears to work well for me; I was successfully able to detect the `IsSmudged` event under the new paradigm.

Setting `SetPause()` has no noticeable effect, but it never did (and it's unused), so this may be intended behavior.

Merging.

PR #1130, opened 2020-10-20
Added example of pinch locomotion

fmaurer:


fmaurer:
image

zalo:
Closing this for now; eventually I'd like to give locomotion a proper scene (perhaps even its own module) with multiple alternatives and more options.

I'm also worried the script isn't _quiite_ robust enough to call into the spotlight just yet (there are some micro jumps that happen upon unpinching that I'd like to dig more deeply into).

PR #1132, opened 2020-10-21
Call dispose methods when shutting down

robert-evans-ultraleap:
This prevents a hang on shutdown running the VRVisualizer with the Gen 2 platform


zalo:
Interesting; it seems like this line in `Connection.cs` is what it supposed to call `Dispose()`:
```csharp
AppDomain.CurrentDomain.DomainUnload += (arg1, arg2) => Dispose(true);
```

But I guess Gen2 may have a race condition where this callback's ordering is not ideal.   Since this doesn't seem to cause issues in the editor or builds (upon pause or close) in the current Public Service, I'm merging this.

PR #1133, opened 2020-10-22
[FEAT] Simplify the Topology of the Rigel Interaction Cone Mesh

zalo:
This PR will vastly simplify and retopologize the SIR170's Interaction Cone mesh for easier viewing (thank you, @flo !)

image


zalo:
Thank you for flipping the normals on that!  Merging...

PR #1134, opened 2020-11-02
Add Screentop support LeapC 4.9.2

Niram7777:
- Screentop tracking mode added (the view will be identical to Desktop though)
- Replace the setPolicy(HMD) by the Tracking Mode setters.
- Updated the LeapC.dll from LeapDeveloperKit_4.9.2+52299_win

NB : even if HMD/Screentop policies are still there they should not be used anymore!


Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/Connection.cs
```diff
@@ -888,4 +902,12 @@ private void handleDeviceStatusEvent(ref LEAP_DEVICE_STATUS_CHANGE_EVENT statusE
       _lastResult = result;
     }
   }
+
+  public UInt64 SetTrackingMode(eLeapTrackingMode mode) {
```
AlexKentUL, line 906: In C# you cannot have free functions inside a namespace - these will need to go into a class.

Assets/Plugins/LeapMotion/Internal/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
```diff
@@ -91,26 +106,31 @@ void Update()
...
-        else
-        {
-          m_trackingText.text += " (Press '" + keyToToggleHMD + "' to switch to head-mounted mode)";
-          if (Input.GetKeyDown(keyToToggleHMD)) {
-              m_controller.SetPolicy(Controller.PolicyFlag.POLICY_OPTIMIZE_HMD);
-          }
-        }
-      } 
+        case eLeapPolicyFlag.eLeapTrackingMode_Desktop:
+          m_trackingText.text += "Desktop";
```
AlexKentUL, line 113: I haven't tried compiling this yet, but this line (and the ones in the other cases) probably shouldn't end with a semi-colon (the next line needs to join on).

Consider using formatted strings: https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=netcore-3.1.

Assets/Plugins/LeapMotion/Core/Plugins/LeapCSharp/Connection.cs
Niram7777: It was indeed not in the Connection class (move one lined above)

Assets/Plugins/LeapMotion/Internal/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
Niram7777: Indeed forgot to fuse the instructions.

Let me know if you are happy now with formating

AlexKentUL:
I have tested this locally and it looks good.

AlexKentUL (approved):

PR #1135, opened 2020-11-03
Remove ABI breaking tracking mode event

Niram7777:
Remove the setTrackingMode that sends a new message type event

### DIFF
--- a/Assets/Plugins/LeapMotion/Internal/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
+++ b/Assets/Plugins/LeapMotion/Internal/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
@@ -9,7 +9,6 @@
 using UnityEngine;
 using UnityEngine.SceneManagement;
 using Leap;
-using LeapInternal;
 using System;
 
 namespace Leap.Unity.VRVisualizer{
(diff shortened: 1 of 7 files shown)


AlexKentUL (approved):

Assets/Plugins/LeapMotion/Internal/VRVisualizer/ScenesResources/Scripts/VisualizerManager.cs
```diff
@@ -62,7 +61,7 @@ private void goScreenTop()
         m_PCVisualizer.gameObject.SetActive(true);
         m_VRVisualizer.gameObject.SetActive(false);
         m_warningText.text = "ScreenTop tracking mode activated";
-        m_controller.SetTrackingMode(eLeapTrackingMode.eLeapTrackingMode_ScreenTop);
+        m_controller.SetPolicy(Controller.PolicyFlag.POLICY_OPTIMIZE_SCREENTOP);
```
MaxPalmer-UH, line 64: It's a bit strange that we set the policy flag here for screentop, when we don't set the flag for HMD in goVR

Niram7777:
We are going to revert develop, and put the screentop work on a new branch

PR #1142, opened 2020-12-09
Removed reference to Orion in warning message

jj-tetraquark:
As referenced by GTPS-63 on JIRA


zalo:
I'm changing the target of this to the Screentop PR (Pull Request) because of a PR (Public Relations) snafu where V4 is actually still considered to be a part of the "Orion" beta (and non-HMD Tracking doesn't actually get better until the ScreenTop/V5 version of Platform anyway).

And merging.

PR #1145, opened 2020-12-18
[FEAT] Adding new Hand Binder scripts which enable Hand Rigging

ElliotPadfieldUL:


Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+namespace Leap.Unity.HandsModule {
+
+    public static class HandBinderAutoRigger {
+
+        /// <summary>
+        /// This function is used to search the HandBinder scipts children transforms to auto assign them for the user
+        /// </summary>
+        /// <param name="handBinder">The binder that the found transforms will get assigned too</param>
+        public static void AutoRig(HandBinder handBinder) {
+            handBinder.SetEditorPose = false;
```
MattGrayUL, line 14: Why set to false to then set to true at the end? it is not used elsewhere in this function before being set to true always

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// Get all the children of a transform
+        /// </summary>
+        /// <param name="_t"></param>
+        /// <returns></returns>
+        public static List<Transform> GetAllChildren(Transform _t) {
+            List<Transform> ts = new List<Transform>();
+            foreach(Transform t in _t) {
+                ts.Add(t);
+                if(t.childCount > 0)
+                    ts.AddRange(GetAllChildren(t));
```
MattGrayUL, line 65: Missing braces

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// <returns></returns>
+        private static Transform[] SelectBones(List<Transform> children, string[] definitions, bool isThumb = false) {
+            //Can only ever be 4 bones per hand
+            var bones = new Transform[4];
+            int foundBonesIndex = 0;
+            for(int i = 0; i < definitions.Length; i++) {
+                foreach(var child in children) {
+                    //We have found all the bones we need
+                    if(foundBonesIndex == 4)
+                        break;
```
MattGrayUL, line 85: Missing braces

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// </summary>
+        /// <param name="children">The found Children</param>
+        /// <param name="definitions">The definition to sort through the children</param>
+        /// <param name="isThumb">is this a thumb?</param>
+        /// <returns></returns>
+        private static Transform[] SelectBones(List<Transform> children, string[] definitions, bool isThumb = false) {
+            //Can only ever be 4 bones per hand
+            var bones = new Transform[4];
+            int foundBonesIndex = 0;
+            for(int i = 0; i < definitions.Length; i++) {
```
MattGrayUL, line 81: can 'i' be more explicitly 'definitionIndex'?

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+
+            if(middleProximal != null && indexProximal != null && pinkyProximal != null && wrist != null) {
+                //Get the Direction from the middle finger to the wrist
+                var wristForward = middleProximal.position - wrist.position;
+                //Get the Direction from the Proximal pinky finger to the Proximal Index finger
+                var wristRight = indexProximal.position - pinkyProximal.position;
+
+                //Swap the direction based on left and right hands
+                if(handBinder.Handedness == Chirality.Right)
+                    wristRight = -wristRight;
```
MattGrayUL, line 182: Missing braces

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
```diff
@@ -0,0 +1,208 @@
...
+                var hand = TestHandFactory.MakeTestHand(handBinder.Handedness == Chirality.Left, unitType: TestHandFactory.UnitType.LeapUnits);
+                hand.Transform(TestHandFactory.GetTestPoseLeftHandTransform(TestHandFactory.TestHandPose.DesktopModeA));
+                var leapRotation = hand.Rotation.ToQuaternion();
+
+                //Get the rotation of the calculated hand Basis
+                var modelRotation = Quaternion.LookRotation(wristForward, wristUp);
+
+                //Now calculate the difference between the models rotation and the leaps rotation
+                var wristRotationDifference = Quaternion.Inverse(modelRotation) * leapRotation;
+                var wristRelativeDifference = (Quaternion.Inverse(wrist.rotation) * wristRotationDifference).eulerAngles;
```
MattGrayUL, line 200: A lot of 'var's here, makes it hard to follow

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderBoneDefinitions.cs
```diff
@@ -0,0 +1,24 @@
...
+    [System.Serializable]
+    public class BoneDefinitions {
+        public string[] DefinitionThumb = { "thumb" };
+        public string[] DefinitionIndex = { "index" };
+        public string[] DefinitionMiddle = { "middle" };
+        public string[] DefinitionRing = { "ring" };
+        public string[] DefinitionPinky = { "pinky", "little" };
+        public string[] DefinitionWrist = { "wrist", "hand", "palm" };
+        public string[] DefinitionElbow = { "elbow", "upperArm" };
+        //public string[] DefinitionShoulder = { "shoulder", ""};
```
MattGrayUL, line 22: Perhaps remove this and re-add it in future if/when we add shoulders

MattGrayUL (changes requested):
Looks generally good, just some housekeeping and maybe a little cleaner way to do a few things now that the scope has been decided

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderDocumentationWindow.cs
```diff
@@ -0,0 +1,302 @@
...
+        [MenuItem("Ultraleap/Hand Rigging Documentation")]
+        private static void Init() {
+            leapController = AssetDatabase.LoadAssetAtPath("Assets/Plugins/LeapMotion/Core/Prefabs/LeapHandController.prefab", typeof(Leap.Unity.HandModelManager));
+            // Get existing open window or if none, make a new one:
+            var window = (HandBinderDocumentationWindow)EditorWindow.GetWindow(typeof(HandBinderDocumentationWindow));
+            window.Show();
+        }
+
+        private void OnGUI() {
+            SetUp();
```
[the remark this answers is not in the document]
ElliotPadfieldUL: This cant be done because the parameters of the style need to be called from a GUI Function.

MattGrayUL (approved):
All the changes I requested have been made

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderDocumentationWindow.cs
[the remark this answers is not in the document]
TheMunro: Any reason why the properties are all static? Also, could just null check them so they only initialise once?

e.g.
```
        if(headerStyle == null){
            headerStyle = new GUIStyle() {
                alignment = TextAnchor.MiddleCenter,
                fontStyle = FontStyle.Bold,
                fontSize = Screen.width / 25,
                normal = new GUIStyleState() {
                    textColor = Color.white,
                }
            };
        }
```

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs
```diff
@@ -0,0 +1,718 @@
...
+                fixedHeight = 20,
+            };
+
+            previousCol = GUI.color;
+        }
+
+        /// <summary>
+        /// Draw the inspector GUI
+        /// </summary>
+        public override void OnInspectorGUI() {
```
TheMunro, line 167: There is an awful lot happening in this function. Maybe use your IDE refactoring tools to extract some methods as subfunctions or local functions so that you can give things names to explain what they draw (i.e. the code self-documents itself that way)?

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs
```diff
@@ -0,0 +1,718 @@
...
+                    break;
+
+                default:
+                    break;
+            }
+
+            return offsetProperty;
+        }
+
+        private SerializedProperty OffsetPropertyFromLeapTypes(int fingerType, int boneType) {
```
TheMunro, line 620: What about changing this function to 

```private SerializedProperty OffsetPropertyFromLeapTypes(Finger.FingerType fingerType, Finger.BoneType boneType) ```

then do the int casts in here?

File: Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs [hunk 11]
```diff
@@ -0,0 +1,718 @@
...
+            }
+
+            binder.LeapHand = hand;
+        }
+
+        /// <summary>
+        /// Converts
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
```
TheMunro, line 519: Could make create a dictionary of BoundType to a tuple:

```
Dictionary<BoundType, (Finger.FingerType, Finger.BoneType)> map = new Dictionary<BoundType, (Finger.FingerType, Finger.BoneType)>{
    {BoundTypes.THUMB_METACARPAL, (Finger.FingerType.TYPE_THUMB, Bone.BoneType.TYPE_METACARPAL)}
    //etc
};

private SerializedProperty BoundTypeToProperty(int id)
{
    SerializedProperty offsetProperty = null;
    var boundType = (BoundTypes)id;
   if(!map.ContainsKey(boundType))
        return null;

    var (fingerType, boneType) = map[boundType];
    return OffsetPropertyFromLeapTypes(fingerType, boneType);
}
```

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinder.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        //Reset is called when the user hits the Reset button in the Inspector's context menu or when adding the component the first time.
+        private void Reset() {
+            ResetHand();
+        }
+
+        /// <summary>
+        /// Update the BoundGameobjects so that the positions and rotations match that of the leap hand
+        /// </summary>
+        public override void UpdateHand() {
```
TheMunro, line 87: Maybe add some comments here? Seems like a bit of a wall of text!

Packages/manifest.json
```diff
@@ -2,13 +2,13 @@
...
-    "com.unity.package-manager-ui": "2.2.0",
-    "com.unity.test-framework": "1.0.18",
+    "com.unity.ide.vscode": "1.2.1",
+    "com.unity.test-framework": "1.1.14",
     "com.unity.textmeshpro": "2.0.1",
-    "com.unity.timeline": "1.1.0",
-    "com.unity.xr.oculus.standalone": "1.38.4",
-    "com.unity.xr.openvr.standalone": "1.0.5",
+    "com.unity.timeline": "1.2.6",
+    "com.unity.xr.arfoundation": "2.1.10",
```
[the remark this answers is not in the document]
TheMunro: Also remove the following as it screwed up Rider integration for me.

```"com.unity.ide.vscode": "1.2.3",```

File: Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs (hunk 9 above)
Comment (ElliotPadfieldUL) [reply Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs:167]: Great idea, I have split them into subfunctions as suggested.

File: Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs (hunk 11 above)
Comment (ElliotPadfieldUL) [reply Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/Editor/HandBinderEditor.cs:519]: Fantastic idea, this cleaned up the code a lot and I was looking for a solution like this.

**Review by TheMunro — APPROVED** [review]

File: Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs [hunk 14]
```diff
@@ -0,0 +1,217 @@
...
+            return newBone;
+        }
+
+        /// <summary>
+        /// Calculate the rotation offset needed to get the rigged hand into the same orientation as the leap hand
+        /// </summary>
+        public static void CalculateWristRotationOffset(HandBinder handBinder) {
+            Transform middleProximal = handBinder.boundHand.fingers[2].boundBones[1].boundTransform;
+            Transform indexProximal = handBinder.boundHand.fingers[1].boundBones[1].boundTransform;
+            Transform pinkyProximal = handBinder.boundHand.fingers[4].boundBones[1].boundTransform;
```
zalo, line 179: It seems like the wrist rotation misalignment I notice can be fixed by changing the `boundBones` indices from `[1]` to `[0]`.

This is an example of the change:
BadToGoodWristOrientation

zalo (changes requested):
I like where this is going; it's a big improvement over what we currently have!

1) The changes to the `manifest.json`, `packages-lock.json`, `Packages-Update.log`, and `ProjectVersion.txt` should be reverted.  Ideally, the scenes should also be opened/serialized in 2019.2 (to maintain backwards compatibility).
2) The autorigging seems to vary between the left and right hands (and rigging one breaks the other!!)   What's going on here?
Left-Right-InvarianceIsBroken
3) The Wrist Orientation Offset calculation seems to put the knuckle joints out of alignment.   This _seems_ like it can be corrected with the change to `HandRiggerAutoBinder.cs` (mentioned below in the code comments).
4) The Skeleton Hands' joints pivot around the wrong location!   Rather than pivoting about the gap between the fingers, they should instead slide around the bulbous "cap" at the end of each bone.
image

Assets/Plugins/LeapMotion/Modules/Hands/Scripts/HandBinder/HandBinderAutoRigger.cs
ElliotPadfieldUL: 1.) Will do, I was using 2019.4 instead of 2019.2 as the UI looks so gross, look at the difference between the custom editor.
image

2.) Great spot!
- So when pressing the auto rig button, the hand gets reset to its initial model pose, this is to stop re-calculating the rotation difference from the hand that is in the editor pose. 
- Then once it is  in its default pose the rotation offset is calculated and scene GUI update is triggered but the call to actually update the hand to its new pose was happening to late. So by pressing the Autorig button on the other hand, it was finally updating the hand to the new rotation offsets that were calculated the time before.

This has been fixed by simply updating the hand at the end of the AutoRig function rather than waiting for scene GUI to get updated.

3.) While I agree this does fix it, some models do not include meta finger bones which is what indices [0] is. So, as a fix I will change the code so that it tries to use indices [0] first then if these do not exist I will then use the proximal bones.

4.) Another awesome spot, I will make sure this also goes into the updated documentation that I have written so that our users are aware of the difference between pivots at the end of the previous bone vs the start of the next bone. 

https://gyazo.com/89338609e7399d84d804c40329c67dec - This is what it is like currently.
https://gyazo.com/91f558c815848b6d1425e50379ab21ba - After your suggestion

TheMunro: 1. The recommendation from Unity is to keep `packages-lock.json` under [source control](https://docs.unity3d.com/Manual/upm-conflicts-auto.html) such that the same versions are pulled down by all users. It probably doesn't really matter for the case of packages managed by Unity directly as everything uses semantic versioning so there should not be a situation where a package or its dependencies change, but it is relevant in the case where git repositories are added since it uses a specific commit hash in the lock file to mark the required version.

Are `"com.unity.xr.oculus.standalone"` and `"com.unity.xr.openvr.standalone"` dependencies that have been intentionally added? If so I guess the best course of action would be to revert `manifest.json` entirely to get back to the base dependencies and undo any automatic package updates between 2019.2 and 2019.4, then optionally add `""com.unity.xr.legacyinputhelpers"` (instead of `"com.unity.xr.arfoundation"`). I don't think things like `"com.unity.ide.vscode"` should be in there by default though unless we add in adapters for all relevant IDEs.

ElliotPadfieldUL:
> All those changes look good (in 2019.2); thank you!
> 
> However, I'm now noticing a new weirder behavior:
> If I open this scene up in 2019.4 now (for forward compatibility checking) all of the HandBinder etc. scripts have disappeared from the scene/prefabs... How were these prefabs ported/reserialized to 2019.2 in the first place?
> 
> It may be necessary to rebuild the assets from scratch in 2019.2 so that their automatic reserialization going forwards knows how to handle them properly...?

So I did a bit of testing this morning, and strangely simply updating the prefabs in 2019.2 was not enough. What I ended up having to do was remove all the models and prefabs in 2019.2, then re-importing the models and create the prefabs again. I could then open them in 2019.4 and all the scripts stayed on the prefabs.

zalo (approved):
Just tested in 2019.2 and 2019.4

Looks good to me!

PR #1146, opened 2021-01-18
[FEAT] Support Unity 2018.4 LTS in ScreenTop-Branch

MaxPalmer-UH:
@zalo  I've spent the day moving back and forth between different Unity versions both testing and improving the user experience in 2018.4 and checking that it doesn't cause any serious side effects. To this end I've created this branch off the feat-screentop branch which forms the basis of the 4.7.0 UnityModules release.  It's worth noting what I've found and changed on this branch (roughly in line with the commits below).

- I've updated the root level README to add a note of Unity 2018.4 support:
"Notes on Unity 2018.4 (LTS).
It is not possible to support Unity 2018 LTS -> Unity 2020.2 in a single package. This release is optimised for Unity 2019.4. However, it is possible to use it in Unity 2018.4.
To do this delete the manifest.json file in the packages folder. Also delete the LeapMotion.asmdef file inside Assets/Plugins/LeapMotion/Core/Scripts and then open the project. Note, a couple of the scenes will not work in this version of Unity. These are **FingerDetectionDemo** and **VRVisualizer**."

- 2. Following the above recommended procedure gets most of the job done. However, there were still some issues worth addressing:

a. I updated some of the assets, particularly the lightmaps that were throwing errors as they were serialised by a more recent version of Unity. These were confined to the interaction engine example scenes (scene 2, 4, 5 & 6).
b. I downgraded the project version file to reference  2018.4.30f1, so Unity 2018.4 will open it without complaining
c. Unity 2019 was showing errors about "Component of type GUILayer is no longer available...", this was affecting the GraphicRenderer based scenes and the Dynamic UI scene.
d. Two scenes were showing warnings about 'Mesh does not have normals for Global Illumination'. This appeared to be related to meshes (usually text) that were static. Changing these to non static fixes those warnings.
e. I noted that the desktop tracking scene no longer shows the handsin the game view as the camera was not in the correct location. I moved it back.

With all those changes, the user experience of moving between 2018 LTS, 2019 LTS and 2020.2 still isn't completely smooth. I noted on loading a the project from a clean fetch of the branch in each showed warnings. In **2018** two scenes have missing scripts - VRVisualizer (not important as it's internal) and FingerDetectorDemo (you lose the instructions but the gestures work). These still work fine in 2019 and 2020, so I don't think it's worth fixing them. I captured a screenshot of the warnings in 2019 and 2020 (edited) 

**Warnings on load in 2019.4:**
image

- The XR plugin warning is expected.

- The 'unreachable code detected is real', we've hardcoded a constant and then have an if clause that never gets called, based off that constant.

-  **The capsule hand material warning crops up in many scenes.** @zalo is this related to some of the performance improvements you've been making?

- Not sure where the list referenced by the 'component at level 3' warning is. I didn't spot any degraded behaviour in the examples.

**Warnings on fresh load into 2020.2:**
image (1)

Most of these warnings are the same as 2019.4. The new ones is:

- Creating missing LeapXRServiceProvider for LeapEyeDislocator in VR camera.

I went through all the scenes in 2018.4, 2019.4 and 2020.2 and they seemed OK, other than the ones called out above for 2018.4.  However, **I saw strange behaviour in the Capsule Hand (Desktop) scene which would be helpful if you could check**. My hands were not tracked correctly and appeared flipped. Matt Gray checked this scene and the hands tracked fine once the (fixed) camera issue was addressed. @zalo Could you check please?


nickjbenson:
Given that:
  - It seems like it's a lot of work to have the repo functionally work with 2018,
  - It seems like it's impossible to have a single source tree that opens in 2018 without complaints _and_ opens in 2019+ without complaints,
  - 2018 support is being dropped from LTS soon by Unity,

Does it make sense to:
  - Convert this work into a dedicated, forever-living release branch for 2018 for this release, maybe calling it `release-{version}-unity-2018` or something,
  - Have that 2018-specific branch in the "non-complaining," "clean-opening" state that you've described the process for in your notes here,
  - Keep the "primary" release branch and `develop`-to-be with e.g. 2019+ project files, lighting data, and the asmdefs and the like, without the changes for 2018 compatibility in this PR,
  - And basically never plan on updating 2018 support beyond this release, given its end-of-LTS-life from Unity

If we did that, we'd be able to keep `develop` and the "primary" release clean and they could open without any warnings and be ready-to-go for any modern projects (probably most of them? Is that actually true?) and all new projects.

MaxPalmer-UH:
@nickjbenson I'm coming round to that conclusion. I no longer believe you can have a smooth single source for 2018->2020 and trying to remove as many wrinkles as possible is painful. If we decide to merge this, then I'd like to drop 2018 support after it. A less risky approach would be to create this as a separate (and dead end) release branch for 2018 at the same time we have the screentop branch for 2019/2020.

zalo:
Just re-set this branch so it's base is the Screentop Branch (it makes the diff cleaner and easier to reason about). 

The only changes seem to be reserializing a few scenes, changing the project version, and adding that note about 2018. @nickjbenson perhaps you want to take a look again?   

I'm not really against this PR going in since into the Screentop PR (which will eventually be merged to `develop`) since it seems to add no additional (unclearable) warnings or incompatibilities.

nickjbenson:
> A less risky approach would be to create this as a separate (and dead end) release branch for 2018 at the same time we have the screentop branch for 2019/2020.

This is how I'm leaning right now too.

> The only changes seem to be reserializing a few scenes, changing the project version, and adding that note about 2018. 

My issue is exactly with the reserialization with 2018 -- I'm worried about forward compatibility beyond 2018 (in 2021 and beyond for example), and would strongly prefer we don't move backward with scene serialization versions, only forward. Additionally, if we just performed the 2018 conversion steps that @MaxPalmer-UH has lined out, then the "dead-end branch fro 2018" approach actually gets us a nice, clean, non-complaining 2018 version that we know we don't need to support further on.

With all that said, I'll review the diff in a bit and circle back around if my thinking changes on this.

zalo:
Closing this as part of the 4.7.0 Release:
https://github.com/leapmotion/UnityModules/releases/tag/UM-4.7.0

PR #1158, opened 2021-05-04
[UPGRADE] Migrate Updates from Internal Codebase

zalo:
This PR:
 - Deprecates Leap.Unity.Pose for UnityEngine.Pose
 - Adds new Utils, QueryOperatorExtensions, and Infixes
 - Adds StringPathUtils.cs - A Utility for working with String Paths

This is part of the general compatibility movement for internal codebases.


Assets/Plugins/LeapMotion/Core/Examples/Example Assets/InertiaPostProcessProvider.cs
```diff
@@ -126,17 +126,17 @@ public class InertiaPostProcessProvider : PostProcessProvider {
     private void integratePose(ref Pose curPose, ref Pose prevPose,
                                Pose targetPose, float deltaTime) {
       // Calculate motion from prevPose to curPose.
-      var deltaPose = curPose.inverse * prevPose; // prevPose in curPose's local space.
+      var deltaPose = PoseExtensions.mul(curPose.inverse(), prevPose); // prevPose in curPose's local space.
```
nickjbenson, line 129: Consider infix/chaining calls instead of manually referencing PoseExtensions (they are extension methods, right?):
`curPose.inverse().mul(prevPose)`

nickjbenson: (also applies to the other PoseExtensions calls in this file and others)

Assets/Plugins/LeapMotion/Core/Scripts/Query/QueryOperatorExtensions.cs
```diff
@@ -19,6 +19,34 @@ namespace Leap.Unity.Query {
...
+      return result;
+    }
+    /// <summary> Int32s, careful with overflow. </summary>
+    private static int Factorial(int N) {
+      int f = 1; for (var n = N; n > 1; n--) f *= n; return f;
+    }
+
+    /// <summary>
+    /// Prints a Debug.LogError if the query results contains more than a single element, but otherwise preserves the Query as-is.</summary>
+    public static Query<T> ComplainIfMany<T>(this Query<T> query, string message = null) {
```
nickjbenson, line 38: weird <summary></summary> style here (recommend sandwich both horizontally on the same line or vertically on distinct lines)
```suggestion
    /// <summary>
    /// Prints a Debug.LogError if the query results contains more than a single element, but otherwise preserves the Query as-is.
    /// </summary>
    public static Query<T> ComplainIfMany<T>(this Query<T> query, string message = null) {
```

nickjbenson (changes requested):
Couple of suggestions, feel free to resolve as you like. Code-wise this gets a "thumbs up as long as you decide one way or another on the suggestions" from me.

Let me know if you'd like me to do a smoke-test in Unity for e.g. examples. (I don't have my headset set up right now, though.)

zalo:
Good catches; I'm not sure why I thought I couldn't do an Infix Extension Method earlier...

nickjbenson (approved):
lgtm!

PR #1147, opened 2021-01-21
[FEAT] Experimental OpenXR hand data provider

PeterNancollis-UH:
Implements a LeapProvider sourcing data from the Microsoft Mixed Reality OpenXR Plugin Preview package for Unity 2020.

Enables basic interactions and hand visualisation with OpenXR hand tracking

PR #1166, opened 2021-06-23
[DRAFT] OVRProvider - Oculus Quest Hand Tracking

julienkay:
Inspired by #1147 this PR adds a similar alternative implementation of a LeapServiceProvider that allows to use Quest hand tracking with the Leap Motion Unity Modules.

Open questions:

- [ ] The OVRProvider still needs to be wrapped in a plaform define in some way, because of the dependency to the Oculus Integration.
- [ ] Some scripts (like InteractionHands) already expose a property setter, which allows to assign a custom LeapProvider implementation (e.g. OpenXRProvider, OVRProvider, Networked provider implementations etc.). However in other scripts (like HandUtils.cs) the Provider is a bit more hard-coded using FindObjectOfType. Maybe there's an opportunity here to move the provider reference to a single place (e.g. HandUtils), thus allowing to easily switch between implementations, maybe even at runtime.

### DIFF
--- a/Assets/Plugins/LeapMotion/Core/Scripts/Encoding/VectorHand.cs
+++ b/Assets/Plugins/LeapMotion/Core/Scripts/Encoding/VectorHand.cs
@@ -219,7 +219,7 @@ namespace Leap.Unity.Encoding {
     /// the camera-local hand rotation uses 4 bytes, and each joint position component is
     /// encoded in hand-local space using 3 bytes.
     /// </summary>
-    public int numBytesRequired { get { return 86; } }
+    public int numBytesRequired { get { return 92; } }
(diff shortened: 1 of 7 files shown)


MaxPalmer-UH:
Will aim to get this pulled into our backlog for review / action.

mrchantey:
This is so cool! I was able to get it up and running in less than 5 minutes, so excited to be using the Ultraleap Interaction Engine with a Quest.

MattGrayUL:
As this has been updated to a newer develop commit by @mrchantey, I am closing this PR in favor of [theirs](https://github.com/ultraleap/UnityPlugin/pull/1186)

PR #1228, opened 2022-07-08
Added formatting step to check copyright year

MattFiler:
Inspired by #1214, this MR adds a step into the formatting check job to automatically validate copyright headers to make sure they display the correct year.


MattGrayUL:
I have checked this and it works great, I suggest we wait until we are making the next release before we merge it as it may cause heavy conflicts on some branches :)

MattGrayUL (approved):

PR #1241, opened 2022-07-20
Fix physics hands arm rotation and make ignores apply on re-entry checks

rorygames:
- Corrects the rotation of the arms when using physics hands.
- Prevents ignored objects from being checked against when hands re-enter the frame.


rorygames:
I've changed the hands to be continuous speculative and it's made the hands way more stable during grabbing 😍

JuliaRonnebergerUL (approved):
tested in Unity 2020 -> works well and arm rotation looks good :)

PR #1247, opened 2022-07-22
Add XR validation warning for not targeting Android 29 when using OpenXR

rblenkinsopp:
## Summary

Adds an XR validation step to ensure an application is targeting Android 29, as per the OpenXR recommendations, if OpenXR is being used for a project.

This ensures that the tracking service can be correctly connected to despite Android 11's package visibility changes.

## Contributor Tasks

## Reviewer Tasks

- [x] Code reviewed.
- [x] Ensure that if an Android SDK version greater than 29 is listed, there is a warning in OpenXR validation.
- [x] Ensure that if the "Fix" button is clicked, it automatically adjusts the target, and the minimum SDK level if it was higher than 29.

## Closes JIRA Issue

Relates to OXR-190 (and is a warning to ensure the workaround continues to operate correctly).

### DIFF
--- a/Packages/Tracking OpenXR/CHANGELOG.md
+++ b/Packages/Tracking OpenXR/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added OpenXR validation check for ensuring applications target Android SDK 29 as per the [OpenXR recommendations] (https://registry.khronos.org/OpenXR/specs/1.0/loader.html#android-active-runtime-location).
+
 ### Changed
 
 ### Fixed
--- a/Packages/Tracking OpenXR/Runtime/Scripts/HandTrackingFeature.cs
+++ b/Packages/Tracking OpenXR/Runtime/Scripts/HandTrackingFeature.cs
@@ -144,10 +144,34 @@ namespace Ultraleap.Tracking.OpenXR
 #if UNITY_EDITOR
         protected override void GetValidationChecks(List<ValidationRule> rules, BuildTargetGroup targetGroup)
         {
+            // If building for Android, check that we are targeting Android API 29 for maximum compatibility.
+            rules.Add(new ValidationRule(this)
+            {
+                message = "Android target SDK version is not set to 29",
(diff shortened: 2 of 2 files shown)


MattGrayUL (approved):
Approve the code and performed some dev testing, setting target and min version above and below 29, works as expected by warning me when I am targeting 30 and sets the min and target as applicable

May require QA eyes on it prior to merging though? :)

Greg-UL:
@rblenkinsopp what happens if you change your project to use another plugin/runtime does the validation go away?

MattGrayUL:
> @rblenkinsopp what happens if you change your project to use another plugin/runtime does the validation go away?

So, if I had Android as a build target, set the target android version to 30, and then swap my build target to standalone.... I am still warned about the android version. I can press fix, and it goes away (and doesn't change my build target)

So essentially, it may warn you about your target android version even when not using android build settings, but it will still allow you to build etc while the warning is there, so is not an issue. and the wording is clear that is it android specific

image

MaxPalmer-UH:
I can't test this, but the changes look good.

rblenkinsopp:
> > @rblenkinsopp what happens if you change your project to use another plugin/runtime does the validation go away?
> 
> @Greg-UL So, if I had Android as a build target, set the target android version to 30, and then swap my build target to standalone.... I am still warned about the android version. I can press fix, and it goes away (and doesn't change my build target)
> 
> So essentially, it may warn you about your target android version even when not using android build settings, but it will still allow you to build etc while the warning is there, so is not an issue. and the wording is clear that is it android specific
> 
> image

Hmm, I should actually be able to make it so it only lists that as a warning if you have Android as a build target. We should probably do that.

MaxPalmer-UH:
> > > @rblenkinsopp what happens if you change your project to use another plugin/runtime does the validation go away?
> > 
> > 
> > @Greg-UL So, if I had Android as a build target, set the target android version to 30, and then swap my build target to standalone.... I am still warned about the android version. I can press fix, and it goes away (and doesn't change my build target)
> > So essentially, it may warn you about your target android version even when not using android build settings, but it will still allow you to build etc while the warning is there, so is not an issue. and the wording is clear that is it android specific
> > image
> 
> Hmm, I should actually be able to make it so it only lists that as a warning if you have Android as a build target. We should probably do that.

Yes, that would be a good change

PR #1264, opened 2022-08-04
Hand Rays visuals refactored to reduce logic duplication in example assets

rorygames:
## Summary

The principle here is so that the hand rays can be interchanged within a scene, while also passing through the logic that they do in the visuals back to the code. Means you don't need to reimplement the logic of the rays to get the raycasthit information.

- Refactors the hand ray visuals to use the same class (means you can interchange them)
- Adds an event when their raycasts return true
- Makes the rays ignore IE and Physics Hands by default
- User can pick layers that they want the ray to ignore.

## Contributor Tasks

## Reviewer Tasks

- [x] Code reviewed.

### DIFF
--- a/Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayParabolicLineRenderer.cs
+++ b/Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayParabolicLineRenderer.cs
@@ -6,8 +6,6 @@
  * between Ultraleap and you, your company or other organization.             *
  ******************************************************************************/
 
-using Leap.Unity.Interaction;
-using Leap.Unity.Interaction.PhysicsHands;
 using Leap.Unity.Preview.HandRays;
 using System.Collections.Generic;
 using UnityEngine;
--- a/Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayParabolicLineRenderer.cs
+++ b/Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayParabolicLineRenderer.cs
@@ -6,13 +6,15 @@
  * between Ultraleap and you, your company or other organization.             *
  ******************************************************************************/
 
+using Leap.Unity.Interaction;
+using Leap.Unity.Interaction.PhysicsHands;
 using Leap.Unity.Preview.HandRays;
 using System.Collections.Generic;
 using UnityEngine;
 
 namespace Leap.Unity
 {
-    public class HandRayParabolicLineRenderer : MonoBehaviour
+    public class HandRayParabolicLineRenderer : HandRayRenderer
     {
         [Tooltip("The scale of the projection of any hand distance from the approximated "
                + "shoulder beyond the handMergeDistance.")]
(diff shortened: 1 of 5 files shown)


Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayStraightLineRenderer.cs
```diff
@@ -5,124 +5,30 @@
  * http://www.apache.org/licenses/LICENSE-2.0, or another agreement           *
  * between Ultraleap and you, your company or other organization.             *
  ******************************************************************************/
+using Leap.Unity.Interaction;
+using Leap.Unity.Interaction.PhysicsHands;
```
MattGrayUL, line 9: these aren't necessary?

Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayParabolicLineRenderer.cs
```diff
@@ -6,13 +6,15 @@
  * between Ultraleap and you, your company or other organization.             *
  ******************************************************************************/
 
+using Leap.Unity.Interaction;
+using Leap.Unity.Interaction.PhysicsHands;
```
MattGrayUL, line 10: These aren't necessary?

Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayRenderer.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            _layerMask ^= interactionManager.contactBoneLayer.layerMask;
+        }
+
+        PhysicsProvider physicsProvider = FindObjectOfType<PhysicsProvider>();
+        if (physicsProvider != null)
+        {
+            _layerMask ^= physicsProvider.HandsLayer.layerMask;
+            _layerMask ^= physicsProvider.HandsResetLayer.layerMask;
+        }
```
MattGrayUL, line 45: It feels like Handrays being dependent on Physics hands is the wrong way around?

Can we make physics hands _request_ its layers be added instead?

Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayRenderer.cs
```diff
@@ -0,0 +1,162 @@
...
+    public UnityEvent<RaycastHit> OnRayUpdate;
+
+    private void Start()
+    {
+        _layerMask = -1;
+        InteractionManager interactionManager = FindObjectOfType<InteractionManager>();
+        if (interactionManager != null)
+        {
+            _layerMask ^= interactionManager.contactBoneLayer.layerMask;
+        }
```
MattGrayUL, line 38: It feels like Handrays being dependent on IE is the wrong way around?

Can we make IE request its layers be added instead?

Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayRenderer.cs
MattGrayUL: Actually, this would cause a dependency if HandRays moved to stable

Packages/Tracking Preview/Examples~/HandRays/Scripts/HandRayRenderer.cs
MattGrayUL: Actually, this would cause tracking to have a dependency on preview

MattGrayUL (approved):
Tested the example scene and works as expected

PR #1271, opened 2022-08-15
Unity-856 - AttachmentHands stuck in loop when resetting the script

JuliaRonnebergerUL:
## Summary

fix attachment hand delete loop when resetting the script and an editor error when there is no leap provider in the scene

## Contributor Tasks

- [x] Check any relevant CHANGELOG files have been updated.

## Reviewer Tasks

- [x] Code reviewed.

## Closes JIRA Issue

Closes UNITY-856


JuliaRonnebergerUL:
@Greg-UL tests I ran:
- Attachment Hands example scene works as expected
- single attachment points can be deleted: on the AttachmentHands component, untick any one of the attachment points
  - result: there is a pop up warning message and confirming the deletion leads to the toggle being unticked in the inspector and the corresponding gameObject disappearing from the scene
- script can be reset: in the attachment hands example scene, reset the AttachmentHands script (right click on AttachmentHands component -> Reset)
  - result 1: all attachment hand axes gameobjects disappear from the scene.
  - result 2: on the AttachmentHands component: only palm and wrist are ticked and the leapProvider is set correctly
  - result 3: in the hierarchy under 'attachment Hand (Left)', there are only two gameObjects (Palm and Wrist)
- script works without using the prefab: in a new scene, add a new empty gameObject and add the AttachmentHands.cs as a component
  - result 1: only palm and wrist are ticked
  - result 2: in the hierarchy under 'attachment Hand (Left)', there are only two gameObjects (Palm and Wrist)
  - result 3: no errors in the console
  - tick another attachment point on the AttachmentHands component - this appears as a gameobject in the hierarchy under 'Attachment Hand (Left)'

Greg-UL:
@JuliaRonnebergerUL that looks good. I found that deleting one of the middle points of a finger would then delete the tip of the finger on one hand and not the other - I don't suppose you could check that.

For tests that use the script without the prefab does that mean that we only bring up a tracked palm and wrist when no other joints are enabled on the attachment hands? Could I add this to say capsule hands and then enable different joints using the attachment hands script?

JuliaRonnebergerUL:
>  I found that deleting one of the middle points of a finger would then delete the tip of the finger on one hand and not the other - I don't suppose you could check that.

I've never seen that happening before and I can also not reproduce it now. If you see that again, maybe we should create a separate ticket for it?

> For tests that use the script without the prefab does that mean that we only bring up a tracked palm and wrist when no other joints are enabled on the attachment hands? Could I add this to say capsule hands and then enable different joints using the attachment hands script?

No, the attachment hands just follow all the tracking data from the service provider (same as the capsule hands). When you just add the script (/or reset the script), it uses its default which only toggles the attachment points for the palm and for the wrist on and all other attachment points off. So that means that by default you only have those two gameobjects in the hierarchy where you can then attach other objects to. But you can always still toggle the other attachment points on and off.

You could use an attachment hand to kind of build a capsule hand though, eg. by adding small spheres to all the attachment points and then you could toggle those on and off to create a sphere hand that uses only some of the joints.

Greg-UL:
I'll raise that if I see it again.

I'm happy with the testing on this thanks.

JuliaRonnebergerUL:
> What happens on the removal of the attachment script after multiple attachment points have been created and assigned game objects?

I just tested this and the created attachment points and the assigned gameObjects stay where they were when deleting the script. I am not sure if this is the expected behavior here or if we would expect them all to be deleted with the script?

Greg-UL:
I'm not sure what the desired functionality is here either. I feel like it should remove the items if their use is only relevant to the scrip, but that's just my opinion.

MattGrayUL (approved):

MattGrayUL:
@Greg-UL  I think it makes more sense to never delete the objects, in a normal scenario, the user will have created these objects and then attached them... the only reason I can think you'd want to delete them is if you were using our example prefab as a base and didn't want the X/Y/Z meshes that we have by default

MattGrayUL (approved):

PR #1279, opened 2022-08-22
Rework dependencies-libtrack into an advisory check

MattFiler:
This reworks the legacy `dependencies-libtrack` job to be a useful `check-plugins` job in the test stage.

This new advisory job will fail if the version of libtrack in the UnityPlugin repo does not match the version of libtrack on its stable branch.

The job completes with artifacts which contain the libs for Android and Windows from stable, formatted for the Core plugins folder directory structure.


MattGrayUL:
The outputs of this step are not the most recent versions of each, in one instance, this meant Android builds using the client in the produced artefacts did not run at all. We may need to either change the origin of these artefacts, or use this step as a reminder to get the correct release clients (i.e. from confluence links) rather than a 'gather from stable'

PR #1285, opened 2022-08-23
Add static AndroidServiceBinder that manages itself on application quit

MattGrayUL:
## Summary

Converts the android bind system to be static and only run when required and only quits when application quits. Avoids breaking changes.

## Contributor Tasks

## Reviewer Tasks

- [ ] Code reviewed.

## Closes JIRA Issue

Closes UNITY-864

### DIFF
--- a/Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
+++ b/Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
@@ -1,0 +1,94 @@
+namespace Leap.Unity
+{
+    using System;
+    using UnityEngine;
+
+    public static class AndroidServiceBinder
+    {
+#if UNITY_ANDROID
+
+        private static bool isBound = false;
+        public static bool IsBound
+        { 
+            get { return isBound; }
+            private set { isBound = value; }
+        }
+        static AndroidJavaObject _serviceBinder;
+        static AndroidJavaClass unityPlayer;
+        static AndroidJavaObject activity;
+        static AndroidJavaObject context;
+        static ServiceCallbacks serviceCallbacks;
+
+        public static bool Bind()
+        {
+            bool isBound = _serviceBinder?.Call<bool>("isBound") ?? false;
+
+            if (!isBound)
+            {
+                isBound = TryBind();
+
+                Application.quitting -= OnApplicationQuitting;
+                Application.quitting += OnApplicationQuitting;
+            }
+
+            return isBound;
+        }
+
+        private static void OnApplicationQuitting()
+        {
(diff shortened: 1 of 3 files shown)


Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
```diff
@@ -0,0 +1,94 @@
+namespace Leap.Unity
+{
+    using System;
+    using UnityEngine;
+
+    public static class AndroidServiceBinder
+    {
+#if UNITY_ANDROID
```
rblenkinsopp, line 8: Why not `#if` the whole class (or indeed at this to an Android only asmdef?) Might be a cleaner solution?

Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private static bool isBound = false;
+        public static bool IsBound
```
rblenkinsopp, line 11: Is there a reason this is isn't a public auto property (as the implementation is a trivial getter/setter?)

Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            return success;
+        }
+
+        private static void Unbind()
+        {
+            if (_serviceBinder != null)
+            {
+                Debug.Log("ServiceBinder.unbind...");
```
rblenkinsopp, line 87: This seems like this log message could match the above `Debug.Logs` a little better?

Packages/Tracking/Core/Runtime/Scripts/Utils/AndroidServiceBinder.cs
```diff
@@ -0,0 +1,94 @@
...
+        static AndroidJavaObject context;
+        static ServiceCallbacks serviceCallbacks;
+
+        public static bool Bind()
+        {
+            bool isBound = _serviceBinder?.Call<bool>("isBound") ?? false;
+
+            if (!isBound)
+            {
+                isBound = TryBind();
```
rblenkinsopp, line 28: This doesn't look like it handles the failure of `TryBind()`. In this situation, it would set the Application.quitting event even though binding was unsuccessful. this would then call `Unbind()` on application termination even though we never successfully bound.

rblenkinsopp:
Comments are mostly for consideration rather than anything that I'd block on.

PR #1317, opened 2022-10-14
[FIX] Scaling XR Providers

zalo:
## Summary

Fixes a bug introduced around https://github.com/ultraleap/UnityPlugin/commit/2e48d9ba1e1d460161cfd45d294379bab7e553ee which breaks `LeapXRPinchLocomotion`.

## Reviewer Tasks

- [x] Code reviewed.

## Closes Issue

Fixes [this bug](https://ultrahaptics.slack.com/archives/C029XFH4085/p1665496663735789?thread_ts=1655370797.300899&cid=C029XFH4085) reported by @Dwarph.


MaxPalmer-UH:
Have pulled this change - will do some testing on it.

MaxPalmer-UH:
@zalo so those changes fix the ability to use the Pinch Locomotion script without weird scaling stuff becoming unstable. What I did notice though is when you pinch with both hands to scale the scene, the capsule hands also scale. Was that intentional or has something else got lost/broken? Also what's the script ordering meant to fix? 

https://user-images.githubusercontent.com/53301524/196404155-d48661d7-78ac-483b-b24d-231eb7aeb1e1.mp4

zalo:
@MaxPalmer-UH On the contrary, the capsule hands are _not_ scaling with the rest of the provider+rig.

Are your capsule hands in the same transform hierarchy as both your Service Provider and main camera rig?   My rig is: Rig Parent -> XR Provider -> Left/Right Capsule Hand.

MaxPalmer-UH:
OK, making the capsule hand a child of the rig stops the scaling. A rigged hand will still scale, I presume this is expected?

MattGrayUL (approved):

PR #1323, opened 2022-11-01
Add XRLeapProviderManager

MattGrayUL:
## Summary

Adds a XRLeapProviderManager which is intended to be used in XR scenarios where the source of Hand Tracking data is unknown. This will first target OpenXR unless it is not available or not being used, and then fall back to LeapC

## Contributor Tasks

## Reviewer Tasks

- [x] Code reviewed.

## Closes JIRA Issue

Closes UNITY-927


Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
```diff
@@ -0,0 +1,58 @@
...
+using Leap.Unity;
+using System;
+using System.Collections;
+using UnityEngine;
+using UnityEngine.XR.Management;
+using UnityEngine.XR.OpenXR;
+using Ultraleap.Tracking.OpenXR;
+
+
+public class XRLeapProviderManager : PostProcessProvider
```
MaxPalmer-UH, line 11: Can we comment the public interface please?

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
```diff
@@ -0,0 +1,58 @@
...
+            return (_leapProvider == null) ? leapXRServiceProvider : _leapProvider;
+        }
+    }
+
+    [Tooltip("Forces the use of the non-OpenXR provider, using LeapC hand tracking")]
+    public bool forceLeapService;
+
+    private IEnumerator Start()
+    {
+        while (XRGeneralSettings.Instance == null) yield return new WaitForEndOfFrame();
```
MaxPalmer-UH, line 32: Will this definitely be ready by the end of the frame?

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        if (XRGeneralSettings.Instance.Manager.activeLoader.name == "Open XR Loader" &&
+            OpenXRSettings.Instance.GetFeature<HandTrackingFeature>() != null &&
+            OpenXRSettings.Instance.GetFeature<HandTrackingFeature>().enabled &&
```
MaxPalmer-UH, line 36: I can't remember what enabled means - there was some weird discussion in OpenXR ages ago about the concept of active/enabled etc? I guess this is about the layer being enabled?

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
```diff
@@ -0,0 +1,58 @@
...
+            Destroy(openXRLeapProvider.gameObject);
+        }
+
+
+        inputLeapProvider = LeapProvider;
+        OnProviderSet?.Invoke(inputLeapProvider);
+    }
+
+    public override void ProcessFrame(ref Frame inputFrame)
+    {
```
MaxPalmer-UH, line 56: Does this do anything?

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            Debug.Log("Using OpenXR for Hand Tracking");
+            _leapProvider = openXRLeapProvider;
+            Destroy(leapXRServiceProvider.gameObject);
+        }
+        else
+        {
+            Debug.Log("Using LeapService for Hand Tracking");
+            _leapProvider = leapXRServiceProvider;
+            Destroy(openXRLeapProvider.gameObject);
```
MaxPalmer-UH, line 47: Will destroying the provider that is not needed sort the issue of other components that look for the provider or could they do this while there are two? Would we expect a client to access the post process provider to use this?

Packages/Tracking/OpenXR/Runtime/Ultraleap.Tracking.OpenXR.asmdef
```diff
@@ -6,7 +6,8 @@
         "GUID:96aa6ba065960476598f8f643e7252b6",
         "GUID:cec9848704c49724192ff47f9b3e9156",
         "GUID:2ee0542102ac1084e8d9a910885c1e81",
-        "GUID:ba171b3dd2a51234ab864770f99741a5"
+        "GUID:ba171b3dd2a51234ab864770f99741a5",
+        "GUID:e40ba710768534012815d3193fa296cb"
```
MaxPalmer-UH, line 10: What's this new reference?

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
MattGrayUL: This just continues the data passing through, as it's an abstract base, it's a necessary evil

Packages/Tracking/OpenXR/Runtime/Ultraleap.Tracking.OpenXR.asmdef
MattGrayUL: This is Unity's XR Management package (should really be necessary for XR work anyway)

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
MattGrayUL: They _could_ still get access and then it be deleted, as we need to wait for OpenXR to spin up, we can't push this class back in the order of execution to avoid this. We should advise devs to only reference the XRLeapProviderManager directly if they are using it to avoid finding invalid/unused Providers

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
MattGrayUL: enabled is something we are using and editing in the Feature, it's one of the blocking reasons that this PR is currently Draft as we need to confirm we can safely use '.enabled' as we hope to

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
MaxPalmer-UH: Thanks

Packages/Tracking/OpenXR/Runtime/Scripts/XRLeapProviderManager.cs
MaxPalmer-UH: Ah yes, we'll need documentation for this. Good thing to put in that.

MaxPalmer-UH (approved):
Tested with some issues fixed - only on Windows. Needs wider testing as part of the release. Matt has done Android testing too.

MattGrayUL:
I tested on Quest2:

- Oculus runtime (all OpenXR toggle combinations)
- SteamVR runtime (all OpenXR toggle options)

Points of note:
- SteamVR and Oculus runtimes act differently in regards to hand tracking data via OpenXR (Oculus falls back to LeapC better than SteamVR, SteamVR likes to provide controllers as OpenXR hand tracking more)
- In Editor, Oculus says there is OpenXR hand tracking when all options for it are disabled

The above notes are the reason we also provide a `forceLeapService` option for the Manager, so users can force it when required (particularly during development)

PR #1330, opened 2022-11-03
Improving physics hand usability and performance

rorygames:
## Summary

- Makes grasp helpers always calculate states, even when ignored.
- Adds functions to get state information of a rigidbody.
- Moves the frame updating on the same tick as fixedupdate to reduce latency.
- Modifies hovers to use a dynamic capsule between the most extended finger and the palm, instead of just a sphere.
- Fixes tip position of the physics hand to be the exact tip of the bones.
- Modifies the overlaps for contact to be a bit larger, and moves the tip overlap forward to provide better pointing contact.
- Renaming idle grasp helper state to hover.

## Contributor Tasks

## Reviewer Tasks

Please just double check the sample scene still works properly 😄 

- [x] Code reviewed.

## Closes JIRA Issue

Closes UNITY-939


Dwarph:
seems good to me - had a small play with the example scene, and the tips matched what I saw in the viz

MattGrayUL:
Code looks good to me, I tested it in editor on a Quest2 and interactions with different objects worked well so no breaks :)

MattGrayUL (approved):

PR #1342, opened 2022-11-16
Don't check for ignore on destroyed rigidbody

MattFiler:
Mini MR to add a null check on the object ignorer to stop looking at destroyed rigidbodies.

This error was killing phys hands in Aurora 👻


Packages/Tracking Preview/PhysicsHands/Runtime/PhysicsHand.cs
```diff
@@ -528,6 +528,9 @@ private void HandleIgnoredObjects()
                         _ignoredData[i].timeout -= Time.fixedDeltaTime;
                     }
 
+                    if (_ignoredData[i].rigid == null)
+                        continue;
```
MattGrayUL, line 532: for consistency, can you whack some braces on this plz? :)

PR #1348, opened 2022-11-24
Embedded OpenXR Layer Support

rblenkinsopp:
## Summary

This MR:
- Adds a new OpenXR Feature which installed our API layer as an embedded layer
- Dynamically removes LeapC and The Android Service Binder from these builds, so they don't conflict
- Adds a new feature group to group them together.
- Adds a warning if the OpenXR version in Unity is less that `1.0.25` as the best proxy for the loader version in use which can't be checked directly.

It is currently a draft as it doesn't include the final version of the API layer that will ship with this option, but is ready for review.

## Contributor Tasks

## Reviewer Tasks

- [ ] Code reviewed.

## Closes JIRA Issue


Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
```diff
@@ -0,0 +1,116 @@
...
+            // Attempt to check the version of OpenXR for at least 1.0.25 as a proxy for what loader version is
+            // included.
+            rules.Add(new ValidationRule(this)
+            {
+                message = "The OpenXR loader must be at least version 1.0.25 to support embedding the Ultraleap API layer",
+                error = false,
+                checkPredicate = () =>
+                {
+                    var versionParts = Array.ConvertAll(OpenXRRuntime.apiVersion.Split("."), int.Parse);
+                    return versionParts[0] >= 1 && versionParts[1] >= 0 && versionParts[2] >= 25;
```
MattGrayUL, line 47: I think this version checking might be insufficient, if I had 2.5.2 it would fail because versionParts[2] would be too small?

it would need something like this:

```
if (versionParts[0] > 1)
{
    return true;
}
else if (versionParts[0] == 1)
{
    if(versionParts[1] > 0)
    {
        return true;
    }
    else if (versionParts[1] == 0 && versionParts[2] >= 25)
    {
        return true;
    }
}
return false;
```

Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
```diff
@@ -0,0 +1,116 @@
...
+#if UNITY_EDITOR
+    [OpenXRFeature(FeatureId = FeatureId,
+        Version = "1.0.0",
+        UiName = "Ultraleap Hand Tracking API Layer",
+        Company = "Ultraleap",
+        Desc = "Embeddable API layer for Ultraleap Hand Tracking",
+        Category = FeatureCategory.Feature,
+        Required = false,
+        OpenxrExtensionStrings = "",
+        BuildTargetGroups = new[] { BuildTargetGroup.Android }
```
MattGrayUL, line 29: I wonder if we can name this feature (and possibly the existing one) more descriptively?

currently it's hard to know what each do via the UI alone
image

Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            // Attempt to check the version of OpenXR for at least 1.0.25 as a proxy for what loader version is
+            // included.
+            rules.Add(new ValidationRule(this)
+            {
+                message = "The OpenXR loader must be at least version 1.0.25 to support embedding the Ultraleap API layer",
+                error = false,
+                checkPredicate = () =>
+                {
+                    var versionParts = Array.ConvertAll(OpenXRRuntime.apiVersion.Split("."), int.Parse);
```
MattGrayUL, line 46: I had an error here where the `.Split(".")` is looking for a char, not a string

Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
rblenkinsopp: I'm happy to entertain suggestions. Most other vendors just have "Hand Tracking" as the feature name. I added the vendor to make it clearer but that might not be needed now it's part of a feature set.

Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
rblenkinsopp: Easy change but interesting that I didn't see an issue with that 🤔.

Packages/Tracking/OpenXR/API Layer/HandTrackingApiLayerFeature.cs
rblenkinsopp: Yeah you're right, I'll fix that. I wish unity had a version string comparison function...

MattGrayUL: We could also convert to System.Version and then do a [.CompareTo](https://learn.microsoft.com/en-us/dotnet/api/system.version.compareto?view=net-7.0)?

rblenkinsopp: I forgot about `System.Version` I've written a much more succinct version using that.

rblenkinsopp:
I've resolved the non-naming-based issues with this.

rblenkinsopp:
We agreed to hold off on merging this till Unity has a default loader which supports embedded layers (>= 1.0.25), this will remain in draft until then.

rblenkinsopp:
We are closing this for now as we are not looking to support this for now.

rblenkinsopp:
Closing as this is now a separate package with no direct dependency on this package.

PR #1359, opened 2022-12-12
Preview/locomotion jump gems

MattGrayUL:
## Summary

Adds Jump Gems and Pinch Teleport examples and supporting code into Preview

Includes 3 scenes:
- 1. Teleportation Anchors : Both locomotion methods, user can only teleport to Anchors
- 2. Pinch Teleportation : Free teleportation using Pinch
- 3. Jump Gems Teleportation : Free teleportation using Jump Gems

## Contributor Tasks

## Reviewer Tasks

- [x] Code reviewed.

## Closes JIRA Issue


Packages/Tracking Preview/Examples~/Locomotion/Teleportation/Scripts/TeleportExampleManager.cs
```diff
@@ -0,0 +1,73 @@
...
+ *                                                                            *
+ * Use subject to the terms of the Apache License 2.0 available at            *
+ * http://www.apache.org/licenses/LICENSE-2.0, or another agreement           *
+ * between Ultraleap and you, your company or other organization.             *
+ ******************************************************************************/
+
+using Leap.Unity.Preview.Locomotion;
+using UnityEngine;
+
+namespace Leap.Unity.Examples.Preview
```
MaxPalmer-UH, line 12: Definitely some future refactoring improvements here when moving to stable.

Packages/Tracking Preview/HandRays/Runtime/Scripts/FarFieldLayerManager.cs
```diff
@@ -0,0 +1,75 @@
...
+        protected void GenerateLayers()
+        {
+            if (_layersGenerated || !AutomaticFarFieldObjectLayer)
+            {
+                return;
+            }
+
+            FarFieldObjectLayer = -1;
+
+            for (int i = 8; i < 32; i++)
```
MaxPalmer-UH, line 43: Presumably we are doing 8 here to skip the first known (always used layers)

Packages/Tracking Preview/HandRays/Runtime/Scripts/RayInteractors/HandRayInteractor.cs
```diff
@@ -0,0 +1,108 @@
...
+        public Action<RaycastHit[], RaycastHit> OnRaycastUpdate;
+
+        /// <summary>
+        /// The points returned by the raycast, which can be used by a HandRayRenderer
+        /// </summary>
+        [HideInInspector] public Vector3[] linePoints;
+        /// <summary>
+        /// The number of points returned by the raycast
+        /// </summary>
+        [HideInInspector] public int numPoints;
```
MaxPalmer-UH, line 44: Seems odd to do this as this info is on the linePoints array.

Packages/Tracking Preview/Locomotion/Runtime/Scripts/Teleportation/Teleport Action/Jump Gem Teleport/JumpGemTeleport.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        public JumpGem PinchedGem => _pinchedGem;
+        private JumpGem _pinchedGem;
+
+        [Header("Jump Gem Teleport - Interaction Setup")]
+        [SerializeField, Range(0.0001f, 1f), Tooltip("The distance the user has to pinch a gem before the ray will be activated.")]
+        private float _minimumDistanceToActivate = 0.1f;
+
+        [SerializeField, Range(0f, 1f), Tooltip("The point at which the ray should interpolate when releasing. " +
+            "E.g. 0.8 will activate when the user is 80% released.")]
```
MaxPalmer-UH, line 33: Needs more info to be clear, this is a percentage of what?

Packages/Tracking Preview/HandRays/Runtime/Scripts/FarFieldLayerManager.cs
MattGrayUL: Yeah it skips the first known Unity-set layers

Packages/Tracking Preview/Locomotion/Runtime/Scripts/Teleportation/Teleport Action/Jump Gem Teleport/JumpGemTeleport.cs
```diff
@@ -0,0 +1,160 @@
...
+            ChangeVisibleJewels();
+            ProcessPinchedJewel();
+        }
+
+        /// <summary>
+        /// Hides all non-active gems.
+        /// </summary>
+        private void ChangeVisibleJewels()
+        {
+            bool anyPinched = _jumpGems.Any(x => x.IsPinched);
```
MaxPalmer-UH, line 72: Might be worth checking if Any allocates as it's being called per frame.

Packages/Tracking Preview/HandRays/Runtime/Scripts/RayInteractors/HandRayInteractor.cs
MattGrayUL: From what I understand, this is how it is used within XRI too

The line renderers reference the interactor to know how to draw their lines, regardless of any input 'rays' as the Interactor is the central source of information

MaxPalmer-UH:
I guess the pinch to teleport doesn't have a 'both' for chirality

MaxPalmer-UH:
It's also a bit weird I can teleport onto the side of an object, but that's minor.

MattGrayUL:
> In general code looks OK for preview. Examples have some issues. 1.The Teleportation anchors level
> 
> > Pinch to teleport assumes you are right handed, no left handed ray
> > Jump gem is placed on the inner side of my left arm (good), but on the outer side of my right arm, which feels wrong in terms of picking up the gem with my left hand.
> > It's a bit weird I can see the text on the sign at the teleport anchor destination from the opposite side (but can't read it because it's mirrored).

I we may be able to add both rays? also, the gem being in the wrong place is my bad, I have the fix for it locally now :)

I can look to block the text, though it's not trivial due to the way the UI renders

MaxPalmer-UH:
Jump gem is also on the wrong side of the right hand in the 3. example.

MattGrayUL:
> Jump gem is also on the wrong side of the right hand in the 3. example.

Yeah I defaulted both prefabs while I was changing them previously, the fix is in the prefab :)

Packages/Tracking Preview/Locomotion/Runtime/Scripts/Teleportation/Teleport Action/Jump Gem Teleport/JumpGemTeleport.cs
MattGrayUL: I think this is fine as the list is only ever 2 in length (unless the dev has more than one set of Jump Gems in the scene)

MaxPalmer-UH (approved):
I'm happy with the examples now, the code is good enough for preview in it's current state. Will need to do some more work when it moves to Core as noted by Matt G.

MattGrayUL:
@Greg-UL this PR has had testing in these areas:

- Quest 2 + oculus runtime - Editor + build
- Index + SteamVR runtime - Editor
- OpenXR and LeapC tracking sources
- Each example scene run through
- Scene 1: teleport from one anchor to the other repeatedly
- Scene 1: jump gems from either hand
- Scene 1: cannot teleport elsewhere than Anchors
- Scene 1: pinch teleport is right hand only
- Scene 2: pinch teleport is right hand only
- Scene 2: can hold the pinch and rotate the anchor to change teleported rotation
- Scene 2: can only teleport to any surface (which includes the sides of objects in the example) cannot teleport to mid-air
- Scene 3: jump gems from either hand
- Scene 3: can only teleport to any surface (which includes the sides of objects in the example) cannot teleport to mid-air

Greg-UL:
@MattGrayUL that looks great and thanks for detailing the testing clearly. 
From a functional aspect the only other things I tried that aren't listed above were some scenarios around occlusion and the ray. 
Not that it's a big deal but it might be a nice touch to have the beam turn red for an interaction that isn't possible(mid air, or tracking lost mid pinch) and I thought it might be nice when the hand tracking is occluded to have the landing/destination circle turn red.
There was one issue I was getting with the pinch teleportation is that if you pinch and aim in the scene and then bring your arm out of your FoV then it will perform a teleport even though I haven't released the pinch.

MattGrayUL:
As the refactor coming up is quite large and relates to those suggested areas, I think we will merge in the current state and pick up those types of improvements with the new structure :)

PR #1391, opened 2023-01-31
Grab ball dynamically respects interaction zone

MattFiler:
Added an option in inspector to have the grab ball always respect the interaction zone regardless of currently being grabbed. Solves issues when moving away from the UI panel without an additional recenter option.

Also removed the orientation in SimpleGrabBallFeedback as it was not visible with the material used (and against our recommendations), and also added a grasp colour to SimpleInteractionGlow as this is more similar to how we have used grab balls in the past.

### DIFF
--- a/Packages/Tracking/Interaction Engine/Runtime/Scripts/UI/GrabBall.cs
+++ b/Packages/Tracking/Interaction Engine/Runtime/Scripts/UI/GrabBall.cs
@@ -52,6 +52,8 @@ namespace Leap.Unity.Interaction
 
         [Tooltip("Enable to restrict the distance the grab ball can move from the head. Useful to keep the attached object close enough to interact with")]
         public bool restrictGrabBallDistanceFromHead = true;
+        [Tooltip("Enabling this will dynamically shift the grab ball into the interaction area as the user moves through space")]
+        public bool dynamicallyLimitHeadDistance = true;
         [Tooltip("The furthest away from your head horizontally that the grab ball can move")]
         public float maxHorizontalDistanceFromHead = 0.5f;
         [Tooltip("The closest to your head horizontally that the grab ball can move")]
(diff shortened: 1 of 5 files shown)


Packages/Tracking/Interaction Engine/Runtime/Scripts/UI/GrabBall.cs
```diff
@@ -234,11 +236,11 @@ private void UpdateAttachedObjectTargetPose()
             }
 
             _transformHelper.position = grabBallPose.position;
-            grabBallPose.rotation = LookAtRotationParallelToHorizon(grabBallPose.position, _head.position);
+            if (updateRotation) grabBallPose.rotation = LookAtRotationParallelToHorizon(grabBallPose.position, _head.position);
```
MattGrayUL, line 239: Can this have braces added on new lines please?

Packages/Tracking/Interaction Engine/Runtime/Scripts/UI/GrabBall.cs
```diff
@@ -234,11 +236,11 @@ private void UpdateAttachedObjectTargetPose()
...
             _transformHelper.rotation = grabBallPose.rotation;
 
             _attachedObjectTargetPose.position = _transformHelper.TransformPoint(_attachedObjectOffset);
-            _attachedObjectTargetPose.rotation = Quaternion.Euler(xRotation, grabBallPose.rotation.eulerAngles.y, grabBallPose.rotation.eulerAngles.z);
+            if (updateRotation) _attachedObjectTargetPose.rotation = Quaternion.Euler(xRotation, grabBallPose.rotation.eulerAngles.y, grabBallPose.rotation.eulerAngles.z);
```
MattGrayUL, line 243: Can this have braces added on new lines please?

Dwarph:
big restriction behaviour regression 

- ghost ball no longer works & lerps strangely back to the grab ball
- grab ball doesn't smoothly follow the restricted ghost ball & jumps about the place

MattGrayUL:
@Dwarph is that specific to the content of this PR? As in does it _not_ occur on Develop?

PR #1395, opened 2023-02-05
Enable mirrored hand tracking support

PikaChokeMe:
## Summary

_Adds an option to make right hands track for left hands, and left hands track for right hands._

I've seen a couple people request this, and I think the Unreal plugin already supports this, but I wanted to enable the end user to be able to make their right hand work as their left hand and their left hand work as their right hand without having to rebind too much. 
Mirrored Tracking

I did this by adding a checkbox to "Mirror Hand Tracking"
image

And I did this at the provider level in the transformFrame method, modifying the frames read from the device directly, which allows other things that depend on the provider such as physics hands to "automagically" support it.
image

I'm not sure if there's a better way to do this currently, but if there is a better way, I can implement those changes, or if this works and is acceptable I think a lot of users would be happy to have support for this. I'm also pretty new to Unity and Unity plugin development in general so if there's any changes or recommendations that should be made, please let me know! 

## Contributor Tasks

## Reviewer Tasks

- [ ] Code reviewed.

## Closes JIRA Issue

### DIFF
--- a/Packages/Tracking/Core/Runtime/Scripts/LeapServiceProvider.cs
+++ b/Packages/Tracking/Core/Runtime/Scripts/LeapServiceProvider.cs
@@ -171,6 +171,10 @@ namespace Leap.Unity
         [SerializeField]
         protected float _physicsExtrapolationTime = 1.0f / 90.0f;
 
+        [Tooltip("Makes right hands track for left hands, and left hands track for right hands.")]
+        [SerializeField]
+        protected bool MirrorHandTracking = false;
+        
         #region Multiple Device
         public enum MultipleDeviceMode
         {
@@ -1042,7 +1046,19 @@ namespace Leap.Unity
 
         protected virtual void transformFrame(Frame source, Frame dest)
         {
-            dest.CopyFrom(source).Transform(new LeapTransform(transform));
+            var frameTransform = new LeapTransform(transform);
+
+            if (MirrorHandTracking)
+            {
+                foreach (Hand hand in source.Hands)
+                {
+                    hand.IsLeft = !hand.IsLeft;
+                }
+
+                frameTransform.MirrorX();
+            }
+
+            dest.CopyFrom(source).Transform(frameTransform);
         }
 
         private TrackingSource CheckLeapServiceAvailable()


MattGrayUL:
Hi, Thanks for the contribution! 🚀 

I was actually looking at ways to do similar this last week and totally forgot about the LeapTransform.MirrorX method, so thanks for spotting it and using it!

I hope to get around to testing it out properly tomorrow, but as a quick response:

As the LeapServiceProvider is specific to LeapC connections and does not cover OpenXR-based tracking, also it fairly substantially editing the way that the hand data is presented

As such, I think it would work better as a [PostProcessProvider](https://docs.ultraleap.com/unity-api/The-Examples/XR/hand-post-process.html#what-code-was-used-to-build-the-example). This way the raw data can be passed from either a LeapServiceProvider, or an OpenXRLeapProvider and then modified via a MirroredHandsPostProcessProvider and finally fed into Hands, or as you point out, a PhysicsProvider

It would also allow for more customization such as: only mirroring one hand, mirroring both (flipping), flipping but keeping position, flipping along a different axis(?) etc

If you would like to implement this, feel free, otherwise I will look to getting the same into the Plugin (possibly with an accompanying example scene 🙂)

PikaChokeMe:
> Hi, Thanks for the contribution! 🚀
> 
> I was actually looking at ways to do similar this last week and totally forgot about the LeapTransform.MirrorX method, so thanks for spotting it and using it!
> 
> I hope to get around to testing it out properly tomorrow, but here's a quick response:
> 
> The LeapServiceProvider is specific to direct LeapC connections and does not cover OpenXR-based tracking, also mirroring data is fairly substantially editing the way that the hand data is presented via the Provider
> 
> As such, I think it would work better as a [PostProcessProvider](https://docs.ultraleap.com/unity-api/The-Examples/XR/hand-post-process.html#what-code-was-used-to-build-the-example). This way the raw data can be passed from either a LeapServiceProvider, or an OpenXRLeapProvider and then modified via a MirroredHandsPostProcessProvider and finally fed into Hands, or as you point out, a PhysicsProvider
> 
> It would also allow for more customization such as: only mirroring one hand, mirroring both (flipping), mirroring but keeping position, mirroring along a different axis(?) etc
> 
> If you would like to implement this, feel free, otherwise I will look to getting the same into the Plugin (possibly with an accompanying example scene 🙂)

😮 ! 
Sorry for the late reply on this. I think from what I remember reading people were saying that PostProcessingProviders didn't work with some of the gesture scripts and controls, which is why I originally didn't pursue going this route. Is that just not true, or does that change when that post processed information is passed into something like a PhysicsProvider which then processes stuff down stream non the wiser that the data entered into it has already been post processed?

Also, I will say, I feel super bad, because I definitely made it sound like I was willing to do all the extra work necessary at the time, but I've kind of ended up a little busier than I expected, and I probably wont have too much extra time to iterate on this at the current moment. ( I also think you'll probably be able to put together a way better solution to this problem than I will! )

All that said, thank you so so so much for your time and all the extra information! Also if time allows at a later point, I hope to contribute to this project in the future! ( Hopefully next time with a proper solution! 😄 ) 

I'm also not sure if at this point I should close this issue, but please feel free to do whatever you have to do with it x)

MattGrayUL:
> 😮 ! Sorry for the late reply on this. I think from what I remember reading people were saying that PostProcessingProviders didn't work with some of the gesture scripts and controls, which is why I originally didn't pursue going this route. Is that just not true, or does that change when that post processed information is passed into something like a PhysicsProvider which then processes stuff down stream non the wiser that the data entered into it has already been post processed?

The Physics Provider _is_ a PostProcessProvider and they can quite happily stack, so long as it's handled properly, once the data is passed out of a PostProcessProvider, it is treated as if it came directly from a LeapProvider

> Also, I will say, I feel super bad, because I definitely made it sound like I was willing to do all the extra work necessary at the time, but I've kind of ended up a little busier than I expected, and I probably wont have too much extra time to iterate on this at the current moment. ( I also think you'll probably be able to put together a way better solution to this problem than I will! )

That's not a problem, we could end up with a bunch of back-and-fourth about it so i'm happy to push for us to implement this in the Plugin. "Better" is subjective! 😀 

> All that said, thank you so so so much for your time and all the extra information! Also if time allows at a later point, I hope to contribute to this project in the future! ( Hopefully next time with a proper solution! 😄 )

Contributions or feedback is always appreciated, particularly on content from the Preview Package like Physics Hands 😁 You can send any feedback to the [Discord server](https://discord.com/invite/3VCndThqxS)

> I'm also not sure if at this point I should close this issue, but please feel free to do whatever you have to do with it x)

I'll close it for now and mention any future developments here so you can see how it's going 🙂 Thanks for the contribution!

PR #1446, opened 2023-05-17
Bug/unity 1137 fixes for image alignment distortion issues

MaxPalmer-UH:
## Summary

This PR addresses issue UNITY-1137, where hands are misaligned with IR images in the control panel on certain devices. This is because the distortion matrix is treated differently between devices. The changes flip the Y value in the distortion matrix data, this also flips the IR image, so we invert the V value in the UV calculation in the shader code. 

This change makes things work in theory across device types (SIR170, 3Di, LMC). However, in fact this will then break LMC alignment. There is a new version of the service which changes the LMC distortion data to work with this fix.

This change backs out the left/right distortion matrix swap, which appeared to fix alignment across all devices but in fact was the wrong solution.

As this bug has taken a while to track down, I've added new comments to boost understanding of the 'rendering pipeline' for images.

Testing requirements:
Needs to be tested with a version of the control panel that uses a plugin with these changes. For IR/Hand alignment to work across all device types, requires a new version of the SDK/Service (5.12 RC4). Note, some devices will not obviously show misalignment so you need one that shows the issues before the fix. Because of the way SIR170s treat the image centre, they also don't show alignment issue.

## Contributor Tasks

- [x] Check any relevant CHANGELOG files have been updated.

## Reviewer Tasks

- [x] Code reviewed.

## Closes JIRA Issue

### DIFF
--- a/Packages/Tracking/CHANGELOG.md
+++ b/Packages/Tracking/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Preview package version dependency mismatch for XRI when using InputSystem 1.4.4
+- Control panel IR image / hand vertical alignment across device types and individual devices, through changing the handdling of the distortion matrix
 
 ### Known issues 
 - Scenes containing the infrared viewer render incorrectly on Android build targets and in scriptable render pipelines such as URP and HDRP. 
--- a/Packages/Tracking/CHANGELOG.md
+++ b/Packages/Tracking/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Preview package version dependency mismatch for XRI when using InputSystem 1.4.4
-- Control panel IR image / hand vertical alignment across device types and individual devices, through changing the handdling of the distortion matrix
+- (Passthrough) change the handling of the distortion matrix for vertical alignment across device types
 
 ### Known issues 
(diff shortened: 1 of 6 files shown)


Packages/Tracking/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Preview package version dependency mismatch for XRI when using InputSystem 1.4.4
+- Control panel IR image / hand vertical alignment across device types and individual devices, through changing the handdling of the distortion matrix
```
MattGrayUL, line 19: Do we need to mention the control panel here? the changelog only really relates to the Plugin, so we could just mention the passthrough distortion matrix, but not the control panel?

MattGrayUL (approved):
Happy with the changes, tested in editor on the Control Panel and the alignment was resolved

Just a couple of comments about changelog and commenting

Packages/Tracking/CHANGELOG.md
MaxPalmer-UH: Good suggestion, changed.

Packages/Tracking/Core/Runtime/Plugins/LeapCSharp/Image.cs
```diff
@@ -151,7 +151,7 @@ public float[] Distortion(CameraType camera)
             if (camera != CameraType.LEFT && camera != CameraType.RIGHT)
                 return null;
 
-            return imageData(camera).DistortionData.Data;
+            return imageData(camera).DistortionData.FlippedData;
```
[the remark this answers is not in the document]
MaxPalmer-UH: I've added a comment.

PR #1450, opened 2023-05-24
Fix finger Direction and hand GrabStrength for OpenXR

rblenkinsopp:
## Summary

For OpenXR, this fixes finger direction incorrectly being set to the tip direction rather than the intermediate, and fixes the `GrabStrength` calculation occurring before the `Hand.Rotation` had been calculated.

## Contributor Tasks

## Reviewer Tasks

- [ ] Code reviewed.

## Closes JIRA Issue

### DIFF
--- a/Packages/Tracking/OpenXR/Runtime/Scripts/OpenXRLeapProvider.cs
+++ b/Packages/Tracking/OpenXR/Runtime/Scripts/OpenXRLeapProvider.cs
@@ -198,6 +198,7 @@ namespace Ultraleap.Tracking.OpenXR
             for (int fingerIndex = 0; fingerIndex < 5; fingerIndex++)
             {
                 var xrTipIndex = 0;
+                var xrIntermediateIndex = 0;
                 var fingerWidth = 0f;
                 var fingerLength = 0f;
 
@@ -216,7 +217,7 @@ namespace Ultraleap.Tracking.OpenXR
                             metacarpalPosition,
                             metacarpalPosition,
                             metacarpalPosition,
-                            _joints[(int)HandJoint.ThumbMetacarpal].Pose.rotation * Vector3.forward,
+                            _joints[(int)HandJoint.ThumbMetacarpal].Pose.forward,
                             0f,
                             _joints[(int)HandJoint.ThumbMetacarpal].Radius * 2f,
                             (Bone.BoneType)boneIndex,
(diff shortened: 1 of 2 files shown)


Packages/Tracking/OpenXR/Runtime/Scripts/OpenXRLeapProvider.cs
```diff
@@ -33,6 +33,18 @@ public class OpenXRLeapProvider : LeapProvider
 
         // Magic numbers for palm width and PinchStrength calculation
         private static readonly float[] DefaultMetacarpalLengths = { 0, 0.06812f, 0.06460f, 0.05800f, 0.05369f };
+        private static readonly Quaternion[] ThumbMetacarpalRotationOffset =
+        {
+            Quaternion.Euler(0, +HAND_ROTATION_OFFSET_Y, +HAND_ROTATION_OFFSET_Z),
```
rblenkinsopp, line 38: I'd love to make these constants (`HAND_ROTATION_OFFSET_Y` and `HAND_ROTATION_OFFSET_Z`) non-public, but I think that ship has sailed until a major version bump.

MattGrayUL: I agree, though it would require that bump due to being public :( 

We could make new private ones, and then mark the public as obsolete and unused?

Packages/Tracking/OpenXR/Runtime/Scripts/OpenXRLeapProvider.cs
```diff
@@ -33,6 +33,18 @@ public class OpenXRLeapProvider : LeapProvider
...
+            Quaternion.Euler(0, -HAND_ROTATION_OFFSET_Y, -HAND_ROTATION_OFFSET_Z),
+        };
+        private static readonly Pose[] PalmOffset =
+        {
+            new Pose(new Vector3(-0.001039105f, -0.008749885f, 0.01165112f), Quaternion.Euler(-8f, -0.63f, -8.4f)),
+            new Pose(new Vector3(0.001039105f, -0.008749885f, 0.01165112f), Quaternion.Euler(-8f, +0.63f, +8.4f)),
+        };
```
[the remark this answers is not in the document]
rblenkinsopp: I've added an explaining comment

Packages/Tracking/OpenXR/Runtime/Scripts/OpenXRLeapProvider.cs
```diff
@@ -91,7 +91,7 @@ private TrackingSource CheckOpenXRAvailable()
             if (XRGeneralSettings.Instance != null &&
                 XRGeneralSettings.Instance.Manager != null &&
                 XRGeneralSettings.Instance.Manager.activeLoader != null &&
-                XRGeneralSettings.Instance.Manager.activeLoader.name == "Open XR Loader" &&
+                XRGeneralSettings.Instance.Manager.activeLoader.name.Replace(" ", "") == "OpenXRLoader" && // Some Unity versions have spaces and some don't so remove the spaces
```
rblenkinsopp, line 94: I think we should be using `XRGeneralSettings.Instance.Manager.ActiveLoaderAs<OpenXRLoaderBase>() != null` instead which actively tries to get it as an OpenXRLoader and will return null if it's not that type. We need to use `OpenXRLoaderBase` as there is `OpenXRLoader` and `OpenXRLoaderNoPreInit` which both are possible (not sure what uses the no pre-init version though.

MattGrayUL: Oh yeah that would be much cleaner - i'll take a look at that now

MattGrayUL: Oh yeah that would be much cleaner - i'll take a look at that now

rblenkinsopp: If the base doesn't work, we might need to test for both `OpenXRLoader` and `OpenXRLoaderNoPreInit` seperately.

MattGrayUL: It works great, and falls back to LeapC if not using OpenXR as the XR Plugin

PR #1464, opened 2023-06-12
Feature/unity 1160 optional subsystem

BryceNewell-UL:
## Summary

Adding a settings file and using it to enable and disable the use for the LeapC subsystem. Settings file can be found in the base of the "Tracking" plugin.

## Contributor Tasks

## Reviewer Tasks

- [x] Code reviewed.

## Closes JIRA Issue

### DIFF
--- a/Packages/Tracking/Core/Runtime/Scripts/UltraleapSettings.cs
+++ b/Packages/Tracking/Core/Runtime/Scripts/UltraleapSettings.cs
@@ -18,7 +18,6 @@ namespace Leap.Unity
             base.OnInspectorGUI();
 
             LeapSubSystemSection();
-        
         }
 
         private void LeapSubSystemSection()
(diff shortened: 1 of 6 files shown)


Packages/Tracking/Core/Runtime/Scripts/UltraleapSettings.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private void LeapSubSystemSection()
+        {
+            EditorGUILayout.Space(10);
+            EditorGUILayout.BeginVertical();
+            EditorGUILayout.LabelField("Leap XRHands Subsystem", EditorStyles.boldLabel);
+            
+            var readOnlyText = new GUIStyle(EditorStyles.textField);
+            readOnlyText.wordWrap = true;
+            EditorGUILayout.HelpBox("The Leap subsystem should be used when not using OpenXR for hand tracking inputs." +
```
MattGrayUL, line 31: Can this read:

> The Leap XRHands Subsystem passes hand input directly from the Leap Service to XRHands. If you are using OpenXR for hand input, do not enable this option. Instead, use the Hand Tracking Subsystem option in the OpenXR XR Plug-in Management settings.
> This option can not be toggled at runtime.

please?

Packages/Tracking/Core/Runtime/Scripts/UltraleapSettings.cs
```diff
@@ -0,0 +1,73 @@
...
+            bool editorBool = settingsTarget.LeapSubsystemEnabled;
+            EditorGUILayout.BeginHorizontal();
+            settingsTarget.LeapSubsystemEnabled = EditorGUILayout.ToggleLeft("Enable the leap subsystem ", editorBool);
+            EditorGUILayout.EndHorizontal();
+            EditorGUILayout.EndVertical();
+        }
+    }
+
+
+    //[CreateAssetMenu(menuName = "Ultraleap/Settings/PluginSettings")]  /// comment out so the user cannot create these but may be needed for development
```
MattGrayUL, line 44: no longer needed?

Packages/Tracking/Core/Runtime/Scripts/XRHandsSubsystem/SubsystemStarter.cs
```diff
@@ -11,17 +11,27 @@ public class SubsystemStarter
     {
         static XRHandSubsystem m_Subsystem = null;
         static XRHandProviderUtility.SubsystemUpdater updater = null;
+        static GameObject leapProviderGO = null;
+
+        static UltraleapSettings ultraleapSettings;
```
MattGrayUL, line 16: This is only used in RunAfterSceneLoad so does not need to be a static here?

MattGrayUL (commented):
This works great with my initial test in editor, need to test in builds etc still.

My only other initial thought is that we should have a quick menu option to select the Ultraleap Settings asset - in the bar at the top, the same as we have the "Ultraleap/Show Readme For Scene" option

Packages/Tracking/Core/Runtime/Scripts/UltraleapSettings.cs
BryceNewell-UL: So this is the only way to create a new SO. Not needed for the user but might be useful if we ever need to delete the scriptable object during development this is the only way to get it back. 

Tempted to leave it in commented out for this reason

MattGrayUL: I think if we were to do that, we should just have an option to reset and some cached defaults - ideally users shouldn't be creating/deleting this asset

MattGrayUL (commented):
The editor content will need to be wrapped in #if UNITY_EDITOR including the "using UnityEditor" or this content will not build

PR #1574, opened 2024-03-08
Hints no longer make new connections when it is not required

MattGrayUL:
hints no longer make new connections when it is not required. LSPs avoid connection coroutines when possible


Dwarph (approved):
matt told me to hit approve, so i have

Issue #1711, opened 2026-04-29
File descriptor exhaustion after repeated Unity Editor start/stop


MariusGr:
**Describe the bug**
In Unity Editor 6000.4.3f1, with Ultraleap Tracking 7.3.0 installed in a project, after starting and stopping play mode a few times, this error occurs:

`Unhandled exception during build: System.NotSupportedException: Could not register to wait for file descriptor <file descriptor number>
`

The number at the end is usually something between 1000-1200. It seems like something in Ultraleap leaks open file descriptors. The number increases with each start and stop of the play mode until the number of open file descriptors for the Unity process exceeds 1000. From this point on, the exception above occurs sometimes during play mode, and sometimes during edit mode. In any case, it prevents starting another play mode since Unity recognizes this as compilation error and refuses to play.

If it occurs during play mode, other functionalities sending SocketIO messages in my project do not work anymore. Thus, I suspect that something regarding Ultraleap's connection attempts to the Tracking Software might be causing this.

As of now, the only workaround for this is to close the Editor and open it again as this resets the open file descriptors.

**To Reproduce**
1. Open a completely new Unity project with "Universal 3D"
2. Start & Stop play mode once
3. Execute in terminal
`lsof -n | awk '{print $1}' | uniq -c | sort -rn | head -n $LINES
`
You will see an entry with title "Unity" and some number around 200-300. This is the number of currently open file descriptors for Unity.

5. Start and stop the play mode a few times. Then, execute the command again. You will see no increase in that number.
4. Install Ultraleap Tracking 7.3.0 via the Package Manager.
6. Retry step 5. Now, you will see an increase of the open file descriptors, which should be roughly 20-40 for one start & stop of play mode.
7. Repeat until the number reaches a value above 1000 and a few times more, then you should get the exception mentioned above.

**Project details**
- Unity version: Unity Editor 6000.4.3f1, also encountered in 6000.3.11f1
- Plugin/package version: Ultraleap Tracking 7.3.0
- Installation method: OpenUPM
- Packages affected: Ultraleap Tracking

rblenkinsopp:
Thank you for reporting this issue, I've just been able to also reproduce it (on macOS) and will investigate further. Can I double check which platform you are seeing this on?

rblenkinsopp:
Could you see if this PR helps with this issue: https://github.com/ultraleap/UnityPlugin/pull/1713

It resolves it for me in my testing.

MariusGr:
[...]

PR #1713, opened 2026-04-30
Fix resource leaks in Connection

rblenkinsopp:
This makes a number oc changes to improve `Connection` teardown logic:

- Fixed a file descriptor leak in `Connection` that exhausted the Unity process's OS handle limit after repeated play-mode start/stop cycles.
- Fixed an unmanaged memory leak where the marshalled server namespace string passed to LeapC was not freed.
- Fixed a managed memory leak in `Controller` where event subscriptions on the pooled `Connection` were not removed on disposal.
- Fixed a missing timeout on `Connection` shutdown which could lead to delays in applications exiting.


MaxPalmer-UH:
From a testing perspective looks like this sorts the file descriptor issue from unity when stop/starting in MacOS.

rblenkinsopp:
@MaxPalmer-UH Can you re-test this now it has updated LeapC binaries. I think this should resolve all the outstanding issues on hanging/teardown (#1711, #1716) as well as the other Unity internal issues.

DrMaxP:
Yes, that's fine

MaxPalmer-UH:
I'm happy with these changes.

rblenkinsopp:
@MaxPalmer-UH I've updated the change log, can you approve and I think we should merge, we can supersede the binaries later if required with a newer release.

MaxPalmer-UH (approved):

MaxPalmer-UH:
Done

PR #1714, opened 2026-06-10
Fixes for API deprecation warnings when using plugin in Unity 6.4 

MaxPalmer-UH:
Fixes several warnings when importing the plugin in Unity 6.4

### DIFF
--- a/Packages/Tracking Preview/AggregationProviders/Runtime/Scripts/JointOcclusion.cs
+++ b/Packages/Tracking Preview/AggregationProviders/Runtime/Scripts/JointOcclusion.cs
@@ -28,7 +28,11 @@ public class JointOcclusion : MonoBehaviour
     /// </summary>
     public void Setup()
     {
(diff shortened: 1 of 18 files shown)


rblenkinsopp:
The code changes all look good, have you checked that the `.meta` file changes increase in `serializedVersion` don't mean they have issues importing in older but still supported  versions?

MaxPalmer-UH:
> The code changes all look good, have you checked that the `.meta` file changes increase in `serializedVersion` don't mean they have issues importing in older but still supported versions?

No, that's a good shout. There will likely be issues there.

rblenkinsopp (changes requested):
Looks good but check that serialised assets and it needs a changelog entry as well.

MaxPalmer-UH:
Just checked importing the plugin in Unity 6.0, seemed to work fine and not show any warnings related to meta files or those components. Also added a changeling entry.

rblenkinsopp (approved):

